PR #2, opened 2017-02-23
separate uniq instruction into setShared and uniq

o-:
our uniq instruction was a mixture of shallow copying the value and
bumping the named count. this worked well in some situations but in
general is not fine grained enough.

the new semantics is:
* uniq: if the tos value is shared then copy it (such that the tos is
        now an unique copy of the value)
* set_shared: bump the named count to 2 such that tos is marked as
              shared (ie. will be copied by a subsequent uniq).

### DIFF
--- a/rir/src/code/dataflow.h
+++ b/rir/src/code/dataflow.h
@@ -93,6 +93,19 @@ class FValue {
         return u.use != UseDef::unused();
     }
 
+    bool singleUse() const {
+        if (hasUseDef()) {
+            assert(u.use != UseDef::unused());
+            return u.use != UseDef::multiuse();
+        }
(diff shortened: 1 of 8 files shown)


o-:
btw. my hope is that this commit will finally fix the benchmarks where we are slower than the ast interpreter...

petamaj:
I would really like `uniq_` to be renamed more clearly, something like `make_unique_` or so so that we it is obvious from its name what it does. Like `set_shared_` which is perfect.

o-:
I have renamed the instruction. Are you happy with the rest of the PR?

petamaj:
Yep, thank you!

Issue #13, opened 2017-02-24
Removal of as-package code


petamaj:
Initially we have envisioned `rir` to work in two modes, the latter being as a standalone package to vanilla `gnu-r`. This has soon been abandoned. Check that there are no remnants of this functionality in current code base if there are, delete them.

kondziu:
I found a few things that might also warrant removal in there, I'd like an opinion on that.

Thing 1: `USE_RIR_CONTEXT_SETUP`
It's a define that was only set to true if `RIR_AS_PACKAGE` was off. It's used in a couple of places in `interp.c`, I think when calling external functions. Don't know if that is actually a separate switch used for something else or if it would be always on if `RIR_AS_PACKAGE` is off.

Things 2 and 3: `createWrapperAst` and `createWrapperPromise`
When `RIR_AS_PACKAGE` is turned off these functions just return their argument. It looks like  `createWrapperAst` has a couple of occurances. `createWrapperPromise` doesn't have any that I can see.

o-:
Hi, thanks for looking into this.
1. So this flag is not really related. R internally uses the RCNTXT structs to track call frames. At some point we had an experimental workaround to support creating contexts in RIR. But by now we have made the RCNTXT struct and the required apis public. As you can see for example in https://github.com/reactorlabs/rir/blob/master/rir/src/interpreter/interp.c#L504 we are perfectly capable of creating proper contexts and as you saw USE_RIR_CONTEXT_SETUP is always true. Feel free to remove it in the same go.

2./3. They can both be removed. The Idea there was that we could execute RIR methods even if gnu R does not know our bytecode. The wrappers would be legal R functions which would tail-call a C primitive which would tail-call into our interpreter. :)

PR #20, opened 2017-03-13
Removal of as-package code

kondziu:
This is with relation to issue #13 and the comments there. 

It removes:
1. The `RIR_AS_PACKAGE` def and all the code that depended on it.
2. The `USE_RIR_CONTEXT_SETUP` def and its occurrences.
3. `createWrapperAst` and `createWrapperPromise` functions, which became identities after `RIR_AS_PACKAGE` was removed.

### DIFF
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -86,9 +86,6 @@ void endClosureContext(RCNTXT* cntxt, SEXP result) {
 
  */
 INLINE SEXP createPromise(Code* code, SEXP env) {
-#if RIR_AS_PACKAGE == 1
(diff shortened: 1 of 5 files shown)


rir/src/interpreter/interp.c
```diff
@@ -97,7 +94,6 @@ INLINE SEXP createPromise(Code* code, SEXP env) {
     SET_ATTRIB(p, a);
```
o-, line 94: I think this hack is (a) broken and (b) not needed. We should remove it as well. But I can look into it.

o- (approved):
Thanks! Looks good to me.

PR #21, opened 2017-03-13
cleanup some interpreter stuff

o-:
Remove some currently unused cruft from the interpreter

### DIFF
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -82,17 +82,8 @@ void endClosureContext(RCNTXT* cntxt, SEXP result) {
     endcontext(cntxt);
 }
 
-/** Creates a promise from given code object and environment.
-
- */
 INLINE SEXP createPromise(Code* code, SEXP env) {
     SEXP p = mkPROMISE((SEXP)code, env);
-    PROTECT(p);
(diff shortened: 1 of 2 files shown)


rir/src/interpreter/interp.c
```diff
@@ -494,27 +480,22 @@ INLINE SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
 
     ostack_push(ctx, newEnv);
 
-    // Create a new R context and store it on the stack
```
kondziu, line 497: @o- I don't understand this part at all (not criticism, just statement of fact). Why can this be so drastically simplified?

o-: So at some point the context of a function was allocated on the heap. I can't remember the exact details. To be able to put it on the interpreter stack it was boxed in a RAWSXP. This changes it to allocate the context on the C runtime stack. Since the execution of the R function coincides with the execution of the rirCallClosure C function this works well.

For example in the case of the loop context that does not work and we still allocate loop contexts on the R heap and store it in the RIR interpreter stack.

kondziu:
Looks fine. I don't have write access to the repo so I can't merge though.

PR #27, opened 2017-03-29
Added bc for rel. ops, negation, unary minus

JanJecmen:
Plus some minor formatting and script to plot benchmark speedup of one given revision


tools/bench-speedup.sh
```diff
@@ -0,0 +1,130 @@
...
+
+export R_NO_BASE_COMPILE=1
+
+pushd $PLAIN_DIR
+make clean
+if [ $USING_OSX -eq 1 ]; then
+  # Mac OSX
+    F77="gfortran -arch x86_64" FC="gfortran -arch x86_64" CXXFLAGS="-g3 -O2" CFLAGS="-g3 -O2" ./configure --enable-R-shlib --without-internal-tzcode --with-ICU=no --disable-byte-compiled-packages
+else
+    CXXFLAGS="-g3 -O2" CFLAGS="-g3 -O2" ./configure --with-ICU=no --disable-byte-compiled-packages
```
o-: --disable-byte-compiled-packages will give you quite different perf

maybe it's a fairer comparison? but still, not comparable to what people will get out of the box

tools/bench-speedup.sh
```diff
@@ -0,0 +1,130 @@
...
+fi
+make -j 8
+popd
+
+
+# checkout rir revision and build rir
+git checkout $RIR_REVISION
+
+ninja clean || true
```
o-: should probably use `cmake --build`. Can't assume everyone to use ninja

rir/src/interpreter/interp.c
```diff
@@ -1910,40 +1910,212 @@ INSTRUCTION(sub_) {
...
+    SEXP rhs = ostack_at(ctx, 0);
+    SEXP res;
+
+    DO_RELOP(!=);
+
+    ostack_popn(ctx, 2);
+    ostack_push(ctx, res);
+}
+
+INSTRUCTION(not_) {
```
o-, line 2113: wouldn't you expect booleans to be common here? I think they should be in the fast path here and maybe also for others like `==`

o- (commented):
Hi, awesome. Thanks for creating the patch. It looks good to me except for some inline comments. Do you measure speedups? One of the travis jobs timed out. Sometimes it indicates a performance regression, but sometimes it's just travis being slow...

tools/bench-speedup.sh
JanJecmen: OK, good point. Probably should then also change https://github.com/reactorlabs/rir/blob/master/tools/benchmark.sh#L34 and https://github.com/reactorlabs/rir/blob/master/tools/benchmark-revisions.sh#L35?

tools/bench-speedup.sh
JanJecmen: Yeah, the idea was to take the ast interpreter as baseline and then see how gnu-r compiler and rir matched against each other. I thought that it would be better if both had to compile everything just in time during the benchmark, so rir wouldn't use gnu-r bytecode from the default packages.

But it's true that out of the box users will get the base and recommended packages pre-compiled, and rir will run on top of that, so maybe this is unnecessary.

rir/src/interpreter/interp.c
JanJecmen: Thanks, will change that..

tools/bench-speedup.sh
o-: oh I see, I am correcting my mistakes and not yours :) As you want

tools/bench-speedup.sh
o-: Actually running RIR with R_ENABLE_JIT should recompile the package code. As soon rir is loaded then bytecompiled code and uncompiled code will be treated the same. So I think disabling precompiling packages is unfair to the stock R. That's also why we do one round of `execute()` without measuring as warmup. Otherwise it would be unfair to us, since we compile all the package code at runtime.

o-: (Sidenote, maybe we should switch to R_ENABLE_JIT=3 to be extra sure...)

JanJecmen: Oh ok... shouldn't have assumed that rir just runs gnu-r bytecode when it sees it.
Thanks for the explanation!

o-:
thanks!

Issue #5, opened 2017-02-23
adjust benchmark time


o-:
On our [perf site](http://ginger.ele.fit.cvut.cz/~oli/r-we-fast/) you can see that some benchmarks run for less than 1 second. Those are useless. We should go through them and give them better defaults. To do so just change the default value for the `execute` function in each affected benchmark in `benchmarks/shootout`.

o-:
done #28

PR #28, opened 2017-04-03
Increase benchmark running time

JanJecmen:
Some benchmarks were running almost instantenously. This changes their defaults.

Measured on ginger, aim was to get all to take at least a second

### DIFF
--- a/benchmarks/shootout/knucleotide/knucleotide.r
+++ b/benchmarks/shootout/knucleotide/knucleotide.r
@@ -82,7 +82,7 @@ paste. <- function (..., digits=16, sep=" ", collapse=NULL) {
     }
 }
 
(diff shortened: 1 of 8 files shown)


o-:
awesome. this closes issue #5 .

btw. are you interested in replacing the benchmark site (http://ginger.ele.fit.cvut.cz/~oli/r-we-fast/)?

JanJecmen:
@o- What do you mean by replacing?

o-:
@JanJecmen oh,  I meant that we could work towards using your combined benchmark script as the "official" benchmarks. But I think we also need to migrate to a dedicated benchmark machine, if you look at the latest runs the error intervals are huge.

Btw. nicely done. Since your changes we are not slower than ast anymore and we beat the BC on one additional benchmark.

PR #29, opened 2017-04-05
Fix typo in rir.eval

JanJecmen:
The default argument for env should be globalenv() (all lowercase).

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -19,7 +19,7 @@ rir.compile <- function(what) {
     .Call("rir_compile", what)
 }
 
-rir.eval <- function(what, env = globalEnv()) {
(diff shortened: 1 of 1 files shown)


o-:
that's all good of course! I would not mind if you merge such simple things with the comment "to be reviewed", then we can still retroactively comment...

JanJecmen:
I don't have write access so I couldn't merge, sorry. I know it's just one letter change..

o-:
ah sorry that was not the intention. Nobody should be able to push directly but everybody should be allowed to merge a PR where the checks are green. Can you try again. Changed some settings.

JanJecmen:
Works fine now, thanks

PR #31, opened 2017-04-11
Removal of unnecessary loop contexts

JanJecmen:


JanJecmen:
@petamaj `return` does not trigger loop context in gnur:
```
> f
function(x) repeat foo(if (x) return(42) else 3)
<bytecode: 0x16acfc0>
> compiler::disassemble(f)
list(.Code, list(8L, GETFUN.OP, 1L, MAKEPROM.OP, 2L, CALL.OP, 
    3L, POP.OP, GOTO.OP, 1L, LDNULL.OP, INVISIBLE.OP, RETURN.OP), 
    list(repeat foo(if (x) return(42) else 3), foo, list(.Code, 
        list(8L, GETVAR.OP, 1L, BRIFNOT.OP, 0L, 9L, LDCONST.OP, 
            2L, RETURNJMP.OP, LDCONST.OP, 3L, RETURN.OP), list(
            if (x) return(42) else 3, x, 42, 3)), foo(if (x) return(42) else 3)))
```

rir/R/rir.R
```diff
@@ -16,7 +16,12 @@ rir.disassemble <- function(what) {
 
 # compiles given closure, or expression and returns the compiled version.
 rir.compile <- function(what) {
-    .Call("rir_compile", what)
+    # TODO: gnu-r compile function takes optional environment argument for expressions
+    # this just uses globalenv for now
```
o-, line 20: why not change the compile function? globalenv seems like a strange thing to pass here.

rir/src/ir/Compiler.cpp
```diff
@@ -27,44 +27,58 @@ class Context {
...
+        loops.emplace(next_, break_);
+    }
+
+    void popLoop() { loops.pop(); }
 
     void push(SEXP ast) { code.emplace(ast, fun); }
 
+    void pushPromiseContext(SEXP ast) {
+        push(ast);
+        code.top().in_promise_ = true;
```
o-: I would suggest to make PromiseContext a subtype of CodeContext instead of the boolean flag.

rir/src/ir/Compiler.cpp
```diff
@@ -27,44 +27,58 @@ class Context {
...
 
     class CodeContext {
       public:
         CodeStream cs;
-        std::stack<LoopContext> loops;
+        bool in_promise_ = false;
         CodeContext(SEXP ast, FunctionHandle& fun) : cs(fun, ast) {}
     };
 
+    std::stack<LoopContext> loops;
```
o-: This looks very dangerous because you loose track of where the loop is (inside promise or toplevel). What was the reason you pulled it out?

Because it will be very easy to confuse the situation where the loop is toplevel and the next in the promise and the situation where both are in a promise.

For example in
```
foo(for (...) {next})
```
I think you don't need a context. Your code is conservative and creates one. But still in the future we might want to elide this one too so we need to know if the loop belongs to the promise or the function.

The context is only needed if there is a `next/break` and the loop-ctx is not in the innermost code-ctx.

rir/src/ir/Compiler.cpp
```diff
@@ -797,18 +912,39 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
     if (fun == symbol::Next && ctx.inLoop()) {
         assert(args.length() == 0);
 
+#ifdef GNUR_COMPATIBILITY
+        if (!ctx.inPromise()) {
```
o-: (also see comment above regarding the placement of the loop contexts)

I believe you can sometimes also elide the context inside a promise. I think the correct condition is:
1. in a code/prom-ctx with a loop-ctx -> no context needed, we know where to jump
2. in a promise-ctx below a code-ctx in the ctx stack where the code-ctx has a loop-ctx -> context is needed for the longjmp to the outer function
3. no code-ctx on the ctx stack has a loop-ctx -> this seems like an error in the user code (I think the bc compiler prints a warning)

CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ set(CMAKE_C_FLAGS_RELEASE "-O2 -std=gnu99")
 set(CMAKE_C_FLAGS_DEBUG "-std=gnu99 -DENABLE_SLOWASSERT")
 set(CMAKE_C_FLAGS "-std=gnu99")
 
+add_definitions(-DGNUR_COMPATIBILITY)
```
o-, line 21: I would not make this an ifdef but just have it as the default behavior

rir/src/ir/CodeStream.h
```diff
@@ -230,6 +230,80 @@ class CodeStream {
...
+
+    unsigned currentPos() const {
+        return pos;
+    }
+
+    unsigned currentSourcesSize() const {
+        return sources.size();
+    }
+
+    void insertIntoStream(BC instr, unsigned beforePos, unsigned sourcesIdx) {
```
o-: I find it not a super good idea to insert into the code stream. There is another class called a CodeEditor that has this functionality.

Also it seems a bit of a questionable design decision to insert the context instructions after the fact, but before finalize. I would have assumed that you do one of the two:
1. Find out beforehand if context is needed by analyzing the AST and create code accordingly.
2. Find out after the fact by analyzing the bytecode and removing the context if unnecessary (this is what you could do in an analysis pass).

I guess 1 is hard to do, because it depends on the optimizations that the compiler does. and 2 is annoying because of how the for loop juggles the context on the stack.

Would you be interested in reimplementing it in an optimization pass? I could help you by removing the ugly stack juggling code from the for loop (I wanted to do this for long). Then the only change needed would be to remove the create-context and end-context instructions.

If you find that overblown I would suggest to have an override instruction function on the stream and replace the context creation by 'nop' instructions (that we'd still need to add). At least I would really prefer not to add the ability to insert into the stream.

o- (commented):
Hi, thank a lot for creating the patch. I see you tried really hard to make it work on the fly while compiling. But maybe it would be better suited for an optimization pass that removes the context after the fact? (See my long comment below the insert function for the codestream.)

I have some comments (also regarding this point) inline. I tried to use `ctx` for the C++ context class and `context` for the R loop contexts to avoid confusion.

rir/R/rir.R
JanJecmen: This was inspired by the gnur functions `cmpfun` and `compile`. The first one gets a closure `f` and compiles it in a new compiler environment that is created from the `environment(f)`. The second gets an expression and an environment (that defaults to `.GlobalEnv`) and compiles with this. Since we only have one function as API, I tried to put all this in it..
Anyway, this was just in preparation to actually working with environments in the compiler, which is not yet the case.

JanJecmen:
@o- Ok, so I added the `nop_` instruction, then changed the inserting into `CodeStream` procedure to use the `nop_`s instead. Then I moved the loop context stack handling back to `CodeContext` and added `PromiseContext`. I hope I got it right... Also I changed `compileClosure` to optimize promises, too, is that OK?

However the problem with the build remains, rir now produces the long error message containing source location...

PS: all these give the same results as gnur compiler with regards to loop contexts:
```
function() break  # here gnur issues a warning
function() repeat break
function() repeat foo(break)  # context needed
function() repeat foo(repeat break)
function() repeat foo(bar(break))  # context needed
function() repeat foo(bar(repeat break))
function() foo(break)  # here also a warning
function() foo(repeat break)
```

rir/src/ir/Compiler.cpp
```diff
@@ -1100,6 +1205,10 @@ Compiler::CompilerRes Compiler::finalize() {
     CodeEditor code(function.entryPoint(), formals);
     Optimizer::optimize(code);
 
+    for (size_t i = 0; i < code.numPromises(); ++i)
+        if (code.promise(i))
+            Optimizer::optimize(*code.promise(i));
```
o-, line 1206: good catch!

rir/src/ir/Compiler.cpp
```diff
@@ -35,33 +36,89 @@ class Context {
...
+        virtual ~CodeContext() {}
+        bool inLoop() {
+            return !loops.empty() ||
+                    (parent && parent->inLoop());
+        }
+        LabelT loopNext() {
+            return !loops.empty() ? loops.top().next_ : parent->loopNext();
+        }
+        LabelT loopBreak() {
+            return !loops.empty() ? loops.top().break_ : parent->loopBreak();
```
o-: I think those two are wrong. Wouldn't this cause the next inside the promise to use the label of the outer function which is not a valid label in the promise. (The bug probably did not happen in practice since gotoOK returns false in that case.)

rir/src/ir/Compiler.cpp
```diff
@@ -35,33 +36,89 @@ class Context {
...
+        }
+        void setContextNeeded() {
+            if (loops.empty() && parent)
+                parent->setContextNeeded();
+            else
+                loops.top().context_needed_ = true;
+        }
+        virtual bool gotoOK() {
```
o-: can you give this a more meaningful name (maybe loopIsLocal() or knownLoopHeaderPos)

rir/src/ir/CodeStream.h
```diff
@@ -230,6 +230,26 @@ class CodeStream {
...
+    void remove(unsigned pc, unsigned size, unsigned sourcesIdx) {
+
+        for (unsigned i = 0; i < size; ++i) {
+            *reinterpret_cast<Opcode*>(&(*code)[pc + i]) = Opcode::nop_;
+            // patchpoints are fixed by just removing the binding to label
+            patchpoints.erase(pc + i);
+        }
+
+        // need to insert source slots for the nops occupying the immediate places
+        sources.insert(sources.begin() + sourcesIdx + 1, size - 1, 0);
```
o-: damn, didn't think about that... is it possible to compute the sourceIdx from the pc?

rir/src/ir/CodeStream.h
```diff
@@ -230,6 +230,26 @@ class CodeStream {
...
 
+    unsigned currentPos() const {
+        return pos;
+    }
+
+    unsigned currentSourcesSize() const {
+        return sources.size();
+    }
+
+    void remove(unsigned pc, unsigned size, unsigned sourcesIdx) {
```
o-: you can use `BC(code[pc]).size()` to get this size, no need to pass it.

o- (commented):
Looks great, I am very happy with the changes. After the remaining small bits this lgtm!

rir/src/ir/CodeStream.h
JanJecmen: Well assuming that every insertion of a BC adds one slot, then if you knew how many instructions are before the pc, you would get sourceIdx?
So you could advance from 0 until you got to pc?

o-: Yeah, that should work. Something like
```
unsigned idx = 0;
Opcode* pos = &code[0];
while (pos < &code[pc]) {
  BC::advance(&pos);
  idx++;
}
```

Not sure what is better. To keep the sourcesIdx on the side (seems dangerous) or to do this (is slow..)

o-: maybe a slight preference to this. But feel free to decide what you like more.

o-:
Regarding the changed test, let's not waste too much time on it and change the test case there. For this the gnur repository needs to be changed. That will be interesting since we have not done it in the new setup :)

If you look at the commit history of the gnur repo:
https://github.com/reactorlabs/gnur/commits/rir-patch-3-3-branch
I tried to keep the patches really clean. This allows us to easily port our patches to newer R version.

Therefore I would checkout the `rir-patch-3-3-branch` in gnur repo and then use `git rebase -i R-3-3-branch` to edit this commit:
https://github.com/reactorlabs/gnur/commit/4634598787432b70893b4457fb2c25324d17423a
and then force push the updated branch.

You should have write access there. If you want I can also do it, or give you more hints.

Hopefully the `tools/sync.sh` script from rir should automatically switch to the updated version on the next make...

o-:
Oh and btw. can you also do a `git rebase -i master` on this branch and squash everything into one commit before you merge?

rir/src/ir/Compiler.cpp
JanJecmen: True. You should never go to the parent for the labels, so maybe an assert?

JanJecmen:
Sure. Hope I got that gnur rebase right...

o-:
thanks

PR #33, opened 2017-04-13
Add colon_ instruction for integer sequences, ie. 1:10

JanJecmen:


JanJecmen:
I am stuck here.. The build fails with an assertion error:
```
honza@ubuntu:~/Documents/fork-rir/external/custom-r/tests$ EXTRA_LOAD_SO=../../../build/librir.so EXTRA_LOAD_R=../../../rir/R/rir.R R_ENABLE_JIT=2 LANGUAGE=en LC_ALL=C SRCDIR=. R_DEFAULT_PACKAGES= ../bin/R --vanilla < reg-tests-1a.R 
...
> arima0(y, xreg = x, seasonal = list(order=c(0,1,0)))
R: ../rir/src/interpreter/interp.c:1064: ins_close_: Assertion `isValidFunctionSEXP(body)' failed.
Aborted (core dumped)
```
However I had no luck isolating this. When I try to remove anything from the test, the error doesn't happen, even if I just simply copy/paste the contents of `reg-tests-1a.R` everything works fine.

Any tips?

EDIT: Found a bug where inserting into cp caused gc and this reclaimed the bc object of a newly compiled closure

o-:
Sorry I was a bit preoccupied. Is your latest commit related to the bug?

What I started using in such situations is [rr](http://rr-project.org/). It's incredible. You can record a trace with `R -d rr -f file.r` then just `rr replay`. When it asserts on the corrupt SEXP you set a watchoint. (ie. `p &body->sxpinfo` then using the address it gives you: `watch *0x123` and finally `reverse-continue`. :)

JanJecmen:
Oh ok thanks, I will try that next time :)
Yes the commit solves the issue, although I am not sure if something similar could still occur some place else...

o-:
> Yes the commit solves the issue,

oh my. I think this is a failure of Pool, or even better in rir/src/interpreter/interp_context.h rl_append should protect val in the grow case.

> although I am not sure if something similar could still occur some place else...

sure do :)

JanJecmen:
You are right, that is a better place to put this.

JanJecmen:
@o- Other than that, do you think it is ok to merge?

o-:
Yeah, lgtm

PR #35, opened 2017-04-17
some random performance fixes

o-:


JanJecmen:
about the for loops in general:
- you mentioned before the use of another stack for loop contexts and removal of the context juggling instructions
- for `:` maybe we could avoid allocating the whole sequence?

o-:
> you mentioned before the use of another stack for loop contexts and removal of the context juggling instructions

yeah, not today :)

> for : maybe we could avoid allocating the whole sequence?

That should be possible. But we need some sort of local variables in the interpreter that are not accessible to the program. Because otherwise it could change the loop variable.
(OK we could also do some stack-foo. But that's annoying)

PR #37, opened 2017-04-19
Superassign

JanJecmen:


rir/src/interpreter/interp.c
```diff
@@ -1216,6 +1238,17 @@ INSTRUCTION(stvar_) {
...
 
+INSTRUCTION(stvar2_) {
+    SEXP sym = readConst(ctx, pc);
+    int wasChanged = FRAME_CHANGED(env);
+    SLOWASSERT(TYPEOF(sym) == SYMSXP);
+    SEXP val = ostack_pop(ctx);
+    INCREMENT_NAMED(val);
+    setVar(sym, val, ENCLOS(env));
+    if (!wasChanged)
+        CLEAR_FRAME_CHANGED(env);
```
o-: the FRAME_CHANGED keeps track of whether some non-local change to the env happened. Here this is the case (super assign is non local). Also the change happens on ENCLOS(env). So you can just delete lines 124[3,8,9]

rir/src/ir/Compiler.cpp
```diff
@@ -475,6 +478,11 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
             //            if (getter == symbol::Bracket)
         }
 
+        if (superAssign) {
```
o-, line 484: can you move the guard after the condition (only inserting it if needed) instead of removing it afterwards?

o- (commented):
thanks, that looks good (except the two nitpicks)

rir/src/interpreter/interp.c
JanJecmen: Thanks for the explanation. Fixed that.

PR #38, opened 2017-04-20
cache bindings of the local frame

o-:
a very small cache per eval invocation keeps track of the most recently
looked up variable from the inner most environment. The location of
the binding in the environmen is stored in the cache. This saves us
looking through the linked list of bindings in the local env.

The mechanism is very similar to the gnur BC interpreter. Though a bit
more lightweight.

long term something more intelligent would be nice. But for now this
seems like a good addition. Speeds up some things by 20%.

### DIFF
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -1192,8 +1192,75 @@ void debug(Code* c, OpcodeT* pc, const char* name, unsigned depth, Context* ctx)
     }
 }
 
+#define BINDING_CACHE_SIZE 11
+typedef struct {
+    SEXP loc;
+    Immediate idx;
+} BindingCache;
+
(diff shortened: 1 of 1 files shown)


o-:
@JanJecmen are you interested in looking at this?

I needed access to a new local variabel (bindingCache) in the interpreter loop. Since we are moving in that direction anyways I decided to already inline the affected instructions. I hope this does not interfere too much with your planned changes.

If it causes you pain I we can also delay and I can rebase after you.

JanJecmen:
Yes, it looks great!
I took your gist and attempted to clean it up a bit. I will create PR, then we could merge the bindings cache and then I would like to look at threading and see how that helps.

o-:
> Yes, it looks great!

mhyeah, except it breaks :) I'll have to look at it again tomorrow.

> I took your gist and attempted to clean it up a bit. I will create PR, then we could merge the bindings 

nice. Yeah let's merge your changes first

PR #39, opened 2017-04-20
Interpreter speedup

JanJecmen:


o-:
ah I didn't catch it myself. but we should try to avoid broken intermediate commits because the benchmark script will choke on them

JanJecmen:
Oh sorry, haven't thought about that. I will be more careful next time

PR #40, opened 2017-04-21
Speed up loops

JanJecmen:
Shuffled loop code around so that some jumps are avoided (one for every iteration but the first):

Instead of 
```
COND:
    <cond>
    brfalse_ END
    <body>
    br_ COND
END:
```

We now have
```
    br_ COND
BODY:
    <body>
COND:
    <cond>
    brtrue_ BODY
END:
```

`for` loop is similar, but a bit messier. `repeat` is not affected.

### DIFF
--- a/rir/src/ir/Compiler.cpp
+++ b/rir/src/ir/Compiler.cpp
@@ -830,29 +830,27 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
 
         cs << BC::guardNamePrimitive(fun);
 
-        LabelT loopBranch = cs.mkLabel();
-        LabelT breakBranch = cs.mkLabel();
-        LabelT endForBranch = cs.mkLabel();
+        LabelT condBranch = cs.mkLabel();
+        LabelT bodyBranch = cs.mkLabel();
+        LabelT afterBranch = cs.mkLabel();
 
-        ctx.pushLoop(loopBranch, breakBranch);
+        ctx.pushLoop(condBranch, afterBranch);
 
         compileExpr(ctx, seq);
         cs << BC::setShared()
            << BC::push((int)0);
 
         std::vector<unsigned> pcs;
-
         pcs.push_back(cs.currentPos());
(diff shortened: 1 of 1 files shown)


JanJecmen:
Well now I am not so sure about this...
While on something like `function() { i <- 10000000; while (i > 0) i <- i - 1 }` it gets somewhat faster, other stuff (notably the benchmarks) seem to slow down with this :(

JanJecmen:
It looks like it doesn't even mess with optimizations, so I guess it has to do with caches? Maybe this causes increased miss rate because of the jump?
Anyway, it is probably not worth it...

EDIT: here is what I used to measure (from the mandelbrot benchmark):
```
f <- function() {
    n = 1000L
    for (y in 0:(n-1)) {
        c <- 2 * 0:(n-1) / n - 1.5 + 1i * (2 * y / n - 1)
        z <- rep(0+0i, n)
        i <- 0L
        while (i < 200) {
            z <- z * z + c
            i <- i + 1L
        }
    }
}
system.time(f())
t <- c()
for (x in 1:10) t <- c(t, system.time(f())[[3]])
mean(t)
```
Original: 2.3907 s, modified: 2.4443 s

JanJecmen:
I have just noticed, in the branch instructions, there is a call to `incPerfCount`, but only if the jump offset is negative, i.e. we are jumping backwards in the bytecode.

I am not sure what this is supposed to do (the value is used in `rirCallClosure` for some optimization logic), but maybe this could cause the slowdown?

Before the jump was forward to the end of the loop body, here it leads back before it...

```
    if (ostack_pop(ctx) == R_TrueValue) {
            pc = pc + offset;
            if (offset < 0)
                incPerfCount(c);
    }
```

o-:
I think the jmp is probably almost for free. So I am not surprised that it does not pay off...

Regarding the perf counter. That's supposed to measure how hot a method is. It is supposed to increment for every loop iteration. It's probably not a huge slowdown factor. And it should increment in both versions. (either on jmp or on branch).

JanJecmen:
Ok thanks for the explanation.

PR #41, opened 2017-04-24
Binop less alloc

o-:
The meat is in the last commit. The rest are cleanups.

The idea is to reuse the box on the stack to store the result if it has no references.

@JanJecmen I am sorry that I made you move all local variables out of the instructions. In this patch I propose to move them back. It does not seem to make a perf difference. But it is sooo much easier for debugging.


JanJecmen:
No problem. I didn't like how unreadable it became, either. 

I changed the loop to indirect threading, but that will be quite simple to edit after this gets merged, so I will wait with it.

o-:
@JanJecmen oh, sorry. Now I am not sure if I was to fast or if you where ok with merging this?

JanJecmen:
No it is fine :)

PR #42, opened 2017-04-24
binding cache: bugfix and some improvements

o-:
the ldvar instruction can return missing.


JanJecmen:
So from what I understand about gnur, they create the cache as large as the constant pool, up to a maximum of 256 elements. We have fixed size of 5, isn't it too small?

o-:
Honestly I am not really sure. It does not seem to help as much as I hoped. But if we want it to be bigger (and variable sized) we cannot allocate it on the C stack.

o-:
so my plan is to see if going from 11 to 5 tanks any of the benchmarks. If not trying something bigger is long term and would involve moving it from the C stack to the interpreter stack or heap.

PR #43, opened 2017-04-25
Indirect threading

JanJecmen:
Changed the interpreter switch loop to indirect threading: one time initialization computes the addresses of labels for each instruction into an array, and dispatching is then done in each instruction through this jump table.

Uses GNU computed goto's, so there is also a condition that disables this if not using gcc (the same as gnur)

Also changed the functions that access stack and constant / src pool to macros, so we do not rely on gcc inlining.

### DIFF
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -36,29 +36,14 @@ INLINE SEXP getSrcForCall(Code* c, OpcodeT* pc, Context* ctx) {
 
 #ifdef THREADED_CODE
 
-/* Declare volatile to prevent gcc 6 from making a local copy
-   in bcEval stack frames and thus increasing stack usage
-   dramatically */
-volatile static void * opAddr[numInsns_];
-
-#define INITIALIZE_MACHINE() if (c == NULL) goto init
-#define BEGIN_MACHINE  NEXT(); init: { loop: switch(opcode++)
+#define BEGIN_MACHINE NEXT();
 #define INSTRUCTION(name) \
-    case name: \
-        opAddr[name] = (__extension__ &&op_##name); \
-        goto loop; \
     op_##name:  // debug(c, pc, #name, ostack_length(ctx) - bp, ctx);
 #define NEXT() (__extension__ ({goto *opAddr[advanceOpcode()];}))
-#define LASTOP \
-    case numInsns_: \
-        return R_NilValue; \
-    default: \
-        assert(false && "wrong or unimplemented opcode"); \
-    }
+#define LASTOP {}
 
 #else
 
-#define INITIALIZE_MACHINE()  /* nothing */
 #define BEGIN_MACHINE  loop: switch(advanceOpcode())
 #define INSTRUCTION(name)  case name:  // debug(c, pc, #name, ostack_length(ctx) - bp, ctx);
 #define NEXT()  goto loop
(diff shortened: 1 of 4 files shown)


o-:
Generally that looks good. But, we have a list of all instructions. No need to dynamically generate the array of instruction labels. We can have it static:

```
diff --git a/rir/src/interpreter/interp.c b/rir/src/interpreter/interp.c
index 60ee7ba..ae55f27 100644
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -36,29 +36,15 @@ INLINE SEXP getSrcForCall(Code* c, OpcodeT* pc, Context* ctx) {
 
 #ifdef THREADED_CODE
 
-/* Declare volatile to prevent gcc 6 from making a local copy
-   in bcEval stack frames and thus increasing stack usage
-   dramatically */
-volatile static void * opAddr[numInsns_];
-
-#define INITIALIZE_MACHINE() if (c == NULL) goto init
-#define BEGIN_MACHINE  NEXT(); init: { loop: switch(opcode++)
+#define BEGIN_MACHINE NEXT();
 #define INSTRUCTION(name) \
-    case name: \
-        opAddr[name] = (__extension__ &&op_##name); \
-        goto loop; \
     op_##name:  // debug(c, pc, #name, ostack_length(ctx) - bp, ctx);
 #define NEXT() (__extension__ ({goto *opAddr[advanceOpcode()];}))
-#define LASTOP \
-    case numInsns_: \
-        return R_NilValue; \
-    default: \
-        assert(false && "wrong or unimplemented opcode"); \
-    }
+#define LASTOP                                                                 \
+    {}
 
 #else
 
-#define INITIALIZE_MACHINE()  /* nothing */
 #define BEGIN_MACHINE  loop: switch(advanceOpcode())
 #define INSTRUCTION(name)  case name:  // debug(c, pc, #name, ostack_length(ctx) - bp, ctx);
 #define NEXT()  goto loop
@@ -1310,11 +1296,13 @@ INLINE void cachedSetVar(SEXP val, SEXP env, Immediate idx, Context* ctx,
 SEXP evalRirCode(Code* c, Context* ctx, SEXP env, unsigned numArgs) {
 
 #ifdef THREADED_CODE
-    Opcode opcode = 0;
+    static void* opAddr[numInsns_] = {
+#define DEF_INSTR(name, ...) (__extension__ && op_##name),
+#include "ir/insns.h"
+#undef DEF_INSTR
+    };
 #endif
 
-    INITIALIZE_MACHINE();
-
     assert(c->magic == CODE_MAGIC);
 
     BindingCache bindingCache[BINDING_CACHE_SIZE];
diff --git a/rir/src/interpreter/runtime.c b/rir/src/interpreter/runtime.c
index 15b5976..7a85c53 100644
--- a/rir/src/interpreter/runtime.c
+++ b/rir/src/interpreter/runtime.c
@@ -85,9 +85,6 @@ void initializeRuntime(CompilerCallback compiler, OptimizerCallback optimizer) {
     // initialize the global context
     globalContext_ = context_create(compiler, optimizer);
     registerExternalCode(rirEval_f, compiler, rirExpr);
-#ifdef THREADED_CODE
-    evalRirCode(NULL, NULL, NULL, 0);
-#endif
 }
```

JanJecmen:
Nice, thanks for pointing that out!

o-:
nice

PR #46, opened 2017-05-03
Changed AbstractEnvironment to accept KEY as a template parameter

aviralg:
This pull request changes AbstractEnvironment and code that uses it to accept or pass the KEY as a template parameter. All tests pass.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -14,12 +14,6 @@ rir.disassemble <- function(what) {
     invisible(.Call("rir_disassemble", what))
 }
 
-# prints the disassembled versions of list of rir compiled expressions
-rir.disassemble.file <- function(code_objects) {
-  Map(rir.disassemble, code_objects)
-  NULL
-}
-
 # compiles given closure, or expression and returns the compiled version.
 rir.compile <- function(what) {
     # TODO: gnu-r compile function takes optional environment argument for expressions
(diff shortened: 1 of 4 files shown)


aviralg:
Merge this before #47

o-:
Nothing wrong here and you can certainly merge it if you need it.

Just one question, I am not quite sure what you intend to do with it. The abstract environment is used to store the value of bindings. For example given

```
a <- 2
```

You could use the following Abstract env
```
class IntConstant {
    int num;
    ...
}
AbstractEnvironment<IntConstant> env;
env[Rf_install("a")] = IntConstant(2);
```
The key is supposed to be the symbol `a` (we use the symbol from R that's why the key is a SEXP).

What is your usecase where you need to index by something else than a symbol?

petamaj:
I would agree with all that Oli says, while I do not see the need for different keys I am not particularly opposed have them templated. I might consider the oder, and provide a default key value, because the vast majority would likely use SEXPs anyways:

`template<typename AVALUE, typename KEY = SEXP> ...`

aviralg:
In my analysis, I have to keep track of which promises are forced or not. Each promise has a unique id which is a string. So I need an abstract environment which maps strings to lattice. Similarly I have other maps where keys are not sexp.

o-:
I see, so you basically want to use it for something else than an R environment. I would probably advice to create your own class for your domain. You can start with a copy of AbstractEnvironment but it will likely diverge at some point because the idea with AbstractEnvironment is that it's tailored to R environments (ENVSXP)

aviralg:
It may make sense to rename `AbstractEnvironment` to `AbstractREnvironment` and have an `AbstractEnvironment` which does not assume the key to be a `SEXP`. This renaming, if people agree to it, can be done later. For now, I will create a separate class for my domains.

PR #47, opened 2017-05-03
Analysis reorg

aviralg:
Reorganized the analysis libraries and clients. Only changed folder structure and `#include` declarations.


aviralg:
Merge this after merging PR#46

rir/src/analysis/Printer.h
```diff
@@ -1,22 +1,21 @@
 #pragma once
```
o-: This is not an analysis. Can we move it somewhere else?

rir/src/analysis/cleanup.h
```diff
@@ -1,7 +1,7 @@
 #ifndef RIR_OPTIMIZER_CLEANUP_H
```
o-: This is not an analysis. I think it was well placed in the optimizations folder

rir/src/analysis/cp.h
```diff
@@ -1,8 +1,8 @@
 #ifndef RIR_OPTIMIZER_CP_H
```
o-: again not an analysis

rir/src/analysis/localize.h
```diff
@@ -1,7 +1,7 @@
 #ifndef RIR_LOCALIZE_H
```
o-: modifies the codestrean -> transformation, not an analysis

o- (commented):
Hi aviral. Thanks for the patch. I like moving the stuff to dirs with more intuitive names.

I think we should keep the distinction between analysis and transformation. Basically everything that looks at the code read-only can go to /analysis but everything which changes it goes to something else. /optimization or may /transformation

I guess all the other changes are from clang-format, are they?

rir/src/analysis/Printer.h
petamaj: I would suggest a `tools` or `utils` folder for the things like the printer, because it is neither an optimization under which we filed it before.

aviralg:
After your feedback, we have, as I understand -
1 The analysis framework which provides AbstractState, Dispatcher, ForwardAnalysis etc.
2 Analyzers (strictness, etc)
3 Transformers which just modify code
4 Optimizers which do some actual optimization.

I re-re-organized in the last commit.
Yes, other changes seem to be from clang-format. Is this different from what you expect from the clang-format file in the top-level of this project? I ask because there is a possibility that my emacs is picking some other clang-format config.

o-:
> Is this different from what you expect from the clang-format file in the top-level of this project?

not at all, looks good. Just wanted to double check


Ok, sorry I know now we are entering the bike shedding phase. But analysis vs. analyzers is a bit strange. Also transformers sounds a bit strange.

Also I prefer singular for the folders because it reads more natural. `optimization/cp` selects *the* cp optimization. It's a similar argument to why `element[i]` is better than `elements[i]` because you select one element.

My proposal is

```
analysis_framework/
analysis/
optimization/
utils/
```

I don't think we need to distinguish between optimization and transformation. That's a very fuzzy line.

But let's all agree first on the categories before we waste your time anymore (@petamaj)

aviralg:
Ok, sounds good. Next commit will fix this.

mhyee:
Sorry for the nitpick after you already committed, but shouldn't it be the singular `util` and not `utils`?

o-:
> Sorry for the nitpick after you already committed, but shouldn't it be the singular util and not utils?

at least I could live with it. no rule without exception. Don't ask me why, but utils sounds better :) feel free to merge either version. (or rather discuss with @mhyee which one)

o- (approved):

aviralg:
Somehow `utils` seemed so natural that I never realized that its an exception to this proposal. The point of this PR is an intuitive organization of the program analysis and optimization code. I will go ahead and merge the changes.

petamaj:
Thanks!

PR #48, opened 2017-08-14
Clean up Function/SEXP casts and related ops

mhyee:
~~Maybe more commits to come.~~

This is a refactor to simplify/rename some of the conversion/cast code. I also searched for places that were doing things like `(Function*)INTEGER(s)` and replacing them with function calls, e.g. `sexp2function(s)`.

Assuming the tests pass, this is ready for review/merging. I'll open separate PRs (but reusing the same branch) for further work.

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -70,7 +70,7 @@ REXPORT SEXP rir_markOptimize(SEXP what) {
         return R_NilValue;
     SEXP b = BODY(what);
     isValidFunctionSEXP(b);
-    Function* fun = (Function*)INTEGER(b);
+    Function* fun = sexp2function(b);
     fun->markOpt = true;
     return R_NilValue;
 }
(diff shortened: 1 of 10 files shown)


rir/src/api.cpp
```diff
@@ -81,14 +81,14 @@ REXPORT SEXP rir_eval(SEXP what, SEXP env) {
         f = isValidClosureSEXP(what);
     if (f == nullptr)
         Rf_error("Not rir compiled code");
-    return evalRirCode(functionCode(f), globalContext(), env, 0);
+    return evalRirCode(function2code(f), globalContext(), env, 0);
```
o-: not sure if I like that name, since a function has multiple code objects. so maybe bodyCode(f)?

o- (approved):
thanks!

mhyee:
Thanks, will merge once CI passes.

PR #50, opened 2017-09-05
Backward analysis driver

JanJecmen:
This was written last fall for the reading group class. I moved it from its old branch and added the class to retrieve analyses results (which was not implemented).
(the liveness analysis was just a "toy" to test the driver)

### DIFF
--- a/rir/src/analysis/liveness.h
+++ b/rir/src/analysis/liveness.h
@@ -27,13 +27,8 @@ public:
 
     bool mergeWith(ASet const & other) {
         // This is just set union
-
         auto originalSize = variables.size();
-
-        for (auto const & var : other.variables) {
-            variables.insert(var);
-        }
-
+        variables.insert(other.variables.begin(), other.variables.end());
         return originalSize != variables.size();
     }
 
@@ -52,56 +47,39 @@ public:
     }
 
     std::unordered_set<SEXP> variables;
-
 };
 
 template <typename AVALUE>
-class LivenessAbstractState : public State
-{
-public:
+class LivenessAbstractState : public State {
+  public:
     
     LivenessAbstractState() = default;
 
     LivenessAbstractState(LivenessAbstractState const &) = default;
-    
-    virtual ~LivenessAbstractState()
-    {}
-    
+
+    virtual ~LivenessAbstractState() {}
+
     /** Creates a deep copy of the state and returns it.
      */
-    virtual LivenessAbstractState * clone() const
-    {
+    virtual LivenessAbstractState* clone() const {
         return new LivenessAbstractState(*this);
     }
 
     /** Merges the information from the other state, returns true if changed.
      */
-    virtual bool mergeWith(const State *other)
-    {
+    virtual bool mergeWith(const State* other) {
         return this->state.mergeWith(dynamic_cast<const LivenessAbstractState*>(other)->getState());
     }
-    
-    const ASet& getState() const
-    {
-        return state;
-    }
-    
-    void setState(const ASet& s)
-    {
-        state = s;
-    }
-    
-    void insert(SEXP e)
-    {
-        state.variables.insert(e);
-    }
-    
-    void erase(SEXP e)
-    {
-        state.variables.erase(e);
-    }
-    
-private:
+
+    const ASet& getState() const { return state; }
+
+    void setState(const ASet& s) { state = s; }
+
+    void insert(SEXP e) { state.variables.insert(e); }
+
+    void erase(SEXP e) { state.variables.erase(e); }
+
+  private:
     
     AVALUE state;
 };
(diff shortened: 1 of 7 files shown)


rir/src/analysis/liveness.h
```diff
@@ -0,0 +1,158 @@
...
+    static ASet const & Absent() { return bottom(); }
+
+    bool mergeWith(ASet const & other) {
+        // This is just set union
+
+        auto originalSize = variables.size();
+
+        for (auto const & var : other.variables) {
+            variables.insert(var);
+        }
```
o-: `variables.insert(other.begin(), other.end())` is more idiomatic

rir/src/analysis/liveness.h
```diff
@@ -0,0 +1,158 @@
...
+
+    LivenessAbstractState(LivenessAbstractState const &) = default;
+    
+    virtual ~LivenessAbstractState()
+    {}
+    
+    /** Creates a deep copy of the state and returns it.
+     */
+    virtual LivenessAbstractState * clone() const
+    {
```
o-: uhm I think we usually put the brackets on the same line. Also, if you install `clang-format` there should be a commit hook that fixes your code style on the lines you try to commit.

rir/src/analysis_framework/analysis.h
```diff
@@ -254,6 +254,278 @@ class ForwardAnalysisIns : public ForwardAnalysisFinal<ASTATE> {
...
+                if (ins.isJmp()) {
+                    JmpT l = ins.immediate.offset;
+                    jumpOrigins_[l].push_back(it);
+                }
+                // count merge points
+                if (isBasicBlockEnd(it)) {
+                    ++mpCount;
+                }
+                // add entry point to working list
+                // this assumes that the last instruction is always ret_
```
o-: does that assumption currently hold? also you make this assumption in various places. Maybe better have a isExitPoint() helper, so we can change it later?

rir/src/analysis_framework/analysis.h
```diff
@@ -254,6 +254,278 @@ class ForwardAnalysisIns : public ForwardAnalysisFinal<ASTATE> {
...
+    std::unordered_map<CodeEditor::Iterator, State*> mergePoints_;
+
+
+private:
+    bool shouldFollowJumpFrom(CodeEditor::Iterator ins) {
+        State*& stored = mergePoints_[ins];  // first call to [] initializes to nullptr
+        if (stored == nullptr) {
+            stored = currentState_->clone();
+            return true;
+        } else {
```
o-: if you don't write the else, it looks nicer (at least to my eyes) you additionally avoid a potential compiler warning (not returning a bool in default case).

rir/src/analysis_framework/analysis.h
```diff
@@ -254,6 +254,278 @@ class ForwardAnalysisIns : public ForwardAnalysisFinal<ASTATE> {
...
+
+                    // merge all origin points to this label and continue to prev
+                    for (auto& origin : jumpOrigins_[cur.immediate.offset]) {
+                        if (shouldFollowJumpFrom(origin)) {
+                            q_.push_front(origin);
+                        }
+                    }
+                    // if prev is unconditional jump elswhere or return, terminate branch
+                    BC prev = *(currentIns_ - 1);
+                    if ((prev.is(Opcode::br_) && code_->target(prev) != currentIns_) ||
```
o-: that seems a bit brittle and is prone to break as soon as we add another instruction that unconditionally jumps...

o- (commented):
In general I like the patch.

What worries me is a bit the iterator logic. By now we already have it implemented twice (once in the forward driver and once in the codeEditor) and now the backwards driver has again hardcoded `bc == br_` type of iterator logic. This will bite us at some point...

One solution would be to only ever use one logic. For example the backwards driver could use the iterator from the CodeEditor to discover the graph and then remember the jumps, but that is probably too slow.

So what I would love to see is a highlevel API on the instructions that all those can use. Maybe `next(bc, pc) -> list<pc>` is already enough? We can implement most with this. Eg.
```
isExit() { return next(pc, bc).size() == 0 }
```

o-:
btw. regarding factoring out the iterator logic: I am ok with merging this now and then clean it up later.

JanJecmen:
For now I changed the details you mentioned. I'd say no need to merge yet, I will first try the iterator logic.

o-:
I like it very much, thanks. Personally I don't find it over the top.

JanJecmen:
@o- What about this PR? Can I merge it?

Also in the future, maybe the CodeEditor could provide some higher-lever control-flow graph queries? (like getting the possible previous basic blocks / instructions) That way this would all be in one place... But for now maybe this is enough?

o-:
yes, of course

PR #52, opened 2017-09-13
Fix cleanup for dup_

JanJecmen:
The set_shared_ instruction potentially follows the dup_ (not the pop_).


o-:
ah nice catch

JanJecmen:
yea, but we still allocate a lot - the same as before it seems :(

o-:
yeah, that would have been too easy :)

PR #53, opened 2017-09-18
Fix cleanup pass

JanJecmen:
This skipped patterns such as guard X; guard X...


o-:
ups.

I would suggest that we merge trivial bugfixes right away. Others can still look at the (closed) PR after the fact...

Issue #49, opened 2017-08-30
Implement vtables (for different calling conventions)


mhyee:
Opening this for tracking/discussion.

**On Wed, Aug 30, 2017 at 8:31 AM, @o- wrote:**
>     The main idea for our next step was to support different calling
>     conventions (CC). As a first example we would like to have support for
>     value (instead of promise) CC.
> 
>     The idea is that a caller, who knows that the arguments are already
>     evaluated can call by value.
> 
>     For example "f(a+b)" or "f( (a) )"  (because the `(`-function is eager).
>     In this case we know the arg is a forced promise, so we might as well
>     pass it by value. Special care will be necessary to support reflection
>     in the callee, but that's another topic.
> 
>     As a first step we decided to introduce a layer of indirection in our
>     closures. Now the idea is that we have a vtable that can hold multiple
>     functions:
> 
>     closure
>        |---> formals
>        |---> env
>        |---> vtable
>                | ----> compiled_bc1
>                | ----> compiled_bc2
>                | ...
>                | ----> compiled_bcn
> 
>     the existing "call" bytecode will pick the first entry. And we'll add a
>     new bytecode that can pick other entries.

This issue is only for the implementation of the vtables. For reference, here's the plan for next steps (which will be tracked in a separate issue):

>     When that is ready the next step will be to extend the compiler to
>     support compiling under the assumption that all arguments are values
>     passed on the stack. and after that we'll have to find a solution for
>     reflection.
> 
>     even more advanced CC might come in the future. for example type
>     signature (akin julia dispatch), or compilation under assumptions (for
>     example "function compiled under the assumption that + is plus)".

mhyee:
I've run into an issue.

Currently, our plan is to store the vtable in the closure's body:
```
closure
   |---> formals
   |---> env
   |---> body (actually a vtable)
           | ----> compiled_bc1
           | ----> compiled_bc2
           | ...
           | ----> compiled_bcn
```

(Each function has a pointer to the closure, so it can access the vtable it belongs to. Note that as we compile more versions of a function, the vtable may grow, so we have to allocate a larger one and update the body pointer.)

However, this breaks the API. `body` should return the function body, not the vtable:

```
> f <- function(x) x^5
> e <- expression(y <- x^2, return(y)) # or a list
> body(f) <- as.call(c(as.name("{"), e))
> f
function (x)
{
    y <- x^2
    return(y)
}
> f(8)
[1] 64
> body(f)
<external code: 0x3562038>
[...]

PR #54, opened 2017-09-19
Add vtables

mhyee:
Implements #49.

To support multiple (compiled) versions of a function, we introduce vtables. The body of a closure is a vtable (instead of a function):

    closure
       |---> formals
       |---> env
       |---> body (actually a vtable)
               | ----> compiled_bc1
               | ----> compiled_bc2
               | ...
               | ----> compiled_bcn

Currently, when a closure is called, the first function in the vtable is selected.

This is only the first step, but I'd like to get the code reviewed and merged. There will be some performance regression, as the call instruction cannot select an optimized function version to call.

There are a number of things we should discuss, probably in a meeting. For now, I left a bunch of TODOs.

I separated the work into multiple commits, to make it easier to review. I'll squash everything before merging.

### DIFF
--- a/rir/src/interpreter/interp.c
+++ b/rir/src/interpreter/interp.c
@@ -310,21 +310,33 @@ SEXP rirCallTrampoline(RCNTXT* cntxt, Code* code, SEXP env, unsigned nargs,
 static SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
                            unsigned nargs, Context* ctx) {
 
-    SEXP body = BODY(callee);
-    Function* fun = sexp2function(body);
+    SEXP vtableStore = BODY(callee);
+    DispatchTable* vtable = sexp2dispatchTable(vtableStore);
+    SEXP funStore = vtable->entry[0];
+    Function* fun = sexp2function(funStore);
 
     static bool optimizing = false;
 
+    // NOTE(mhyee): The introduction of a dispatch table for different function
+    // versions changes many things. This function handles a closure call, and
+    // will optimize the closure function and update the linked list of
+    // functions, so that subsequent closure calls pick the most optimized
+    // version.
+    // This is no longer true with the dispatch table, where the first function
+    // is always selected, unless specified by the call instruction.
+    // For now, to keep the changes simpler, we'll just optimize the function
+    // once and add it to the dispatch table. But the call will still select
+    // the original, unoptimized version.
+
     if (!optimizing &&
         !fun->next
-         /* currently there is a bug if we reoptimize a function twice:
-          *  Deopt ids from the first optimization and second optimizations
-          *  will be mixed and the deoptimizer always only goes back one
-          *  optimization level!
-          *  To avoid this bug we currently only optimize once
-          */
-        &&
-        !fun->origin) {
+        /* currently there is a bug if we reoptimize a function twice:
+         *  Deopt ids from the first optimization and second optimizations
+         *  will be mixed and the deoptimizer always only goes back one
+         *  optimization level!
+         *  To avoid this bug we currently only optimize once
+         */
+        && !fun->origin) {
         Code* code = bodyCode(fun);
         if (fun->markOpt ||
             (fun->invocationCount == 1 && code->perfCounter > 100) ||
@@ -333,21 +345,50 @@ static SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
             optimizing = true;
 
             Function* oldFun = fun;
-            SEXP oldBody = body;
-            cp_pool_add(ctx, oldBody);
-            body = globalContext()->optimizer(callee);
+            SEXP oldFunStore = funStore;
+            cp_pool_add(ctx, oldFunStore);
+            funStore = globalContext()->optimizer(funStore);
 
-            // TODO: there might be promises with references to the old code!
-            // Therefore we keep it around.
-            // TODO: first I tried to use R_PreserveObject, but it did not work
-            // for large vectors, not sure why, need to investigate
-            cp_pool_add(ctx, body);
+            oldFun->next = funStore;
+            fun = sexp2function(funStore);
+            fun->origin = oldFunStore;
 
-            fun->next = body;
+            // Add the new function to the vtable.
+            if (vtable->length < TRUELENGTH(vtableStore)) {
+                vtable->entry[vtable->length++] = funStore;
+            } else {
+                // Allocate a new, larger vtable.
+                DispatchTable* oldVtable = vtable;
+                SEXP oldVtableStore = vtableStore;
 
-            SET_BODY(callee, body);
-            fun = sexp2function(body);
-            fun->origin = oldBody;
+                size_t capacity = TRUELENGTH(oldVtableStore) * 2;
+                size_t size = sizeof(DispatchTable) +
+                              (capacity * sizeof(DispatchTableEntry));
+                vtableStore = PROTECT(Rf_allocVector(EXTERNALSXP, size));
+                vtable = sexp2dispatchTable(vtableStore);
+
+                // Initialize the new vtable, copying old entries over.
+                SET_TRUELENGTH(vtableStore, capacity);
+                vtable->magic = DISPATCH_TABLE_MAGIC;
+                vtable->length = oldVtable->length + 1;
+
+                for (size_t i = 0; i < oldVtable->length; i++) {
+                    vtable->entry[i] = oldVtable->entry[i];
+                }
+
+                // Insert the new function version.
+                vtable->entry[oldVtable->length] = funStore;
+
+                // NULL out the remaining entries.
+                for (size_t i = vtable->length; i < capacity; i++) {
+                    vtable->entry[i] = NULL;
+                }
+
+                // Update the closure with the new vtable.
+                SET_BODY(callee, vtableStore);
+
+                UNPROTECT(1);
+            }
 
             fun->invocationCount = oldFun->invocationCount + 1;
             fun->envLeaked = oldFun->envLeaked;
@@ -497,7 +538,7 @@ SEXP doCall(Code* caller, SEXP callee, unsigned nargs, unsigned id, SEXP env,
         // if body is INTSXP, it is rir serialized code, execute it directly
         SEXP body = BODY(callee);
         if (TYPEOF(body) == EXTERNALSXP) {
-            assert(isValidFunctionSEXP(body));
+            assert(isValidDispatchTableSEXP(body));
             result =
                 rirCallClosure(call, env, callee, argslist, nargs, ctx);
             UNPROTECT(1);
@@ -637,7 +678,7 @@ SEXP doCallStack(Code* caller, SEXP callee, size_t nargs, unsigned id, SEXP env,
         // if body is INTSXP, it is rir serialized code, execute it directly
         SEXP body = BODY(callee);
         if (TYPEOF(body) == EXTERNALSXP) {
-            assert(isValidFunctionSEXP(body));
+            assert(isValidDispatchTableSEXP(body));
             res = rirCallClosure(call, env, callee, argslist, nargs, ctx);
             UNPROTECT(1);
             break;
@@ -755,7 +796,7 @@ SEXP doDispatchStack(Code* caller, size_t nargs, uint32_t id, SEXP env,
             // if body is INTSXP, it is rir serialized code, execute it directly
             SEXP body = BODY(callee);
             if (TYPEOF(body) == EXTERNALSXP) {
-                assert(isValidFunctionSEXP(body));
+                assert(isValidDispatchTableSEXP(body));
                 res =
                     rirCallClosure(call, env, callee, actuals, nargs, ctx);
                 break;
@@ -868,7 +909,7 @@ SEXP doDispatch(Code* caller, uint32_t nargs, uint32_t id, SEXP env,
             // if body is INTSXP, it is rir serialized code, execute it directly
             SEXP body = BODY(callee);
             if (TYPEOF(body) == EXTERNALSXP) {
-                assert(isValidFunctionSEXP(body));
+                assert(isValidDispatchTableSEXP(body));
                 res = rirCallClosure(call, env, callee, actuals, nargs, ctx);
                 break;
             }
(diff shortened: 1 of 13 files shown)


rir/src/interpreter/interp.c
```diff
@@ -310,21 +310,33 @@ SEXP rirCallTrampoline(RCNTXT* cntxt, Code* code, SEXP env, unsigned nargs,
...
+    // version.
+    // This is no longer true with the dispatch table, where the first function
+    // is always selected, unless specified by the call instruction.
+    // For now, to keep the changes simpler, we'll just optimize the function
+    // once and add it to the dispatch table. But the call will still select
+    // the original, unoptimized version.
+
     if (!optimizing &&
         !fun->next
-         /* currently there is a bug if we reoptimize a function twice:
```
mhyee, line 320: Looks like `clang-format` fixed the indentation.

As a side note, hopefully this bug will be fixed when we implement deoptimization between versions.

rir/src/interpreter/interp.c
```diff
@@ -333,21 +345,56 @@ static SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
...
             // for large vectors, not sure why, need to investigate
-            cp_pool_add(ctx, body);
+            cp_pool_add(ctx, funStore);
+
+            oldFun->next = funStore;
+            fun = sexp2function(funStore);
+            fun->origin = oldFunStore;
+
+            // Add the new function to the vtable.
+            if (vtable->length < TRUELENGTH(vtableStore)) {
```
mhyee, line 357: This is the trickiest bit of code, because we have to re-allocate the table, copy pointers over, and update pointers.

I *think* it's fine, but I've stared at this for so long that I'm no longer sure. So more eyeballs here would help.

Because the tests don't really cover the case (only one optimization is done, so the table of size 1 grows to a table of size 2 and that's it), I hacked the code to compile a new function version *every* time a closure is called. This passed the RIR tests, and I tried running the GNU R tests. It ate up all my RAM, so I killed it, but it didn't segfault.

rir/src/interpreter/interp.c
```diff
@@ -1544,13 +1591,7 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP env, unsigned numArgs) {
             SEXP body = ostack_at(ctx, 1);
             SEXP formals = ostack_at(ctx, 2);
             res = allocSExp(CLOSXP);
-
-            assert(isValidFunctionSEXP(body));
```
mhyee, line 1548: This code updates the closure body to use the most optimized version of the function. However, this no longer makes any sense, since the body of a closure is a vtable.

rir/src/interpreter/interp.h
```diff
@@ -18,8 +18,6 @@ SEXP rirExpr(SEXP f);
 
 SEXP rirEval_f(SEXP f, SEXP env);
 
-SEXP rirExpr(SEXP f);
```
mhyee, line 21: This was already declared on line 17.

rir/src/interpreter/interp_data.h
```diff
@@ -381,11 +383,17 @@ typedef struct Function {
     unsigned markOpt : 1;
     unsigned spare : 28;
 
+    // TODO(mhyee): remove origin and next
```
mhyee, line 386: I tried removing `origin` and `next` in this patch, but it's actually really invasive. It would break deoptimization and the [hack](https://github.com/reactorlabs/rir/pull/54/files#diff-ddb2b0e78b44bcab154f459281266a97R333) that prevents optimizing a function twice.

o-: uh, ok, too bad

rir/src/interpreter/interp.c
```diff
@@ -333,21 +345,56 @@ static SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
...
+                SET_TRUELENGTH(vtableStore, capacity);
+                vtable->magic = DISPATCH_TABLE_MAGIC;
+                vtable->length = oldVtable->length + 1;
+
+                for (size_t i = 0; i < oldVtable->length; i++) {
+                    vtable->entry[i] = oldVtable->entry[i];
+                }
+
+                // Insert the new function version.
+                vtable->entry[oldVtable->length] = funStore;
```
o-, line 380: uh, if you add the optimized version to the, it won't ever be executed, not?

o-: also, shouldn't the length be increased here

JanJecmen: This uses the old vtable length, the new one is already incremented (l. 379)

mhyee: Yeah, the optimized version is never executed. This is intentional, because I didn't want to wait until we added a new call instruction, and I also didn't want to rip out the optimizer completely.

Maybe I should change the index/increment so that it's a bit clearer that I'm using the old length...

o-: > Yeah, the optimized version is never executed

huh, but that's a regression then and has nothing to do with the call instruction, not?

because the old code just replace the existing function with the optimized function and henceforth we'd just use the optimized one.

o-: so basically it used to be that the optimized function replaces the existing one.

now the optimized version is just added to the dispatch table, but never used.

mhyee: Correct. The optimized version won't be used until we implement a mechanism for calling a particular version.

I considered reordering the table, so that the optimized version is the first entry. But in general, I think we want the order to be consistent, and have the first entry be the original, most general version.

If you're concerned about the performance regression (do we know how big it would be? We're only optimizing once), I can reorder the table as part of this merged-but-in-progress work, and then change it back (i.e. not change the order when we have a new version) when we have a new call instruction.

o-: Well, the baseline version has profiling and tracing enabled. So you should see the same method getting optimized over and over again (since you never replace it by the optimized one)

o-: so this is a (1) a perf issue and (2) a testing issue, because we never execute optimized code we will not notice if we break the optimizer

JanJecmen: Maybe we don't need the unoptimized version? Because the only optimization that is done when compiling a function for the first time is the cleanup pass, and that should always be safe, right?

It is only later in reoptimizeFunction that some more stuff happens (although it is currently just the inliner)

But this could of course change in the future...

o-: > So you should see the same method getting optimized over and over again

ah that is probably not true, due to the `!fun->next` test before optimizing. But that's a bit artificial, since prev/next should really go away.

o-: > I can reorder the table as part of this merged-but-in-progress work

also I think we might not have the same idea about signatures.

imho the signatures are about things that change the semantics of a version. e.g. the version with the "strict" signature will break when called with other stuff.

but what our "optimizer" (edit: re-optimizer) currently produces is something which has the same signature as the original version and does not have differing semantics. so for all intents and purposes it is drop in replacement of the source version and I don't see why it should not be used by default (regardless of any future call instructions we have)

o-: > It is only later in reoptimizeFunction that some more stuff happens 

I think that is what we are currently talking about, not?

(edit: i don't think we store the unoptimized version anywhere. sorry, when I say unoptimized, I mean "basline" ie. optimized, but not re-optimized :) )

JanJecmen: Yes, my point was that since we only do cleanup when compiling a new function, we maybe don't need the "unclean" version.

JanJecmen: Oh my bad, you are right

o-: so the way I see it it is kinda orthogonal issue. on the x-axis we have signature and on the y-axis we have optimization level.

When we call a function we pick the most optimized version of the matching signature. Picking the signature must be sound for the whole duration (its not an assumption, because we match it to the actual arguments). So the x-coordinate stays fix.

The different optimization levels on the other hand represent speculative optimizations. If we need to deopt while executing a function we lower the y-coordinate, so to speak.

Note: I am not proposing a 2d dispatch table :) that's just to make it easier to talk about it.

mhyee: We are in agreement. When I say a "mechanism for calling a particular [function] version," I mean that it uses some identifier that distinguishes the signature (or calling convention, but maybe that's too specific) and optimization level, and possibly other metadata. This identifier would be the (x,y) coordinates in your description. (In the call, I might have misspoke and called this identifier the "signature.")

The problem is that this PR is intentionally incomplete. There are these leftovers, such as `origin`, `prev`, and `next`, and the fact that an optimized version is never called.

I could have postponed this PR until all the work is complete (profiling, optimizing for a calling convention, calling a particular version, handling deopt). But then the code is sitting around on the side and getting out of sync (especially since my attention is split across other things). (In fact, there's already a merge conflict.) So I thought it would be better to submit at a reasonable intermediate point.

Now the question is, what is a reasonable intermediate point? I had actually discussed with @janvitek about using the optimized version. I considered swapping the order so that the optimized version is the first (and therefore default) entry, or adding a field with a default offset. But the deoptimizer would also need to be patched, and then all of this reverted when the other things are implemented. Jan suggested it would be simpler to just add the new version to the end of the table and not call it (for now), so that's what I implemented.

o-: > But the deoptimizer would also need to be patched

I am certainly fine with submitting an intermediate version. But this is the part I do not understand. If you let prev/origin in place then how come the doptimizer does not work? Sorry, maybe I just don't remember, since I haven't looked at rir myself in a long time. But it seems to me that you should be able to just swap the order and it should work and I fear that if it doesn't that there is a bug.

o-: I am not complaining that it's not finished, it's more that I do not understand the scoping. On the one hand you let prev/next in place which is only used by the deoptimizer, but on the other hand you disable optimized code, hence the deoptimizer is not exercised at all.

o-: btw, I tried to look at it, but I get

```
failed test rir_regression.R:
   /home/user/Documents/rir/external/custom-r/bin/R -f /tmp/r-test.gzGmu2

R: ../../rir/src/interpreter/interp.c:1382: evalRirCode: Assertion `false && "wrong or unimplemented opcode"' failed.
```

in a debug build.

JanJecmen: Sorry about the merge, I just wanted to look at the conflict :/
Anyway, the tests pass for me

rir/src/interpreter/interp.c
```diff
@@ -333,21 +345,56 @@ static SEXP rirCallClosure(SEXP call, SEXP env, SEXP callee, SEXP actuals,
...
+            SEXP oldFunStore = funStore;
+            cp_pool_add(ctx, oldFunStore);
+            funStore = globalContext()->optimizer(funStore);
 
             // TODO: there might be promises with references to the old code!
             // Therefore we keep it around.
             // TODO: first I tried to use R_PreserveObject, but it did not work
             // for large vectors, not sure why, need to investigate
-            cp_pool_add(ctx, body);
+            cp_pool_add(ctx, funStore);
```
o-: btw, this should no longer be necessary, since all versions are now reachable by the gc through the dispatch table

rir/src/interpreter/interp.c
o-: ah crap, I did not update gnur. we really need a solution for this.

mhyee: > But it seems to me that you should be able to just swap the order and it should work and I fear that if it doesn't that there is a bug.

I don't remember if I tried this and it was messy, or if I didn't try it because I wanted to keep it simple.

> I am not complaining that it's not finished, it's more that I do not understand the scoping. On the one hand you let prev/next in place which is only used by the deoptimizer, but on the other hand you disable optimized code, hence the deoptimizer is not exercised at all.

Here's my reasoning: I enabled the optimizer, because I needed a way to grow the vtable. I also didn't want to comment out a bunch of code. I left prev/next because removing them would be extremely invasive. I don't call the optimized code, because it's simpler.

I'll take a look at swapping the order.

@JanJecmen That's OK, I was planning to rebase and mess around with the git history.

rir/src/interpreter/interp.c
mhyee: OK, thanks.

rir/src/interpreter/interp.c
mhyee: Ah, I see now. If I rearrange the vtable so that the optimized function is first, then I need to fix deopt.

Before this PR, when a function is optimized, a new version is compiled and added to the linked list. Each version has a pointer to `origin`, the baseline version. The closure is updated to point to the new function version. When a function `f`  is deoptimized, the closure is updated to point to the baseline, `f->origin`.

With this PR, and a local change, new versions are prepended to the vtable, so that the optimized version is the default version that is called. But when a function `f` is deoptimized, yes, `f->origin` is the baseline, but now the closure points to a dispatch table and not a function version. So I can't just update the body of the closure.

I could just swap the order of functions in the vtable again, so that the first entry is now the baseline. But this assumes we only ever have two versions. This is currently true (but may not be in the future), so a more robust solution would involve scanning the vtable to compare pointers with `f->origin` and then making the swap...

I would expect tests to blow up immediately, but this isn't happening, so I'll take a look after lunch...

o-: ok, I see

I don't want to block this. If you have a plan in which order to do the steps, then go ahead. I mean the PR is fine, I was just not sure about the scope.

btw. maybe it would make sense to keep slot0 reserved for the "default" method? we don't have to swap the entries, we could just duplicate. eg. initially:  "base, base" and then "opt, base, opt".

mhyee: I'd like to merge as-is (well, after addressing the comment below and rebasing). I don't want to add machinery that I'd remove later.

I don't have a *specific* plan for what to do next, but this is what I have in mind:

1. Generate different versions of a function (maybe optimizations based on some profiling data). This includes saving the "identifier" as some function metadata, which describes the calling convention, optimization level, etc.
2. Add a mechanism for calling a particular function version: given an identifier, scan the dispatch table for that function. Obviously, this will be slow. Also, the deopt mechanism needs to be updated.
3. Improve performance, e.g. with an inline cache.

(I won't be surprised if many things go wrong and there's a 1a, 1b, 1c, etc.)

I guess I could swap 1 and 2: only consider "optimization levels" for different versions, and be able to call and deoptimize. Then, after this is done, we can look at other kinds of versions, e.g. calling convention.

mhyee:
@o- If there are no objections to [the plan I outlined](https://github.com/reactorlabs/rir/pull/54#discussion_r143041391) then I'd like to merge this today.

o-:
sure

mhyee:
Thanks! I rebased and will merge once the tests pass.

PR #58, opened 2017-10-23
Formals

JanJecmen:
Default values for formal arguments are now also compiled.

They are written into the `Function` object as separate `Code` objects with a flag `isFormalPromise` set to true.

`CodeEditor` now takes care to keep these `Code`s in the function.

The interpreter uses these after it matches all actual arguments to a call and tries to see if any of the missing ones have a default value. If so, it wraps the `Code` to a promise and puts it into the arglist.

The compiler now does not change the formals of a closure, it only compiles them.

### DIFF
--- a/rir/src/utils/CodeHandle.h
+++ b/rir/src/utils/CodeHandle.h
@@ -21,6 +21,8 @@ class CodeHandle {
         code = new (insert) Code;
 
         code->magic = CODE_MAGIC;
+        code->_padding1 = 0xabababab;
+        code->attributes = R_NilValue;
         code->header = offset;
         code->src = src_pool_add(globalContext(), ast);
         code->codeSize = codeSize;
--- a/rir/src/utils/CodeHandle.h
+++ b/rir/src/utils/CodeHandle.h
@@ -17,7 +17,7 @@ class CodeHandle {
     Code* code;
 
     CodeHandle(SEXP ast, unsigned codeSize, unsigned sourceSize,
-               unsigned callSiteLength, unsigned offset, void* insert, bool isFormal) {
+               unsigned callSiteLength, unsigned offset, void* insert, bool isDefaultArg) {
         code = new (insert) Code;
 
         code->magic = CODE_MAGIC;
@@ -28,7 +28,7 @@ class CodeHandle {
         code->skiplistLength = skiplistLength(sourceSize);
         code->callSiteLength = callSiteLength;
         code->perfCounter = 0;
-        code->isFormalPromise = isFormal;
+        code->isDefaultArgument = isDefaultArg;
     }
 
     CodeHandle(Code* code) : code(code) { assert(code->magic == CODE_MAGIC); }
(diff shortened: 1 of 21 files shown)


o-:
sorry, I did not manage to look at it today. Will do it tomorrow

rir/src/utils/FunctionHandle.h
```diff
@@ -48,23 +51,27 @@ class FunctionHandle {
...
+        FunctionHandle res(store);
+        res.storeOwner_ = true;
+
+        return res;
     }
 
-    FunctionHandle(SEXP store)
-        : store(store), payload(INTEGER(store)), capacity(XLENGTH(store)),
-          function((Function*)payload) {
+    explicit FunctionHandle(SEXP store)
```
o-, line 60: I like your style. explicit is good :)

rir/src/utils/FunctionHandle.h
```diff
@@ -48,23 +51,27 @@ class FunctionHandle {
...
+        : storeOwner_(false), store(store), payload(INTEGER(store)),
+          capacity(XLENGTH(store)), function((Function*)payload) {
         assert(function->magic == FUNCTION_MAGIC);
         assert(function->size <= capacity);
     }
 
     ~FunctionHandle() {
-        R_ReleaseObject(store);
+        if (storeOwner_)
+            R_ReleaseObject(store);
```
o-, line 69: so did you find out, why sometimes the handle is the owner and sometimes not? is that really a good design, or could it be improved at the client side?

for example, I could see an api, where you can either:
pass a store to the constructor   `FunctionHandle(store)` in which case it is *not* owned by the handle. Or let the handle create the store, eg. `FunctionHandle::create()`, in which case the store will be internal to the function handle and also owned by it.

rir/src/utils/CodeHandle.h
```diff
@@ -28,6 +28,7 @@ class CodeHandle {
         code->skiplistLength = skiplistLength(sourceSize);
         code->callSiteLength = callSiteLength;
         code->perfCounter = 0;
+        code->isFormalPromise = isFormal;
```
o-: I don't really like the name "formal promise". It is kind of confusing. can we have something more meaningful, like "(compiled) default arguments", or such.

o- (commented):
Hey, thanks for the patch. This looks like a good solution now. I have only two nitpicks with the PR:

1. there are unrelated changes in it :)
2. the name "formal promise"

No need to fix 1. Otherwise it looks good to me.

rir/src/utils/FunctionHandle.h
JanJecmen: That is exactly what the code does, only the handle had no way to know in its destructor which way the store was obtained.

The problem I think was that the public constructor was used both implicitly (by the verifier and when retrieving the function from a `CodeHandle`) and explicitly (in the static `FunctionHandle::create` method). Hence I added the `explicit` modifier and made not owning the store the default. If the store is allocated by `create`, it changes the flag in the created handle and it later cleans up appropriately.

The design is ok imo if used in the way you described. You can create a handle from a SEXP and not care about the store, or if you need a new store, you use `create` and then have to take care to protect it after the handle is destroyed.

JanJecmen:
Sorry about the unrelated changes.. I will do them separately next time. The name is confusing, I agree.

o-:
👍

PR #59, opened 2017-10-27
assert that variable lookup does not return a promise

o-:
I just want to make sure that what we get back from a variable lookup
can not be just another promise.


o-:
tbr

PR #60, opened 2017-10-27
do not hoist guards over assignments to the guarded function

o-:


rir/src/optimization/cleanup.h
```diff
@@ -187,8 +187,15 @@ class BCCleanup : public InstructionDispatcher::Receiver {
...
                 // We cannot move the guard across those instructions
-                if (cur.is(Opcode::label) || !cur.isPure() || cur.isReturn())
-                    break;
+                if (cur.is(Opcode::label) || !cur.isPure() || cur.isReturn()) {
+                    if (!cur.is(Opcode::stvar_)) {
+                        break;
+                    }
+                    // stvar that modiefies the guarded value is a barrier,
+                    // but otherwise we can push through
+                    if (Pool::get(cur.immediate.pool) == expected)
```
JanJecmen: Doesn't this compare a symbol (`stvar_`'s immediate) with the actual guarded function (`guard_fun_`'s expected immediate)?

JanJecmen: Also, I think there is a mechanism in R that allows you to have a variable and a function with the same name, and if you try to call a variable, it continues to search for a callable:
```
function() {
  c <- 10
  c(c, c, c)
}
```
So is this even something we can check?

o-: ups, good catch.
unfortunately I can't write a testcase, since currently our guard just aborts the program :)

o-: good point. I didn't think of that.

the idea of this patch is to only hoist a guard above an assignment, if it is not interfering. In your above example the guard would not be hoisted in case c is a function or a different type. In the later case it *could* theoretically be hoisted, since (as you pointed out) the write to c does not affect lookup of c. So this patch is conservative but sound.

o-:
better now?

JanJecmen:
Yes :)

PR #61, opened 2017-10-27
ldvar returns Value or R_MissingArg

o-:
In the case:

  function(a) a

when we force a we get either a value or R_MissingArg

In the case:

  function() a

the same (since we do not know what a is, it could be an argument to an
outer function).

But in the case:

  function(b) { a <- b; a }

a must be a value (if b where R_MissingArg, we would have failed on the
first instruction).

also the result of a function call cannot (afaik) be a missing argument.

Currently we have to mark those as `Any` (which also includes promises),
so we loose a lot of precision.

In this commit I do not include Missing in the normal "t" lattice.
Because we are interested in the opposite (if something is *not* an
argument, I did not find a good way to include it.)

Maybe longterm we could introduce an additional not-t lattice that
records what the value is not, such that we can close the bounds from
both ends...

### DIFF
--- a/rir/src/analysis/dataflow.h
+++ b/rir/src/analysis/dataflow.h
@@ -11,9 +11,29 @@
 namespace rir {
 
 /*
+ *  This dataflow analysis computes various things about stack slots and
+ *  environment variables.
+ *
+ *  First of all there is a lattice of "availability" of a value stored in the
+ *  local variable "t".
+ *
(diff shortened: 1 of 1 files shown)


JanJecmen:
How about this one? Merge?

o-:
so if pir pans out, this is not needed anymore...

PR #62, opened 2017-10-30
Expose SEXPs in RIR objects to R's GC.

JanJecmen:
This patch adds a header to all RIR objects that contains information for the GC about pointers that we want to trace.

Every RIR object (with the exception of Code, since it is not a valid SEXP), now has to have rir_header as its first field and fill it with
1) offset in bytes to the first SEXP to trace
2) number of SEXPs to trace

Then it should make sure that all traced SEXPs are in one continuous block of memory.

For now, this applies to DispatchTable and Function, and to function Signature as soon as we have it.

Also, gnur needs to be patched as well, but I forgot how that was done. For now, I have 
[the patch](https://github.com/reactorlabs/rir/files/1427417/gnur.zip)... the build will fail until we update

### DIFF
--- a/rir/src/interpreter/interp_data.h
+++ b/rir/src/interpreter/interp_data.h
@@ -37,11 +37,11 @@ extern "C" {
 // magic in his vector too...
 #define FUNCTION_MAGIC (unsigned)0xCAFEBABE
 
-#define DISPATCH_TABLE_MAGIC (unsigned)0xBEEF1234
-
 // Offset between function SEXP and Function* struct
 // This is basically sizeof(SEXPREC_ALIGN)
 #define FUNCTION_OFFSET 40
+
+#define DISPATCH_TABLE_MAGIC (unsigned)0xBEEF1234
 
 // Code magic constant is intended to trick the GC into believing that it is
 // dealing with already marked SEXP.
@@ -88,12 +88,37 @@ typedef enum {
  * Aliases for readability.
  */
 typedef SEXP FunctionSEXP;
+typedef SEXP SignatureSEXP;
 typedef SEXP ClosureSEXP;
 typedef SEXP PromiseSEXP;
-typedef SEXP IntSEXP;
+typedef SEXP DispatchTableEntry;
 
 // all sizes in bytes,
 // length in element sizes
+
+
+/** Header for all RIR objects embedded inside an EXTERNALSXP R object.
+ *  This is used to expose SEXPs in RIR objects to R's garbage collector.
+ *
+ *  RIR objects that want to expose some of their internal SEXPs to the
+ *  GC to trace need to place those SEXPs consecutively one after another.
+ *
+ *  gc_area_start is the offset in bytes to the first exposed SEXP,
+ *  relative to the start of the RIR object (i.e. INTEGER(obj)).
+ *
+ *  gc_area_length is the number of exposed SEXPs.
+ */
+typedef struct {
+    uint32_t gc_area_start;  /// First SEXP to be marked by the GC
+    uint32_t gc_area_length;  /// Number of SEXPs to expose to the GC
+
+    // TODO:  Later maybe also add type of the object here and have just
+    //        one EXTERNALSXP with a union of other types.
+    // For now just make sure that this header is in all RIR objects.
+    // The exception is Code, since it is not a valid SEXP in the
+    // first place.
+} rir_header;
+
 
 // ============
 // Please do not change those without also changing how they are handled in the
@@ -375,6 +400,18 @@ INLINE Code* next(Code* c) {
 #pragma pack(push)
 #pragma pack(1)
 typedef struct Function {
+    rir_header info;  /// for exposing SEXPs to GC
+
+    FunctionSEXP origin; /// Same Function with fewer optimizations,
+                         //   NULL if original
+
+    FunctionSEXP next;
+
+    ClosureSEXP closure; /// pointer to Closure
+                         //    which has a pointer to DispatchTable
+
+    SignatureSEXP signature;  /// pointer to this version's signature
+
     unsigned magic; /// used to detect Functions 0xCAFEBABE
 
     unsigned size; /// Size, in bytes, of the function and its data
@@ -386,17 +423,6 @@ typedef struct Function {
     unsigned deopt : 1;
     unsigned markOpt : 1;
     unsigned spare : 28;
-
-    // TODO(mhyee): remove origin and next
-    FunctionSEXP origin; /// Same Function with fewer optimizations,
-                         //   NULL if original
-
-    FunctionSEXP next;
-
-    ClosureSEXP closure; /// pointer to Closure
-                         //    which has a pointer to DispatchTable
-
-    // TODO(mhyee): member for signature
 
     unsigned codeLength; /// number of Code objects in the Function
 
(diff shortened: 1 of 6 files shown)


rir/src/utils/FunctionHandle.h
```diff
@@ -32,25 +32,25 @@ class FunctionHandle {
...
-        function->origin = nullptr;
-        function->next = nullptr;
-        function->closure = nullptr;
-        // TODO(mhyee): signature
         function->codeLength = 0;
         function->foffset = 0;
         function->invocationCount = 0;
         function->markOpt = false;
 
+        R_PreserveObject(store);
```
JanJecmen, line 53: This calls CONS and thus can cause a collection, so the store needs to be filled by then, otherwise it can segfault...

rir/src/utils/FunctionHandle.h
```diff
@@ -32,25 +32,25 @@ class FunctionHandle {
         SEXP store = Rf_allocVector(EXTERNALSXP, initialSize);
         void* payload = INTEGER(store);
 
-        TYPEOF(store) = EXTERNALSXP;
-        R_PreserveObject(store);
-
-        Function* function = new (payload) Function;
```
JanJecmen: Not sure why placement new here?

rir/src/interpreter/interp_data.h
```diff
@@ -387,17 +424,6 @@ typedef struct Function {
     unsigned markOpt : 1;
     unsigned spare : 28;
 
-    // TODO(mhyee): remove origin and next
```
JanJecmen, line 390: From what I understood, we keep them and have a 2d dispatch table, right?

rir/src/utils/FunctionHandle.h
```diff
@@ -89,15 +89,19 @@ class FunctionHandle {
             SEXP newStore = Rf_allocVector(EXTERNALSXP, newCapacity);
             void* newPayload = INTEGER(newStore);
 
-            TYPEOF(newStore) = EXTERNALSXP;
-
             memcpy(newPayload, payload, capacity);
-            memset(payload, 0xee, capacity);
```
JanJecmen, line 95: Instead, just null out the traced pointers

rir/src/utils/FunctionHandle.h
o-: I'd say it is more idiomatic to use placement new here. but of course since Function is a C struct, it should have no effect.

o- (commented):
lgtm!

rir/src/interpreter/interp_data.h
mhyee: Correct.

o-:
to update gnur: just commit the patch to gnur repo, then rerun travis

mhyee:
> to update gnur: just commit the patch to gnur repo, then rerun travis

Specifically, this repo and branch: https://github.com/reactorlabs/gnur/tree/rir-patch-3-3-branch (which should already be checked out to `external/custom-r`, so just commit it and push).

JanJecmen:
Ok thanks, here it is: https://github.com/reactorlabs/gnur/commit/8d208d64bcfd819a7ce8491557359ff40ae21abd
(I just remembered some wild stuff when I needed to change the outputs of some tests this spring)

rir/src/utils/FunctionHandle.h
JanJecmen: Ok, reverted the change.

mhyee (approved):
Looks good! Thanks for this, it makes the GC tracing a lot more general.

Just one minor comment about the gnur patch. Because macros are not hygienic, to be safe, use `__i__` and `__o__` instead of `i` and `o`.

o-:
oh, I didn't look at the gnur patch. yeah definitely scramble some names. This stuff is in a macro, in a macro, in a macro, ...

JanJecmen:
Done, haven't thought of that :(
(just out of curiosity: it's in a block and the variables are declared there, so they should be local to that block and not overwrite anything outside, right?)

o-:
well dc__action__ is a macro and you could e.g. have:
```
#define dc__action__(ptr, dc__extra_)  i = -1
```
probably not that stupid, but dc_extra_ is another macro that's recursively called through dc_action, so that could easily happen somewhere.

o-:
There seems to be an issue with this patch. Travis is not happy. @JanJecmen

JanJecmen:
Oh yes, and he's not alone, I am not happy either (to put it mildly) :(

The patch still adds new function versions into new vtable slots, which is not what we want -- they ought to go to the linked list in the first slot (for now). So I changed this to add to the linked list, but I introduced a bug. My theory is some protect error, but I haven't found it yet.

(behavior: closure gets compiled, then after it's reoptimized, the vtable is updated; then the function gets called once more and after that when it's invoked for the second time, there is just rubbish inside -- I think because the memory got collected and allocated for sth else -- but I don't really see where the function is not protected and there could be a collection...)

When I add the new version to the constant pool, this gets fixed, but that's not what we want to do.

o-:
@JanJecmen hmm two things:
1. <s>I seems that the gnur patch also breaks without compiling new versions. see my other PR that is now red too, even though it does not change anything related</s> forget that. of course I would need to rebase on top of this PR :/

2.
> When I add the new version to the constant pool, this gets fixed, but that's not what we want to do.

I would not be surprised if that is still necessary. We still have the following problem: It can happen that a promise outlives a function. When that happens, there are no pointers to the CLOSXP anymore, but there will still be PROMSXP that point to Code* structs inside an EXTERNALSXP that was only reachable through that CLOSXP. That is a fundamental problem we have right now and didn't solve yet...

UPDATE:
you can try and test my hypothesys by adding an attribute to the promise in mkpromise that points back to the closure, to keep it reachable. (unfortunately this is to slow to keep in there)

JanJecmen:
1. No, I think you are right, the patch changes how EXTERNALSXPs are collected, but the RIR code is not merged yet, so this breaks everything... I was probably too eager to push into gnur

2. Oh damn, that's nasty. So for now just keep every function ever compiled in the constant pool?

I said that we don't put the old function there anymore, but that was incorrect, I missed that the line of code moved instead of being deleted. However, this concerns the newly optimized function...

Here is a thought: maybe the new function is never marked by the GC, because the closure where we put it already is marked, so the collector skips it? Investigating that now...

o-:
> Here is a thought: maybe the new function is never marked by the GC, because the closure where we put it already is marked, so the collector skips it? Investigating that now...

ah, good point. of course!

if you look at e.g. SET_CDR() it hase a write barrier (CHECK_OLD_TO_NEW) that might trigger. of course we have to do the same if we modify the traced pointers inside the dispatchtable/function

JanJecmen:
Here be dragons for me :(
So I made sure to SET_BODY even if just a vtable entry is modified, but this still doesn't fix it.
I think we need something along the lines of SET_EXTERNAL_VTABLE_ENTRY and put the write barrier there. However, this is another gnur change, because FIX_REFCNT and CHECK_OLD_TO_NEW are just macros in memory.c

o-:
btw. I think we should reconsider having gnur as a git submodule in rir. this would allow us to submit PR against changed gnur without breaking tests for other in-flight PR.

The reason is, that submodules are always checked out at a certain commit. Thus you can update gnur and push to that repository, but RIR still refers to the old commit. Then in your PR you can update the submodule reference. The update of the submodule appears as a normal commit in your branch history. When you merge the PR both changes (your rir changes and the submodule update) happen at the same time.

The problem with submodules is that they are sometimes a bit tricky to use. For example if you just go to the repository of the submodule and commit, you will commit into the void, since the submodule is checkout in "headless mode" by default.

If everybody agrees I could convert gnur to a submodule and document the usage...

mhyee:
Yes, please use submodules. Explicit is better than implicit.

Oh, and while you're at it, could you update the default branch on the gnur repo? Right now it's `rllvm` but it should be `rir-patch-3-3-branch`.

(I'm generally wary about overusing submodules as a solution, but it's probably appropriate in this place.)

PR #63, opened 2017-11-02
use submodules and C++ compiler

o-:
converts gnur to a submodule and compiles the whole project with C++ compiler


o-:
This is the most minimal commit to make the interpreter compile with C++. There is a lot of cruft that can be removed now. But I will do it in a separate PR.

@JanJecmen if this PR (due to R as a submodule) simplifies your life with #62 then we can merge this first. If not, I will rebase when it lands.

JanJecmen:
Cool!
I think I have #62 fixed, just running tests locally now to make sure. So I think rebasing after would be better.

README.md
```diff
@@ -40,6 +40,38 @@ Functions compiled to RIR can be inspected using `rir.disassemble`.
...
+
+### Making changes to gnur
+
+R with RIR patches is a submodule under external/custom-r. This is how you edit:
+
+    # Assuming you are making changes in you local RIR branch
+    cd external/custom-r
+    # By default submodules are checked out headless. We use a
+    # branch to keep track of our changes to R, that is based on
+    # one of the R version branches. If you want ot make changes
```
mhyee, line 53: Typo: ot -> to

mhyee (commented):
👍 

Would it be worth adding a check in the pre-commit hook, to make sure the submodule is properly updated?

o-:
> Would it be worth adding a check in the pre-commit hook, to make sure the submodule is properly updated?

I don't understand. what would you check and how would you know if you want to commit an updated ref or not?

mhyee:
I was thinking more along the lines of "someone updated the submodule ref but didn't push the submodule" but maybe that doesn't actually happen often.

I also wonder -- if you're about to commit some changes to `rir`, but the working directory of `external/custom-r` is dirty, doesn't that mean you should commit the submodule and then update the ref?

It's been a long time since I used submodules...

o-:
> I also wonder -- if you're about to commit some changes to rir, but the working directory of external/custom-r is dirty, doesn't that mean you should commit the submodule and then update the ref?

yes, if you want the changes from the submodule to be commited, sure

though currently it always says dirty, since the .gitignore of our gnur is no good, and all the compiled files show up...

o-:
@JanJecmen so how's #62 goin? You are still blocking the master :)

o-:
addressed the comments here, rest see #65

PR #65, opened 2017-11-03
remove FunctionHandle & move dispatch table function into the DispatchTable Struct

o-:


rir/src/interpreter/interp_data.h
```diff
@@ -5,92 +5,20 @@
...
-/**
- * Aliases for readability.
- */
-typedef SEXP FunctionSEXP;
-typedef SEXP ClosureSEXP;
-typedef SEXP PromiseSEXP;
-typedef SEXP IntSEXP;
+typedef rir::Code Code;
+typedef rir::Function Function;
+typedef rir::DispatchTable DispatchTable;
```
JanJecmen: Just curious, is there a reason to not use `using rir::Code`?

rir/src/interpreter/runtime.cpp
```diff
@@ -17,15 +17,16 @@ Function* isValidFunctionSEXP(SEXP wrapper) {
...
+  returns nullptr.
  */
 Function* isValidClosureSEXP(SEXP closure) {
     if (TYPEOF(closure) != CLOSXP)
         return nullptr;
     DispatchTable* t = isValidDispatchTableObject(BODY(closure));
     if (t == nullptr)
         return nullptr;
-    Function* f = sexp2function(t->entry[0]);
+    Function* f = Function::unpack(t->entry[0]);
```
JanJecmen: The dispatch table wraps this: `Function* f = t->first();`

rir/src/interpreter/interp_context.h
```diff
@@ -38,9 +39,9 @@ typedef struct {
 /** Interpreter frame information.
  */
 typedef struct {
-    struct Code* code;
+    struct rir::Code* code;
     SEXP env;
-    OpcodeT* pc;
+    rir::Opcode* pc;
     size_t bp;
 } Frame;
```
JanJecmen: `Frame` is not used anywhere, do we keep it?
(unrelated, I know)

rir/src/ir/CodeEditor.cpp
```diff
@@ -33,49 +33,50 @@ CodeEditor::CodeEditor(SEXP in) {
...
-                CodeHandle ch(c);
-                CodeEditor* p = new CodeEditor(ch);
-                auto idx = ch.idx();
+                CodeEditor* p = new CodeEditor(c);
                 if (promises.size() <= idx)
                     promises.resize(idx + 1, nullptr);
                 promises[idx] = p;
                 defaultArguments.push_back(idx);
             }
+            ++idx;
```
JanJecmen, line 73: This looks weird. `idx` just goes from 0 to \#codes and so the promises vector is always stretched by just 1.

I reused the resizing from the code further down (I think because of the possibility that code handle could skip some indexes?) but here it should probably just be `promises.push_back(p);`

Anyway, also unrelated, so I will make a note and fix it later...

rir/src/utils/FunctionWriter.h
```diff
@@ -91,39 +61,35 @@ class FunctionHandle {
...
-            store = newStore;
-            payload = newPayload;
-            function = (Function*)payload;
+            function = reinterpret_cast<Function*>(newPayload);
             capacity = newCapacity;
         }
 
         unsigned offset = function->size;
-        void* insert = (void*)((uintptr_t)payload + function->size);
+        void* insert = (void*)((uintptr_t)function + function->size);
```
JanJecmen: `reinterpret_cast<void*>(reinterpret_cast<uintptr_t>(function) + function->size)`?
(gosh that's ugly)

rir/src/utils/FunctionWriter.h
```diff
@@ -164,44 +130,27 @@ class FunctionHandle {
...
 
-        assert(code.code->srcLength == sources_idx);
+        assert(code->srcLength == sources_idx);
         function->codeLength++;
 
         // set the last code offset
         function->foffset = offset;
 
-        memcpy(callSites(code.code), callSiteBuffer, callSiteLength);
+        memcpy((void*)code->callSites(), callSiteBuffer, callSiteLength);
```
JanJecmen: Also C++ cast? Probably a lot of other places as well, maybe we should leave that to a separate cleanup PR?

rir/src/runtime/Code.cpp
```diff
@@ -0,0 +1,56 @@
+#include "Code.h"
+#include "utils/Pool.h"
+#include "ir/CodeEditor.h"
+#include "ir/BC.h"
+
+namespace rir {
+Code::Code(SEXP ast, unsigned cs, unsigned sourceSize, unsigned csl,
+           unsigned offset, bool isDefaultArg) {
+    magic = CODE_MAGIC;
```
JanJecmen, line 9: `magic` is already initialized in the header file.

rir/src/runtime/Code.h
```diff
@@ -0,0 +1,279 @@
...
+const static unsigned CallSiteProfile_maxTargets = 4;
+struct CallSiteProfile {
+    uint32_t taken : 28;
+    uint32_t takenOverflow : 1;
+    uint32_t numTargets : 2;
+    uint32_t targetsOverflow : 1;
+    SEXP targets[3];
+};
+
+struct CallSiteStruct {
```
JanJecmen, line 195: Maybe could rename to `CallSite` now that the original wrapper is gone?

rir/src/runtime/Function.h
```diff
@@ -0,0 +1,149 @@
...
+typedef SEXP IntSEXP;
+
+// Function magic constant is designed to help to distinguish between Function
+// objects and normal EXTERNALSXPs. Normally this is not necessary, but a very
+// creative user might try to assign arbitrary EXTERNAL to a closure which we
+// would like to spot. Of course, such a creative user might actually put the
+// magic in his vector too...
+#define FUNCTION_MAGIC (unsigned)0xCAFEBABE
+
+#define DISPATCH_TABLE_MAGIC (unsigned)0xBEEF1234
```
JanJecmen: Should go to `DispatchTable.h`, no?

JanJecmen (commented):
I like it, but we should've probably went the other way round and merge this before #62... Would probably be less work.

rir/src/interpreter/interp_context.h
o-: oh, I didn't notice, thanks

rir/src/interpreter/interp_data.h
o-: I don't like typing :)

also I expect that we want to move it into the rir namespace at some point, so then you would have to remove all the rir:: again...

rir/src/ir/CodeEditor.cpp
o-: hmm, I am not sure what the resize is supposed to do, honestly. but I think that the code does the same as before

rir/src/runtime/Code.cpp
o-: right

rir/src/runtime/Code.h
o-: good point

rir/src/runtime/Function.h
o-: yep

rir/src/utils/FunctionWriter.h
o-: yeah,  idon't like them. I usually do not write reinterpret_cast, since it is essentially the same as a C style cast, just a more ugly name

o-:
at the moment I am kinda caught up with other stuff. I will try to rebase and merge this PR  tomorrow or Wed

PR #69, opened 2017-11-27
Signature

JanJecmen:
This PR addresses several things:

- New `FunctionSignature` class is added, but not yet used
- New instruction `call_ordered_` is added that at compile time tries to look up what is getting called and if the arguments are all constants, it avoids at runtime creating any promises and the arglist
- Magic numbers are moved to the RIR header to fix a bug (description in 63deecf)

Also I haven't managed to come up with good names, feel free to suggest better ones.

I know these should be three different PRs, sorry about that..


rir/src/runtime/FunctionSignature.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include "R/r.h"
+
+namespace rir {
+
+struct FunctionSignature {
```
mhyee, line 7: Make `FunctionSignature` a `class`? The only difference is that the default access is `private`, instead of `public` for `struct`s.

For convention, I prefer to use `struct` for very simple things, e.g. a collection of fields (sometimes called Plain Old Data). Anything more complicated and I use `class`.

But then I noticed some of the other C++ classes (e.g. `Code` and `Function`) are structs... So maybe it's best to be consistent and use struct.

rir/src/runtime/DispatchTable.h
```diff
@@ -49,9 +49,11 @@ struct DispatchTable {
         return Function::unpack(entry[0]);
     }
 
-    rir::rir_header info;  /// for exposing SEXPs to GC
+    Function* getMatching(FunctionSignature* sig) {
+        return first();
```
mhyee: Since this is intentionally incomplete, leave a TODO?

rir/src/runtime/FunctionSignature.h
```diff
@@ -0,0 +1,71 @@
...
+
+        ArgumentType() = default;
+        ArgumentType(SEXP s) : type(TYPEOF(s)) {
+            if (type == PROMSXP) {
+                isEvaluated = PRVALUE(s) != R_UnboundValue;
+            }
+            if (isEvaluated)
+                length = XLENGTH(s);
+        }
+        ArgumentType(bool evaled, unsigned int type) :
```
mhyee: Is it OK if this constructor doesn't initialize `length`? What if the argument has already been evaluated?

rir/src/runtime/FunctionSignature.h
```diff
@@ -0,0 +1,71 @@
...
+        Rprintf("    on stack?      %s\n", argsOnStack ? "true" : "false");
+        Rprintf("    args (%u):\n", arguments.size());
+        for (auto arg : arguments)
+            arg.print();
+        Rprintf("\n");
+    }
+
+    FunctionSignature() = default;
+
+    bool argsOnStack = false;
```
mhyee, line 69: If we're using `struct` for `FunctionSignature`, then maybe `argsOnStack` and the `arguments` vector should be marked as `private`.

rir/src/runtime/Function.h
```diff
@@ -51,17 +48,16 @@ typedef SEXP SignatureSEXP;
 #pragma pack(1)
 struct Function {
     Function() {
-        magic = FUNCTION_MAGIC;
         info.gc_area_start = sizeof(rir_header);  // just after the header
-        info.gc_area_length = 3; // signature, origin, next
-        signature_ = nullptr;
+        info.gc_area_length = 2;                  // origin, next
```
mhyee, line 52: Maybe I missed it or forgot, but we decided to allocate `FunctionSignature`s on the C++ heap? So we don't trace them with the GC, but won't we leak memory?

rir/src/ir/insns.h
```diff
@@ -260,6 +260,12 @@ DEF_INSTR(call_stack_, 2, -1, 1, 0)
  */
 DEF_INSTR(static_call_stack_, 2, -1, 1, 0)
 
+/**
+ * call_ordered_:: arguments are taken from the stack
```
mhyee: Could you elaborate this comment? Right now the description is basically the same as `call_stack_`.

Also, I don't really like the name. `call_eager_` is the best I can come up with right now...

rir/src/ir/Compiler.cpp
```diff
@@ -1073,6 +1073,87 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
...
+bool compileWithGuess(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
+    RList args(args_);
+    CodeStream& cs = ctx.cs();
+
+    SEXP cls = findRIRClosure(fun, R_GlobalEnv);  // or current env of the closure?
+    if (!cls)
+        return false;
+
+    RList formals(FORMALS(cls));
+    if (formals.length() != args.length() || args.length() == 0)
```
mhyee: Stupid question, but why do we check that the length of args is nonzero?

rir/src/ir/Compiler.cpp
```diff
@@ -1073,6 +1073,87 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
...
+        return false;
+
+    for (auto farg : formals)
+        if (farg == R_DotsSymbol)
+            return false;
+
+    for (auto a = args.begin(); a != args.end(); ++a) {
```
mhyee, line 1103: `for (auto a : args) ...`?

rir/src/ir/Compiler.cpp
```diff
@@ -1073,6 +1073,87 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
...
+    // maybe use promise_ and compile everything either as push or
+    // as promise?
+
+    // do argument matching / shuffling?
+    // cannot reuse matchArgs from gnur (but could rewrite it)
+
+    cs << BC::guardName(fun, cls);
+
+    FunctionSignature* signature = new FunctionSignature();
+    signature->argsOnStack = true;
```
mhyee, line 1139: Oh, that's why `argsOnStack` is public. But I'd probably add a new constructor for `FunctionSignature`. Unless we plan on mutating `argsOnStack` after construction?

mhyee (commented):
I have a bunch of small comments and questions. I'm fine with `FunctionSignature` and `ArgumentType` as names, but I don't really like `call_ordered_`. `call_eager_` was the first thing I came up with; not sure it's better.

Debugging the magic numbers must have been a headache, but at least the fix is simple. Thanks!

rir/src/ir/Compiler.cpp
JanJecmen: No reason, just a leftover from debugging I guess.

rir/src/ir/Compiler.cpp
JanJecmen: Need the iterator itself as I look at the TAGs... range based for would just give me the CARs.

rir/src/ir/insns.h
JanJecmen: Done. Also renamed the instruction, this is definitely better.

rir/src/runtime/DispatchTable.h
JanJecmen: :+1:

rir/src/runtime/Function.h
JanJecmen: Yes, we decided for now to just leak them. The reasoning was that there shouldn't be that many anyway... but when I think about it, we have one for every compiled closure, plus one for every callsite (as of now this only applies to the new instruction, the others just store nullptr) -- that will be quite a bunch

rir/src/runtime/FunctionSignature.h
JanJecmen: Or change all to classes?

o-:
oh my, what an ugly bug to debug, i am sorry. thanks

o-:
sorry, I didn't have time yet, since I was travelling. But it looks good, and @mhyee already did a great job in reviewing :)

One high-level thing: there would also be the option (when recompiling) of getting profiling information from the call site. See https://github.com/reactorlabs/rir/blob/35738799ad2af2723fbe9662f28d2a5341e0f101/rir/src/ir/BC.cpp#L272 for an example.

rir/src/runtime/Function.h
mhyee: Yeah, let's see how bad this ends up becoming, and then fix it later.

rir/src/runtime/FunctionSignature.h
mhyee: @o-, what do you think? Structs or classes?

If we change all to classes, let's do that in a separate PR.

o-: I don't care too much. I would keep the things which are gnur-object mapped as structs. and since the function sigunature will eventually be that way too, I think it's fine.

o-: at some point I was under the impression that structs are guaranteed to not have vtables, but it seems that this is actually not true.

mhyee: OK, let's use structs, and then explicitly make things private.

o-: I don't like private anyway :) It just makes you use 'friend class' more often

mhyee: I don't feel strongly enough to argue this point, and we've spent enough time here, so to keep things simple, just ignore this comment thread and leave it alone. :)

rir/src/runtime/FunctionSignature.h
JanJecmen: Tbh I don't know what to make of the `lenght`... It is initialized to `-1` via the in-class initializer, and the idea was that this would mean don't know / don't care, the constructor only updating the value if it knows better.

But I agree that it should be commented at least.

o-:
ehm, should we merge this?

JanJecmen:
Oh my why did I not split this into more PRs... I will patch this a bit and then perhaps...

JanJecmen:
@o- Merge now? Still not sure about the signatures...

PR #71, opened 2018-02-02
slightly change how we compile for loops

o-:
we used to compile forloops by checking if the index is within the
looped-over array. that is slightly annoying to analyze. So lets
instead get the size of the looped-over array and compare if the
index variable is smaller or equal to the size.

### DIFF
--- a/rir/src/analysis/dataflow.h
+++ b/rir/src/analysis/dataflow.h
@@ -692,7 +692,7 @@ class DataflowAnalysis
         case Opcode::close_:
         case Opcode::lgl_or_:
         case Opcode::lgl_and_:
-        case Opcode::test_bounds_:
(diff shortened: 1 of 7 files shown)


JanJecmen:
👍
Our `length_` assumes it gets a vector though... The failed builds appear to do sth like
`for (pkg in setdiff(...)) ...` where the diff returns `NULL` and `XLENGTH(R_NilValue)` then gives random rubbish.
Maybe this `R_xlen_t len = Rf_length(val);` instead?

o-:
yeah, thanks, that was rubbish of course. I think we really need a dedicate loop sequence lenght instruction, since there is also a dedicated error message if the sequence is not vector/list/nil.

let's see if that works

PR #73, opened 2018-02-07
Optimizer invocation tweak

JanJecmen:
Before, when a RIR closure was called from the GNU R eval, e.g. directly from REPL, we did not track the number of invocations nor did we try to optimize such closures..

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -123,6 +123,50 @@ static void jit(SEXP cls, Context* ctx) {
     SET_BODY(cls, BODY(cmp));
 }
 
+void tryOptimizeClosure(DispatchTable* table, size_t offset, Context* ctx) {
+
+    // This function will optimize the function and update the linked list
+    // of functions, so that subsequent closure calls pick the most optimized
+    // version.
+
+    static bool optimizing = false;
+
+    Function* fun = table->at(offset);
+
+    if (!optimizing && !fun->next() && !fun->origin()) {
+         /* currently there is a bug if we reoptimize a function twice:
+          *  Deopt ids from the first optimization and second optimizations
+          *  will be mixed and the deoptimizer always only goes back one
+          *  optimization level!
+          *  To avoid this bug we currently only optimize once
+          */
+        Code* code = fun->body();
+        if (fun->markOpt ||
+                (fun->invocationCount == 1 && code->perfCounter > 100) ||
+                (fun->invocationCount == 10 && code->perfCounter > 20) ||
+                 fun->invocationCount == 100) {
+            optimizing = true;
+
+            Function* oldFun = fun;
+            cp_pool_add(ctx, oldFun->container());
+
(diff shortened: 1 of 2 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -123,6 +123,50 @@ static void jit(SEXP cls, Context* ctx) {
...
+                 fun->invocationCount == 100) {
+            optimizing = true;
+
+            Function* oldFun = fun;
+            cp_pool_add(ctx, oldFun->container());
+
+            SEXP opt = globalContext()->optimizer(fun->container());
+
+            if (opt != nullptr) {
+                fun = Function::unpack(opt);
```
mhyee, line 156: There used to be a `PROTECT` and `UNPROTECT` here. Was it removed because we're already adding the container to the constant pool (line 151)?

rir/src/interpreter/interp.cpp
```diff
@@ -2705,20 +2705,30 @@ SEXP rirExpr(SEXP f) {
...
-        return evalRirCode(ff->body(), globalContext(), env, 0);
-    } else if ((t = DispatchTable::check(f))) {
-        // Default target is the first version in the dispatch table
-        return evalRirCode(t->first()->body(), globalContext(), env, 0);
+    if ((c = isValidCodeObject(what))) {
+        // nothing
+    } else if ((f = Function::check(what))) {
+        // note: could try to optimize, but no way to update the
+        // containing dispatch table
+        // also this branch is probably not taken at all,
```
mhyee: Can we confirm if this branch is ever taken? If we're sure that it isn't, maybe we should just remove it?

I think this was originally my fault, when I added dispatch tables. Probably I didn't want to break or change too many things at once, and then never actually removed this branch.

mhyee (approved):
Just some small comments. And it looks like the branch needs to be updated.

rir/src/interpreter/interp.cpp
JanJecmen: No, the `PROTECT` was for the container of the newly optimized version. However, I think it is not needed: `vtable->put()` should not cause a GC, since it does not reallocate the table - it just writes to the same slot. And what happens inside is just a write barrier check, and that should not collect either, right?

rir/src/interpreter/interp.cpp
JanJecmen: This is just my guess, not sure how to verify it. The reasoning is:

The function is not called anywhere in RIR, it is only registered in GNUR as an evaluation callback for EXTERNALSXPs. So the question is, can it happen that there is a naked Function object flying around somewhere in the open? The answer is hopefully no...

As for the Code branch, I believe it is used when forcing promises, so that one we need.

mhyee: Yeah, I agree that Code branch is needed, and Function branch seems unlikely.

If we expect all function objects to be in dispatch tables, then naked Function objects should be wrong... so maybe we should let this case fall to the assert?

PR #74, opened 2018-02-07
Change verbosity of rir.disassemble

JanJecmen:


mhyee (approved):

o-:
so I think we should just directly merge those small PR's. If anybody does not agree, that person can still re-open the PR and discuss it there.

PR #76, opened 2018-02-19
let's clean up the horrible naming around `[` and `[[`

o-:
internally we called `[[` extract and `[` subset, but then
`[[<-` and `[<-` both subassign. In the formar case we used
1,2 as a postfix to denote 1 vs. 2-dimensional indexing,
whereas in the later case we used it do distinguish between
`[` and `[[` :/

this commit cleans is up. Now both `[[` and `[` are called
extract. The former is extract2, the later is extract1.

Then we use a second index for the number of indices. So
`a[1,2]` is extract_1_2 `a[1]` is extract1_1 and `a[[1]]`
is extract2_1.

### DIFF
--- a/rir/src/ir/BC_inc.h
+++ b/rir/src/ir/BC_inc.h
@@ -287,10 +287,10 @@ class BC {
     inline static BC isfun();
     inline static BC invisible();
     inline static BC visible();
-    inline static BC extract1();
-    inline static BC subset1();
-    inline static BC extract2();
-    inline static BC subset2();
+    inline static BC extract1_1();
+    inline static BC extract1_2();
+    inline static BC extract2_1();
+    inline static BC extract2_2();
     inline static BC swap();
     inline static BC put(uint32_t);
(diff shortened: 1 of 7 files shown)


rir/src/ir/BC_inc.h
```diff
@@ -285,9 +285,10 @@ class BC {
...
     inline static BC invisible();
     inline static BC visible();
-    inline static BC extract1();
-    inline static BC subset1();
-    inline static BC extract2();
+    inline static BC extract1_1();
+    inline static BC extract1_2();
+    inline static BC extract2_1();
+    inline static BC extract2_2();
     inline static BC subset2();
```
JanJecmen: Is the `subset2` just a leftover from before?

JanJecmen (approved):

rir/src/ir/BC_inc.h
o-: oh, yeah, that should have been deleted too, thanks

PR #77, opened 2018-02-20
refactor instructions

JanJecmen:
- renamed `subassign_` to `subassign1_` to reflect the new naming convention
- reordered the instructions to match in `interp.cpp` and `insns.h`
- fixed `ldfun_` description


o-:
nice

PR #78, opened 2018-02-23
Interpreter cleanup

JanJecmen:
Mainly unify `doCall` with `doCallStack` and `doDispatch` with `doDispatchStack` to reduce some code duplication.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -79,15 +79,15 @@ void initClosureContext(RCNTXT* cntxt, SEXP call, SEXP rho, SEXP sysparent,
        is a straight substitution of the generic.  */
 
     if (R_GlobalContext->callflag == CTXT_GENERIC)
-        begincontext(cntxt, CTXT_RETURN, call, rho, R_GlobalContext->sysparent,
-                     arglist, op);
+        Rf_begincontext(cntxt, CTXT_RETURN, call, rho,
+                        R_GlobalContext->sysparent, arglist, op);
     else
-        begincontext(cntxt, CTXT_RETURN, call, rho, sysparent, arglist, op);
+        Rf_begincontext(cntxt, CTXT_RETURN, call, rho, sysparent, arglist, op);
 }
 
 void endClosureContext(RCNTXT* cntxt, SEXP result) {
     cntxt->returnValue = result;
-    endcontext(cntxt);
+    Rf_endcontext(cntxt);
 }
 
 INLINE SEXP createPromise(Code* code, SEXP env) {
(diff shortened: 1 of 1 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -91,7 +91,7 @@ void endClosureContext(RCNTXT* cntxt, SEXP result) {
 }
 
 INLINE SEXP createPromise(Code* code, SEXP env) {
-    SEXP p = mkPROMISE((SEXP)code, env);
+    SEXP p = Rf_mkPROMISE((SEXP)code, env);
```
mhyee, line 94: What's the reason for this change? Nothing wrong with it -- I'm just not familiar with the code base and conventions.

rir/src/interpreter/interp.cpp
```diff
@@ -536,82 +536,76 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
-        } else {
-            argslist = createArgsList(caller, call, nargs, cs, env, ctx, false);
-        }
-        p(argslist);
+
+    if (TYPEOF(callee) == CLOSXP) {
         // if body is EXTERNALSXP, it is rir serialized code, execute it
         // directly
-        SEXP body = BODY(callee);
+        SEXP body = BODY(callee), result;
```
mhyee: So it's not immediately obvious to me why you're using the comma operator. Is this a typo, or am I just confused?

rir/src/interpreter/interp.cpp
```diff
@@ -614,258 +533,146 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
-        default:
-            assert(false && "Don't know how to run other stuff");
-        }
-    } while (false);
-
-    ostack_popn(ctx, 2);
-    // This line resets the sp after possible non-local return
-    assert(res);
-    return res;
+    // not reached
```
mhyee: If this is never reached, then we should assert false.

mhyee (approved):
Thanks! Just a few quick comments.

I think after the refactoring in 0f133e9c699303e5f75657a4baaaa748a12ac370, the code is slightly more difficult to read, because of the early returns and combining multiple cases. On the other hand, it reduces code duplication, so it's probably a net win.

o-:
heroic :+1:

rir/src/interpreter/interp.cpp
JanJecmen: My reasoning was that this makes it more clear that we are calling into R. `Rf_` is short for `R function`, I think?
So this way: `createPromise(...)` -> calling sth. from RIR, vs. `Rf_mkPROMISE(...)` -> calling R

rir/src/interpreter/interp.cpp
JanJecmen: Oh, this is just a declaration... I agree that it's kind of ugly

JanJecmen:
So the style with early returns is quite natural to me personally, but I can change it a bit more... I had several iterations and there is a middle ground.

rir/src/interpreter/interp.cpp
mhyee: Makes sense, thanks!

rir/src/interpreter/interp.cpp
```diff
@@ -614,258 +533,147 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
+    assert(false);
```
mhyee: Sorry to keep nitpicking, but add an error string?

mhyee:
> So the style with early returns is quite natural to me personally, but I can change it a bit more... I had several iterations and there is a middle ground.

I like early returns when they're short, simple, and at the top of the function. It feels a bit weird to me when there's a dozen lines of code, then a condition with another dozen lines, then an early return.

It's fine the way it is -- better than what was there previously, and probably not worth the effort to keep iterating.

JanJecmen:
Ok so made one last change, inspired by the `doDispatch`: now it's in `do {...} while(0)` and returns only at the end. Hopefully it's more readable this way.

PR #79, opened 2018-02-27
pir the more Powerful IR for R

o-:
This commit instroduces all the parts of the IR and a simple RIR to
PIR compiler in "pir_compiler.cpp".


o-:
(ignore the verifier, I have a new version)

o-:
tbr

rir/src/utils/Bitset.h
```diff
@@ -0,0 +1,62 @@
+#ifndef RIR_BITSET_H
+#define RIR_BITSET_H
+
+#include <initializer_list>
+
+namespace rir {
+
+template <typename STORE, typename BASE_TYPE>
+class BitSet {
```
mhyee, line 9: `std::bitset`?

rir/src/compiler/pir/pir_impl.h
```diff
@@ -0,0 +1,15 @@
+#ifndef PIR_PIR_IMPL_H
+#define PIR_PIR_IMPL_H
+
+// PIR includes. Please only include in .cpp!
```
mhyee, line 4: How do I know which headers are included in this file, and which are not? For example, why is `type.h` not included? If I create a new `foobar.h` file, should I include it here?

If `pir.h` contains all the forward declarations, shouldn't it be the first to be included?

rir/src/compiler/pir/instruction_list.h
```diff
@@ -0,0 +1,61 @@
+#ifndef COMPILER_INSTRUCTION_LIST_H
+#define COMPILER_INSTRUCTION_LIST_H
+
+#define COMPILER_INSTRUCTIONS(V)                                               \
```
mhyee, line 4: I know that this is a common trick, and RIR uses it too, but what if I add a new instruction in `instruction.h` but then forget to update this file?

I don't think there's any good way of enforcing that both files stay in sync, so very prominent documentation is needed.

It might also help if the instructions in this list were grouped in the same way as `instruction.h`, e.g., fixed-length instructions, variable-length instructions, binop, unop, etc.

rir/src/compiler/pir/value_list.h
```diff
@@ -0,0 +1,9 @@
+#ifndef COMPILER_VALUE_LIST_H
+#define COMPILER_VALUE_LIST_H
+
+#define COMPILER_VALUES(V)                                                     \
```
mhyee, line 4: I think seeing `instruction_list.h` threw me off -- I expected a `Missing`, `Env`, and `Nil` to be defined in `value.h`. But `Missing` and `Nil` are defined in `singleton_values.h`, and `Env` is defined in `env.h`.

rir/src/compiler/pir/tag.h
```diff
@@ -0,0 +1,31 @@
...
+#include "instruction_list.h"
+#include "value_list.h"
+
+#include <cstdint>
+
+namespace rir {
+namespace pir {
+
+/*
+ * Every Value has a unique tag, to be able to safely downcast values and
```
mhyee, line 13: Might be useful to reiterate that an `Instruction` is a `Value`.

rir/src/compiler/pir/tag.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+const char* TagToStr(Tag tag) {
+    switch (tag) {
+#define V(I)                                                                   \
+    case Tag::I:                                                               \
+        return #I;
+        COMPILER_INSTRUCTIONS(V)
+#undef V
+#define V(I)                                                                   \
+    case Tag::I:                                                               \
```
mhyee, line 16: Not immediately obvious why these cases are unreachable. If every `Instruction` is a `Value`, and every `Value` has a tag, why wouldn't we print tags for `COMPILER_VALUES`?

rir/src/compiler/pir/tag.h
```diff
@@ -0,0 +1,31 @@
...
+    Unused,
+#define V(I) I,
+    COMPILER_INSTRUCTIONS(V)
+#undef V
+#define V(I) I,
+        COMPILER_VALUES(V)
+#undef V
+};
+
+const char* TagToStr(Tag t);
```
mhyee, line 27: `tagToStr`? What's our naming convention?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+#include <vector>
+
+#include "utils/Bitset.h"
+
+#include "R/r.h"
+
+namespace rir {
+namespace pir {
+
+enum class RType : uint8_t {
```
mhyee, line 16: I assume these are the R types coming from GNU R, and we're using an `enum class` instead of a bunch of `#define`s. Add a short comment about this?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+    raw,
+
+    closure,
+    prom,
+
+    code,
+    env,
+    ast,
+
+    max
```
mhyee, line 40: I'd prefer calling it `MAX` to make it stand out more. Right now it looks like an ordinary RType, and it's easy to accidentally add something after it.

Another trick is to write `MAX = ast` because:
1. That way, it actually *is* the maximum value.
2. It stands out even more.
3. You don't have to include it in switches.

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+
+#include "utils/Bitset.h"
+
+#include "R/r.h"
+
+namespace rir {
+namespace pir {
+
+enum class RType : uint8_t {
+    unused,
```
mhyee, line 17: Change to `UNUSED` so it's clearer that it's a marker and not some "unused type."

Unfortunately, I don't know of a trick similar to `MAX = last_element`, so all the range checks will be half-open intervals, e.g., `t > RType::UNUSED && t <= RType::MAX`. At least it's a bit more intuitive that `UNUSED` is outside the range and `MAX` is included in the range?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+
+    max
+};
+
+/*
+ * Values in PIR are either types from R (RType), or native types (NativeType).
+ *
+ * In both cases we use union types, represented by a bitset.
+ *
+ * There is an additional flags bitset, that adds some modifiers.
```
mhyee, line 68: Assuming this is what `TypeFlags` is for. Mention it in the comment?

rir/src/compiler/pir/type.h
mhyee: I see there's a larger comment below, just before `PirType`. I think it makes more sense to put it at the very top of the file.

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+ *
+ * As an example, an R integer, that is potentially promised, has:
+ * 
+ *  - flag : TypeFlag::rtype | TypeFlag::lazy
```
mhyee, line 72: Should be `flags_`, not `flag`.

rir/src/compiler/pir/type.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+extern "C" Rboolean(Rf_isObject)(SEXP s);
+
+namespace rir {
+namespace pir {
+
+void PirType::print() { std::cout << *this << "\n"; }
+
+PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
+    switch (TYPEOF(e)) {
```
mhyee, line 11: `RType::code` is missing from here?

mhyee: Also, comparing the R sexp types and the PIR RTypes, two things stand out:

1. We declare the RTypes in a different order (which confused me when I was flipping back and forth).
2. What's our naming convention? Do we want to follow the short and terse names, e.g., `sym` and `clos` instead of `symbol` and `closure`? But then `prom` should probably be `promise`.

Other names that stand out to me:
- `env` is OK because it's a common abbreviation for environment.
- `LANGSXP` becomes `ast`?
- Names should probably be singular rather than plural. See `chars`, `dbls`, `strs`, and `vecs`.
- Sort of see why we don't want an RType called `real`, but I'd prefer `double` over `dbls`. Or maybe `real` isn't all that bad...

rir/src/compiler/pir/type.cpp
```diff
@@ -0,0 +1,77 @@
...
+    case WEAKREFSXP:
+    case S4SXP:
+        assert(false);
+    }
+
+    if (Rf_isObject(e)) {
+        flags_.set(TypeFlags::obj);
+    }
+
+    if (PirType::vecs() >= *this) {
```
mhyee, line 71: What is this doing?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+ *  - flag : ()
+ *  - t_.n : NativeType::test
+ *
+ * An R value (not a promise), has:
+ *
+ *  - flag : TypeFlag::rtype
+ *  - t_.r : RType::symbol | ... | RType::ast
+ *
+ */
+struct PirType {
```
mhyee, line 86: Why is most of the implementation here, in the header file, and not the cpp file?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+    PirType(const NativeType& t) : t_(t) {
+        assert(t > NativeType::unused && t < NativeType::max);
+    }
+    PirType(const RTypeSet& t) : flags_(defaultRTypeFlags()), t_(t) {}
+    PirType(const NativeTypeSet& t) : t_(t) {}
+    PirType(SEXP);
+
+    void operator=(const PirType& o) {
+        flags_ = o.flags_;
+        if (flags_.includes(TypeFlags::rtype))
```
mhyee, line 116: `o.isRType()`

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+        t.flags_.set(TypeFlags::lazy);
+        return t;
+    }
+
+    PirType baseType() const {
+        assert(isRType());
+        return PirType(t_.r);
+    }
+
+    static const PirType voyd() { return NativeTypeSet(); }
```
mhyee, line 178: Not sure how I feel about `voyd`... Maybe 😖 or 😱. Can we choose a different name?

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+        assert(t > RType::unused && t < RType::max);
+    }
+    PirType(const NativeType& t) : t_(t) {
+        assert(t > NativeType::unused && t < NativeType::max);
+    }
+    PirType(const RTypeSet& t) : flags_(defaultRTypeFlags()), t_(t) {}
+    PirType(const NativeTypeSet& t) : t_(t) {}
+    PirType(SEXP);
+
+    void operator=(const PirType& o) {
```
mhyee, line 114: This is the copy assignment operator, but don't we also need a copy constructor? In particular, code like

```
PirType t = *this;
...
```

invokes the copy constructor (because `t` hasn't been initialized yet).

So maybe it's the copy constructor we want (does the default one work?) and not the assignment operator? Maybe we should `= delete` the assignment operator?

Urgh, I forgot how annoying C++ can be. I always mix up the copy constructor and assignment operator, and even now I'm not sure I have them straight.

rir/src/compiler/pir/type.h
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    bool operator!=(const PirType& o) const { return !(*this == o); }
+
+    bool operator==(const PirType& o) const {
+        return flags_ == o.flags_ &&
+               (isRType() ? t_.r == o.t_.r : t_.n == o.t_.n);
+    }
+
+    bool operator>=(const PirType& o) const {
```
mhyee, line 215: Some explanation or ASCII art of the lattice would help.

rir/src/compiler/pir/singleton_values.h
```diff
@@ -0,0 +1,44 @@
...
+#include <functional>
+#include <iostream>
+
+namespace rir {
+namespace pir {
+
+template <typename T>
+class SingletonValue : public Value {
+  public:
+    SingletonValue(PirType t, Tag tag) : Value(t, tag) {}
```
mhyee, line 17: Should probably delete the copy constructor and assignment operators. (And the deleted functions should be public. Apparently the error messages will be better.)

Also, shouldn't the `SingletonValue` constructor be private?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+ *
+ * Every instruction is also a Value, and can therefore be used as an argument
+ * for other instructions.
+ *
+ * Instructions have an InstructionDescription, which gives us basic
+ * information about its effects and environment interactions.
+ *
+ * If an instruction needs an envrionment (ie. if its EnvAccess >= None), it
+ * needs to have a dedicated environment argument. This dedicated environment
+ * input is (for technical reasons) the last argument of fixed length
```
mhyee, line 31: I hope there's some later explanation for this limitation (env is first or last argument). Because right now, the inconsistency bothers me...

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+ * instructions and the first argument for variable length instructions. There
+ * is some machinery to enforce passing an environment to the respective
+ * superclassses.
+ *
+ * Every instruction has a unique instruction tag, which is used to "Cast" in
```
mhyee, line 36: Typo? in -> an

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    virtual bool changesEnv() = 0;
+    virtual bool leaksEnv() { return false; }
+    virtual bool needsLiveEnv() { return false; }
+    virtual bool hasEnv() { return false; }
+    virtual bool accessesEnv() { return false; }
+    virtual Value* env() = 0;
+
+    virtual Instruction* clone() = 0;
+
+    typedef std::pair<unsigned, unsigned> Id;
```
mhyee, line 72: What do the elements of the tuple represent?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        }
+        out << ")";
+    }
+
+    void print(std::ostream&);
+    void printRef(std::ostream& out);
+    void print() { print(std::cerr); }
+
+    Value* arg(size_t pos, Value* v) {
+        assert(pos < nargs() && "This instruction has less arguments");
```
mhyee, line 113: Grammar nitpick: less -> fewer

(Use "fewer" for things you can count.)

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        args()[pos] = v;
+        return v;
+    }
+
+    Value* arg(size_t pos) {
+        assert(pos < nargs() && "This instruction has less arguments");
```
mhyee, line 119: less -> fewer

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        assert(bb_);
+        return bb_;
+    }
+
+    Instruction(Tag tag, PirType t) : Value(t, tag) {}
+    virtual ~Instruction() {}
+
+    Id id();
+
+    typedef std::function<void(Value*)> arg_iterator;
```
mhyee, line 84: Maybe some comments to explain these typedefs. I'm still not sure what PirType is for.

I'm also not a big fan of names like `arg_iterator2`.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+enum class Effect : uint8_t {
+    None,
+    Warn,
+    Error,
+    Print,
+    Any,
+};
+
+template <Tag class_tag, class Base, Effect EFFECT, EnvAccess ENV>
+class InstructionDescription : public Instruction {
```
mhyee, line 182: So I get that this class hierarchy (concrete instructions inherit from `InstructionDescription` which inherits from `Instruction`) is so you can reuse the implementation... but I can't help but think "favour composition over inheritance."

Semantically, it's weird to say that an `InstructionDescription` is-a `Instruction`.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        args()[pos] = v;
+        return v;
+    }
+
+    Value* arg(size_t pos) {
+        assert(pos < nargs() && "This instruction has less arguments");
+        return args()[pos];
+    }
+
+    void each_arg(arg_iterator it) {
```
mhyee, line 123: Perhaps more documentation for the API. When should I use `each_arg` or `map_arg`? What are the two different versions for?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        return new Base(*static_cast<Base*>(this));
+    }
+
+    static Base* Cast(Value* i) {
+        if (i->tag == class_tag)
+            return static_cast<Base*>(i);
+        return nullptr;
+    }
+
+    static void If(Instruction* i, std::function<void()> maybe) {
```
mhyee, line 211: Documentation for what `If` is for?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+
+enum class Effect : uint8_t {
+    None,
+    Warn,
+    Error,
+    Print,
+    Any,
+};
+
+template <Tag class_tag, class Base, Effect EFFECT, EnvAccess ENV>
```
mhyee, line 181: Is there a naming convention for how we handle template arguments? We have lower case (`class_tag`), title case (`Base`), and all caps (`EFFECT`). And the extra complication that `Base` is a type but the other template arguments are values...

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+            maybe(b);
+    }
+};
+
+template <Tag class_tag, class Base, size_t ARGS, Effect EFFECT, EnvAccess ENV>
+class FixedLenInstruction
+    : public InstructionDescription<class_tag, Base, EFFECT, ENV> {
+  private:
+    typedef InstructionDescription<class_tag, Base, EFFECT, ENV> Super;
+    std::array<Value*, ARGS> arg_;
```
mhyee, line 229: Not `args_`?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    Value* env() override {
+        // TODO find a better way
+        assert(ENV != EnvAccess::None);
+        Value* env = arg_[ARGS - 1];
+        return env;
+    }
+
+    template <unsigned pos>
+    Value* arg(Value* v) {
+        static_assert(pos < ARGS, "This instruction has less arguments");
```
mhyee, line 248: less -> fewer

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        arg_[pos] = v;
+        return v;
+    }
+
+    template <unsigned pos>
+    Value* arg() {
+        static_assert(pos < ARGS, "This instruction has less arguments");
```
mhyee, line 255: less -> fewer

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    }
+};
+
+template <Tag class_tag, class Base, size_t ARGS, Effect EFFECT, EnvAccess ENV>
+class FixedLenInstruction
+    : public InstructionDescription<class_tag, Base, EFFECT, ENV> {
+  private:
+    typedef InstructionDescription<class_tag, Base, EFFECT, ENV> Super;
+    std::array<Value*, ARGS> arg_;
+    const std::array<PirType, ARGS> arg_type;
```
mhyee, line 230: `arg_types`?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+            (*this)[ARGS - 1] = env;
+        }
+        ArgsWithEnv(Value* env) : std::array<Value*, ARGS>({{env}}) {}
+    };
+
+    FixedLenInstruction(PirType return_type, Value* env)
+        : Super(return_type), arg_(ArgsWithEnv(env)),
+          arg_type(ArgTypesWithEnv()) {
+        assert(env);
+        static_assert(ARGS == 1, "Missing args");
```
mhyee, line 280: "Wrong number of args" is probably a better message.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        static_assert(ENV != EnvAccess::None,
+                      "This instruction has no environment access");
+    }
+
+    FixedLenInstruction(PirType return_type)
+        : Super(return_type), arg_({}), arg_type({}) {
+        static_assert(ARGS == 0, "Missing args");
```
mhyee, line 287: "Wrong number of args" is probably a better message. Especially since we're expecting 0 args, so "missing args" doesn't really make sense...

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        static_assert(ENV == EnvAccess::None,
+                      "This instruction needs an environment");
```
mhyee, line 289: Not entirely sure I understand this error message.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    }
+};
+
+template <Tag class_tag, class Base, Effect EFFECT, EnvAccess ENV>
+class VarLenInstruction
+    : public InstructionDescription<class_tag, Base, EFFECT, ENV> {
+  private:
+    typedef InstructionDescription<class_tag, Base, EFFECT, ENV> Super;
+
+    std::vector<Value*> arg_;
```
mhyee, line 318: `args_`?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    std::vector<PirType> arg_type;
```
mhyee, line 319: `arg_types`?

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    }
+
+    VarLenInstruction(PirType return_type) : Super(return_type) {
+        static_assert(ENV == EnvAccess::None,
+                      "This instruction needs an environment");
+    }
+
+    VarLenInstruction(PirType return_type, Value* env) : Super(return_type) {
+        assert(env);
+        static_assert(ENV > EnvAccess::None,
```
mhyee, line 357: Should the comparison be `>` or `!=`? (FixedLenInstruction uses the latter.)

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+                      "This instruction has no environment access");
+        push_arg(RType::env, env);
+    }
+};
+
+extern std::ostream& operator<<(std::ostream& out, Instruction::Id id);
+
+#define FLI(type, nargs, io, env)                                              \
```
mhyee, line 365: Off topic: I think there's a C++ proposal (metaclasses?) to handle this use case... because C++ has (or will have) everything.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+    type:                                                                      \
+  public                                                                       \
+    FixedLenInstruction<Tag::type, type, nargs, io, env>
+
+class FLI(LdConst, 0, Effect::None, EnvAccess::None) {
```
mhyee, line 370: I don't know if this is the right place (maybe `instruction_list.h` is better?), but I think there should be some documentation on each instruction. Then we don't have to read C++ code to figure out what an instruction does.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+            COMPILER_INSTRUCTIONS(V)
+#undef V
+            return static_cast<Instruction*>(v);
+        default: {}
+        }
+        return nullptr;
+    }
+};
+
+enum class EnvAccess : uint8_t {
```
mhyee, line 163: Explanation of the different `EnvAccess`es and `Effect`s would be useful.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+ *
+ * Instructions are either FixedLength or VariableLength.
+ *
+ * Every instruction is also a Value, and can therefore be used as an argument
+ * for other instructions.
+ *
+ * Instructions have an InstructionDescription, which gives us basic
+ * information about its effects and environment interactions.
+ *
+ * If an instruction needs an envrionment (ie. if its EnvAccess >= None), it
```
mhyee, line 29: Typo: envrionment -> environment

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        return v;
+    }
+
+    template <unsigned pos>
+    Value* arg() {
+        static_assert(pos < ARGS, "This instruction has less arguments");
+        return arg_[pos];
+    }
+
+    struct ArgTypesWithEnv : public std::array<PirType, ARGS> {
```
mhyee, line 259: I wonder if there's a way to modify these structs and constructors, because the double initializer braces `{{...}}` is quite ugly:

```c++
    ForSeqSize(Value* val)
        : FixedLenInstruction(PirType(RType::integer).scalar(),
                              {{PirType::val()}}, {{val}}) {}
```

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+
+    void eachCallArg(arg_iterator it) {
+        for (size_t i = 0; i < nCallArgs(); ++i) {
+            Value* v = callArgs()[i];
+            it(v);
+        }
+    }
+};
+
+typedef SEXP (*CCODE)(SEXP, SEXP, SEXP, SEXP);
```
mhyee, line 694: This is one place where using the type alias syntax might be better than typedef.

```c++
using CCODE = SEXP (*)(SEXP, SEXP, SEXP, SEXP)
```

Actually... it's just as ugly. I guess the advantage is that it pulls `CCODE` out from the rest of the type. That tripped me up the first time I was trying to parse the typedef...

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+class VLI(CallBuiltin, Effect::Any, EnvAccess::WriteKeepAlive) {
+  public:
+    const CCODE builtin;
+    int builtinId;
+
+    Value** callArgs() { return &args()[1]; }
+    const PirType* callTypes() { return &types()[1]; }
+    size_t nCallArgs() { return nargs() - 1; }
+
+    CallBuiltin(Value* e, SEXP builtin, const std::vector<Value*>& args);
```
mhyee, line 705: Is there a reason the implementation for this constructor is in `instruction.cpp` and not here? All the other constructors are implemented in this file.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+
+    void printArgs(std::ostream& out) override;
+};
+
+class VLI(CallSafeBuiltin, Effect::None, EnvAccess::None) {
+  public:
+    const CCODE builtin;
+    int builtinId;
+
+    CallSafeBuiltin(SEXP builtin, const std::vector<Value*>& args);
```
mhyee, line 722: Likewise with this constructor.

rir/src/compiler/pir/instruction.h
```diff
@@ -0,0 +1,787 @@
...
+        for (unsigned i = 0; i < varName.size(); ++i)
+            this->push_arg(PirType::any(), args[i]);
+    }
+
+    Value* parent() { return arg(0); }
+    void parent(Value* v) { arg(0, v); }
+
+    void printArgs(std::ostream& out) override;
+
+    Value** localVars() { return args() + 1; }
```
mhyee, line 749: `&args()[1]`?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <cassert>
+#include <iomanip>
+#include <sstream>
+
+extern "C" SEXP deparse1line(SEXP call, Rboolean abbrev);
+
+namespace rir {
+namespace pir {
+
+static size_t getMaxInstructionNameLength() {
```
mhyee, line 17: Too bad we can't do this statically... not sure `constexpr` is powerful enough.

rir/src/compiler/pir/instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+void Instruction::printRef(std::ostream& out) {
+    if (type == RType::env)
+        out << "e" << id();
+    else
+        out << "%" << id();
+};
+
+Instruction::Id Instruction::id() { return Id(bb()->id, bb()->indexOf(this)); }
+
+bool Instruction::unused() {
```
mhyee, line 70: Some more comments here would be useful, e.g. what do `unused`, `hasSingleUse`, etc. do.

rir/src/compiler/pir/instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+    // TODO: better solution?
+    if (tag == Tag::Branch || tag == Tag::Return)
+        return false;
+
+    return Visitor::check(bb(), [&](BB* bb) {
+        bool unused = true;
+        for (auto i : *bb) {
+            i->each_arg(
+                [&](Value* v, PirType t) { unused = unused && (v != this); });
```
mhyee, line 79: Isn't `PirType t` unused, so you could use the other iterator?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+        return unused;
+    });
+}
+
+Instruction* Instruction::hasSingleUse() {
+    size_t seen = 0;
+    Instruction* usage;
+    Visitor::check(bb(), [&](BB* bb) {
+        for (auto i : *bb) {
+            i->each_arg([&](Value* v, PirType t) {
```
mhyee, line 92: Again, `PirType t` looks like it's unused.

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
+#ifndef COMPILER_BB_H
+#define COMPILER_BB_H
+
+#include "pir.h"
+
+#include <set>
```
mhyee, line 6: Since nothing in the header actually needs `set`, move this include to the .cpp file?

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+
+    Instruction* last() {
+        assert(instr.size() > 0);
+        return instr.back();
+    }
+
+    BB* next0 = nullptr;
+    BB* next1 = nullptr;
+
+    bool jmp() { return next0 && !next1; }
```
mhyee, line 64: I'd prefer `isJmp()`.

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+    bool empty() { return instr.size() == 0; }
```
mhyee, line 65: And `isEmpty()`.

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+
+    Instrs::iterator remove(Instrs::iterator it);
+    Instrs::iterator moveToEnd(Instrs::iterator it, BB* other);
+    Instrs::iterator moveToBegin(Instrs::iterator it, BB* other);
+
+    void swap(Instrs::iterator);
+
+    void print(std::ostream& = std::cout);
+
+    Instrs::iterator begin() { return instr.begin(); }
```
mhyee, line 82: I guess these are really short implementations, but I think I still prefer keeping all the implementation code in the `.cpp` file, rather than putting some there and some in the header.

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+    Instrs::iterator end() { return instr.end(); }
+    size_t size() { return instr.size(); }
+    Instruction* at(size_t i) { return instr[i]; }
+
+    void gc();
+
+  private:
+    Instrs instr;
+    Instrs deleted;
```
mhyee, line 91: Comment explaining what `deleted` is for?

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+
+    Instrs::iterator begin() { return instr.begin(); }
+    Instrs::iterator end() { return instr.end(); }
+    size_t size() { return instr.size(); }
+    Instruction* at(size_t i) { return instr[i]; }
+
+    void gc();
+
+  private:
+    Instrs instr;
```
mhyee, line 90: `instrs`?

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+BB::BB(Code* fun, unsigned id) : id(id), fun(fun) {}
+
+void BB::print(std::ostream& out) {
+    out << "BB " << id << "\n";
+    for (auto i : instr) {
+        out << "  ";
+        i->print(out);
+        out << "\n";
+    }
+    if (next0 && !next1) {
```
mhyee, line 17: `if (jmp())` since we already have it, and the meaning is more obvious than `next0 && !next1`.

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+BB::Instrs::iterator BB::remove(Instrs::iterator it) {
+    deleted.push_back(*it);
+    return instr.erase(it);
+}
+
+BB::Instrs::iterator BB::moveToEnd(Instrs::iterator it, BB* other) {
+    Instruction* i = *it;
+    i->bb_ = other;
```
mhyee, line 40: Replace this line and the following with `other->append(i)`? Or `other->append(*it)`.

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+    other->instr.push_back(i);
+    return instr.erase(it);
+}
+
+BB::Instrs::iterator BB::moveToBegin(Instrs::iterator it, BB* other) {
+    Instruction* i = *it;
```
mhyee, line 46: Again, I'd probably use `other->insert` rather than directly inserting into `other`'s vector.

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+    void append(Instruction* i);
+
+    Instrs::iterator insert(Instrs::iterator it, Instruction* i);
+    void replace(Instrs::iterator it, Instruction* i);
+
+    Instrs::iterator remove(Instrs::iterator it);
+    Instrs::iterator moveToEnd(Instrs::iterator it, BB* other);
+    Instrs::iterator moveToBegin(Instrs::iterator it, BB* other);
+
+    void swap(Instrs::iterator);
```
mhyee, line 78: Comment here that `swap` swaps the instruction at the iterator with the following instruction.

And for consistency, probably want to include `i` in the declaration.

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+BB* BB::cloneInstrs(BB* src, unsigned id, Code* target) {
+    BB* c = new BB(target, id);
+    for (auto i : src->instr) {
+        Instruction* ic = i->clone();
+        ic->bb_ = c;
+        c->instr.push_back(ic);
+    }
+    c->next0 = c->next1 = nullptr;
```
mhyee, line 65: It doesn't clone the basic block successors?

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+BB::Instrs::iterator BB::insert(Instrs::iterator it, Instruction* i) {
+    auto itup = instr.insert(it, i);
+    i->bb_ = this;
+    return itup;
+}
+
+void BB::gc() {
+    // Catch double deletes
+    std::set<Instruction*> dup;
```
mhyee, line 83: `unordered_set`?

rir/src/compiler/pir/bb.cpp
```diff
@@ -0,0 +1,92 @@
...
+    i->bb_ = this;
+}
+
+BB::Instrs::iterator BB::insert(Instrs::iterator it, Instruction* i) {
+    auto itup = instr.insert(it, i);
+    i->bb_ = this;
+    return itup;
+}
+
+void BB::gc() {
```
mhyee, line 81: Is this method called anywhere else? Should it be private, or just put into the destructor?

And isn't this more for debugging? So should we only have this method if we compile with debug flags?

rir/src/compiler/pir/code.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+Code::Code() : entry(new BB(this, 0)) {}
+
+void Code::print(std::ostream& out) {
+    Visitor::run<true>(entry, [&out](BB* bb) { bb->print(out); });
+}
+
+Code::~Code() {
+    std::stack<BB*> toDel;
+    Visitor::run(entry, [&](BB* bb) { toDel.push(bb); });
```
mhyee, line 18: Might as well write `[&toDel]` since it's the only thing you're capturing.

rir/src/compiler/pir/function.h
```diff
@@ -0,0 +1,53 @@
...
+    Function(std::initializer_list<SEXP> a, std::initializer_list<Promise*> p)
+        : arg_name(a), default_arg(p) {}
+
+    Function(const std::vector<SEXP>& a, const std::vector<Promise*>& p)
+        : arg_name(a), default_arg(p) {}
+
+    Function(std::initializer_list<SEXP> a) : Function(a, {}) {}
+    Function(const std::vector<SEXP>& a) : Function(a, {}) {}
+
+    std::vector<SEXP> arg_name;
```
mhyee, line 32: `arg_names`?

rir/src/compiler/pir/function.h
```diff
@@ -0,0 +1,53 @@
...
+    std::vector<Promise*> default_arg;
```
mhyee, line 33: `default_args`?

rir/src/compiler/pir/function.h
```diff
@@ -0,0 +1,53 @@
...
+
+    std::vector<Promise*> promise;
```
mhyee, line 35: `promises`?

rir/src/compiler/pir/module.h
```diff
@@ -0,0 +1,22 @@
...
+#include <iostream>
+#include <vector>
+
+namespace rir {
+namespace pir {
+
+class Function;
+class Env;
+
+class Module {
```
mhyee, line 13: What is `Module` used for? How is it constructed?

rir/src/compiler/pir/module.h
```diff
@@ -0,0 +1,22 @@
...
+  public:
+    std::vector<Function*> function;
```
mhyee, line 15: `functions`?

rir/src/compiler/pir/bb.h
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * A BB has an id, which must remain stable (since visitors and analysis use
+ * the BB id as array indices).
+ *
+ */
+class BB {
+  public:
+    // The visitor relies on stable ids, do not renumber inside a visitor!!!
+    const unsigned id;
+    Code* fun;
```
mhyee, line 33: `function`?

rir/src/compiler/util/builder.h
```diff
@@ -0,0 +1,38 @@
...
+    BB* bb;
+    Builder(Function* fun, Promise* prom);
+    Builder(Function* fun, Env* enclos);
+
+    Value* buildDefaultEnv(Function* fun);
+
+    template <class T>
+    T* operator()(T* i) {
+        bb->append(i);
+        i->bb_ = bb;
```
mhyee, line 24: Shouldn't the above call to `append` already take care of this?

rir/src/compiler/util/builder.h
```diff
@@ -0,0 +1,38 @@
...
+    Function* function;
+    Code* code;
+    Value* env;
+    BB* bb;
+    Builder(Function* fun, Promise* prom);
+    Builder(Function* fun, Env* enclos);
+
+    Value* buildDefaultEnv(Function* fun);
+
+    template <class T>
```
mhyee, line 21: Why is this templated?

rir/src/utils/capture_out.h
```diff
@@ -0,0 +1,39 @@
...
+#define CAPTURE_OUT_H
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string>
+#include <unistd.h>
+
+namespace rir {
+
+class CaptureOut {
```
mhyee, line 11: More documentation please.

rir/src/utils/capture_out.h
```diff
@@ -0,0 +1,39 @@
+#ifndef CAPTURE_OUT_H
+#define CAPTURE_OUT_H
+
+#include <stdio.h>
```
mhyee, line 4: Thoughts on using `cstdio` and `cstdlib` so that the functions are in the `std` namespace?

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
+#ifndef COMPILER_VISITOR_H
+#define COMPILER_VISITOR_H
+
+#include "../pir/bb.h"
+#include "../pir/pir.h"
+
+#include <deque>
+#include <functional>
+#include <set>
```
mhyee, line 9: `unordered_set`?

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+        BB* cur = bb;
+        std::deque<BB*> todo;
+        std::vector<bool> done(64, false);
+        ok(done, cur);
+
+        while (cur) {
+            BB* next = nullptr;
+
+            if (cur->next0 && !isOk(done, cur->next0)) {
+                if (todo.empty())
```
mhyee, line 30: What's our coding convention on braces? (I prefer to always have braces.)

I hadn't commented on this before, but then a few lines down it's basically the exact same code, except with braces...

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            if (!next) {
+                if (!todo.empty()) {
+                    next = todo.front();
+                    todo.pop_front();
+                }
+            }
+
+            action(cur);
```
mhyee, line 53: I would have expected the action to occur first, before queuing up the successors. Any reason why this order is done instead?

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+            if (cur->next0 && !isOk(done, cur->next0)) {
+                if (todo.empty())
+                    next = cur->next0;
+                else
+                    q<STABLE>(todo, cur->next0);
+                ok(done, cur->next0);
+            }
+
+            if (cur->next1 && !isOk(done, cur->next1)) {
+                if (!next && todo.empty()) {
```
mhyee, line 38: Does this extra logic make a noticeable difference in performance? How common is this case?

For simplicity and clarity, I would be fine with just queuing up `cur->next0` and `cur->next1`, rather than being clever.

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+
+            action(cur);
+
+            cur = next;
+        }
+        assert(todo.empty());
+    }
+
+    template <bool STABLE = false>
+    static bool check(BB* bb, BBReturnAction action) {
```
mhyee, line 61: I'd find a name like `predicate` (or `pred` if you don't like typing) to be more descriptive than `action`.

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+            done.resize(done.size() * 2);
+        done[bb->id] = true;
+    }
+
+    static bool isOk(std::vector<bool>& done, BB* bb) {
+        return bb->id < done.size() && done[bb->id];
+    }
+
+    template <bool STABLE>
+    static void q(std::deque<BB*>& todo, BB* bb) {
```
mhyee, line 157: Rename to `enqueue`? I don't like super terse names like this...

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+        // For analysis random search is faster
+        if (STABLE || ((uintptr_t)bb & (1 << 5))) {
```
mhyee, line 159: I think we already complained about this -- it just looks like black magic.

Of course, [pseudo-random number generation in C++](http://en.cppreference.com/w/cpp/numeric/random) has every single bell and whistle you can think of, but a [Bernoulli distribution](http://en.cppreference.com/w/cpp/numeric/random/bernoulli_distribution) is probably what we want here. Too bad the simpler [randint](http://en.cppreference.com/w/cpp/experimental/randint) API is still experimental.

This PRNG business is probably overkill, though... I just don't like the bit twiddling magic, because why choose 5? How do we know it's randomly distributed?

rir/src/compiler/analysis/query.h
```diff
@@ -0,0 +1,24 @@
+#ifndef COMPILER_PIR_QUERY_H
+#define COMPILER_PIR_QUERY_H
+
+#include "../pir/pir.h"
+
+#include <set>
```
mhyee, line 6: `unordered_set`?

rir/src/compiler/analysis/query.cpp
```diff
@@ -0,0 +1,38 @@
...
+        for (auto i : *bb)
+            if (i->mightIO() || i->changesEnv())
+                return false;
+        return true;
+    });
+}
+
+std::set<Value*> Query::returned(Code* c) {
+    std::set<Value*> returned;
+    Visitor::run(c->entry, [&](BB* bb) {
```
mhyee, line 28: Might as well write `[&returned]` since it's the only thing being captured.

rir/src/compiler/util/visitor.h
```diff
@@ -0,0 +1,169 @@
...
+#include "../pir/pir.h"
+
+#include <deque>
+#include <functional>
+#include <set>
+
+namespace rir {
+namespace pir {
+
+class Visitor {
```
mhyee, line 14: How often do we have a visitor that iterates over all `Instruction`s, rather than `BB`s? Would it make sense to add these instruction visitors to the API?

rir/src/compiler/transform/insert_cast.cpp
```diff
@@ -0,0 +1,52 @@
...
+    std::cerr << "Cannot cast " << v->type << " to " << t << "\n";
+    assert(false);
+    return nullptr;
+}
+
+void InsertCast::operator()() {
+    Visitor::run(start, [&](BB* bb) { apply(bb); });
+}
+
+void InsertCast::apply(BB* bb) {
```
mhyee, line 28: Will making this `private` mess with the lambda capture above? Just wondering, because it seems like this is only ever used by the lambda.

mhyee: Also, I'm sufficiently tired that I would like a comment explaining what's happening here.

rir/src/compiler/util/cfg.h
```diff
@@ -0,0 +1,36 @@
+#ifndef PIR_CFG_H
+#define PIR_CFG_H
+
+#include "../pir/pir.h"
+#include <set>
```
mhyee, line 5: `unordered_set`?

rir/src/compiler/util/cfg.h
```diff
@@ -0,0 +1,36 @@
...
+#include <vector>
+
+namespace rir {
+namespace pir {
+
+class CFG {
+    typedef std::vector<BB*> BBList;
+
+  public:
+    std::vector<BBList> preds;
```
mhyee, line 15: Comment explaining the interpretation of this vector? What does it represent?

rir/src/compiler/util/cfg.h
```diff
@@ -0,0 +1,36 @@
...
+
+    size_t size() { return preds.size(); }
+    CFG(BB*);
+};
+
+class DominanceGraph {
+    typedef std::set<BB*> BBList;
+
+  public:
+    std::vector<BBList> doms;
```
mhyee, line 26: Comment explaining the interpretation of this vector? What does it represent?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+        return changed;
+    }
+};
+
+static void computeCfg(CFG* cfg, BB* start) {
+    std::stack<BB*> todo;
+
+    std::vector<BBSet> pred;
```
mhyee, line 35: `preds`?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+    std::vector<BB*> bbPtr;
```
mhyee, line 36: `bbPtrs`?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    pred.resize(start->id + 10);
```
mhyee, line 38: 10 here is a magic number. Any rationale for choosing it?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+    bbPtr.resize(start->id + 10);
+
+    size_t maxId = 0;
+
+    Visitor::run(start, [&](BB* cur) {
+        if (maxId < cur->id)
+            maxId = cur->id;
+
+        if (pred.size() <= cur->id) {
+            pred.resize(cur->id + 5);
```
mhyee, line 48: Any rationale for increasing the size by 5 instead of, e.g., doubling?

rir/src/compiler/util/cfg.cpp
mhyee: And what is this used for?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+            bbPtr.resize(cur->id + 5);
+        }
+
+        bbPtr[cur->id] = cur;
+
+        auto apply = [&](BB* bb) {
+            if (!bb)
+                return;
+
+            if (pred.size() <= bb->id) {
```
mhyee, line 58: All this manual management of vector sizes is beginning to annoy me. And every time we resize a vector, we have to copy everything over. I have to wonder if there's a better way.

Ordered maps? Unordered maps? Deques (which don't copy everything when resizing)? Maybe we should make a wrapper for vector, if we do this resizing a lot?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+        if (cur->next0 == nullptr && cur->next1 == nullptr) {
+            cfg->exits.push_back(cur);
+        }
+    });
+
+    cfg->preds.resize(maxId + 1);
+
+    for (size_t i = 0; i < maxId + 1; ++i) {
+        for (auto p : pred[i])
+            cfg->preds[i].push_back(bbPtr[p]);
```
mhyee, line 76: So it looks like there's some level of indirection, using indexes instead of the actual basic block. But why not use pointers to basic blocks (or maybe I'm being stupid because I'm tired).

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+static void computeDominanceGraph(DominanceGraph* cfg, BB* start) {
+    std::stack<BB*> todo;
+    todo.push(start);
+
+    std::vector<BBSet> dom;
```
mhyee, line 84: Naming convention: `doms`, `preds`, `bbPtrs`.

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+    std::vector<BBSet> pred;
+    std::vector<BB*> bbPtr;
+
+    dom.resize(start->id + 10);
+    pred.resize(start->id + 10);
+    bbPtr.resize(start->id + 10);
+
+    size_t maxId = 0;
+
+    do {
```
mhyee, line 94: A while loop feels more natural to me, especially since `todo` is already nonempty for the first iteration.

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+            maxId = cur->id;
+
+        if (pred.size() <= cur->id) {
+            pred.resize(cur->id + 5);
+            bbPtr.resize(cur->id + 5);
+        }
+
+        bbPtr[cur->id] = cur;
+
+        auto apply = [&](BB* bb) {
```
mhyee, line 54: To make it clearer what relation `bb` has to `cur`, maybe rename it to `succ`?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+            pred.resize(cur->id + 5);
+            bbPtr.resize(cur->id + 5);
+        }
+
+        BBSet& front = dom[cur->id];
+        bbPtr[cur->id] = cur;
+
+        todo.pop();
+
+        auto apply = [&](BB* bb) {
```
mhyee, line 110: `succ` instead of `bb`?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    cfg->preds.resize(maxId + 1);
+
+    for (size_t i = 0; i < maxId + 1; ++i) {
+        for (auto p : pred[i])
+            cfg->preds[i].push_back(bbPtr[p]);
+    }
+}
+
+static void computeDominanceGraph(DominanceGraph* cfg, BB* start) {
```
mhyee, line 80: Some comments explaining what this is doing would help. Took me a while to convince myself that this was correct.

rir/src/compiler/analysis/verifier.h
```diff
@@ -0,0 +1,16 @@
+#ifndef COMPILER_PIR_VERIFIER_H
+#define COMPILER_PIR_VERIFIER_H
+
+#include "../pir/pir.h"
+
+namespace rir {
+namespace pir {
+
+class Verify {
```
mhyee, line 9: What does the verifier verify?

rir/src/compiler/util/cfg.cpp
```diff
@@ -0,0 +1,160 @@
...
+                changed = true;
+            } else {
+                it++;
+            }
+        }
+        return changed;
+    }
+};
+
+static void computeCfg(CFG* cfg, BB* start) {
```
mhyee, line 32: Isn't `static` redundant if this is already in an anonymous namespace?

rir/src/compiler/util/cfg.cpp
mhyee, line 80: Isn't `static` redundant if this is already in an anonymous namespace?

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (bb->empty()) {
+            if (!bb->next0 && !bb->next1) {
+                std::cerr << "bb" << bb->id << " has no successor\n";
+                ok = false;
+            }
+        } else {
+            Instruction* last = bb->last();
+            if ((Branch::Cast(last))) {
+                if (!bb->next0 || !bb->next1) {
+                    std::cerr << "split bb" << bb->id
```
mhyee, line 58: It feels like the two checks should be flipped, or the error message should be updated.

Reading the code, this is the case where the last instruction *is* a branch, but the bb does not have two successors. So the message should probably be "bb with a branch instruction must have two successors".

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -0,0 +1,141 @@
...
+                              << " must end in branch\n";
+                    ok = false;
+                }
+            } else if ((Deopt::Cast(last)) || (Return::Cast(last))) {
+                if (bb->next0 || bb->next1) {
+                    std::cerr << "exit bb" << bb->id << " must end in return\n";
```
mhyee, line 64: Likewise, the message here should be "bb with deopt or return must have zero successors".

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -0,0 +1,141 @@
...
+                      << " but points to " << i->bb() << "\n";
+            ok = false;
+        }
+
+        Phi* phi = Phi::Cast(i);
+        i->each_arg([&](Value* v, PirType t) -> void {
+            Instruction* iv = Instruction::Cast(v);
+            if (iv) {
+                if (phi) {
+                    // TODO: what is a good condition for phi inputs?
```
mhyee, line 94: Stupid sanity check: input BBs to a phi node should have the phi node as a successor.

rir/src/compiler/pir_tests.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "analysis/verifier.h"
+#include "ir/Compiler.h"
+#include "pir/pir_impl.h"
+#include "pir_compiler.h"
+#include "util/visitor.h"
+
+namespace {
+using namespace rir;
+
+static pir::Module* compile(const std::string& inp) {
```
mhyee, line 14: Isn't `static` redundant when you're already in an anonymous namespace?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+#include "util/cfg.h"
+#include "util/visitor.h"
+
+#include <deque>
+#include <vector>
+
+namespace {
+
+using namespace rir::pir;
+typedef rir::pir::Function Function;
```
mhyee, line 19: Isn't this typedef redundant, since we already have `using namespace rir::pir`?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+    };
+
+    typedef std::pair<BB*, Value*> ReturnSite;
+    std::vector<ReturnSite> results;
+
+    void push(Value* v) { m.stack.push_back(v); }
+    Value* top() { return m.stack.back(); }
+    Value* at(size_t n) {
+        assert(n < m.stack_size());
+        return m.stack[m.stack_size() - n - 1];
```
mhyee, line 58: Why not push/pop to the front of the deque? All the stack operations on a deque are constant time.

That would simplify the stack access. No need to do the subtraction and risk the off-by-one.

You could even do `m.stack.at(n)`, which does a bounds check internally.

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+                    return false;
+            }
+            m(pc);
+            return true;
+        }
+    };
+
+    Value* operator()(bool addReturn);
+
+    void run(BC bc) {
```
mhyee, line 97: Documentation please. What's the basic idea for what this is doing?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+                BC::advance(&pc);
+                if (pc == end)
+                    return false;
+            }
+            m(pc);
+            return true;
+        }
+    };
+
+    Value* operator()(bool addReturn);
```
mhyee, line 95: So it was a bit odd to see all the implementations in the header file, and then this declaration being separate from its implementation.

I guess dependencies in C++ are hard...

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+    rir::Function* src;
+    rir::Code* srcCode;
+    TheCompiler* cmp;
+
+    CodeCompiler(Builder& b, rir::Function* src, rir::Code* srcCode,
+                 TheCompiler* cmp)
+        : b(b), src(src), srcCode(srcCode), cmp(cmp) {}
+
+    struct MachineState {
+        std::deque<Value*> stack;
```
mhyee, line 38: So both operands and instructions go on the stack?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+            push(b(new MkArg(prom, val, b.env)));
+            break;
+        }
+        case Opcode::static_call_stack_: {
+            unsigned n = bc.immediate.call_args.nargs;
+            rir::CallSite* cs = bc.callSite(srcCode);
+            SEXP target = rir::Pool::get(*cs->target());
+
+            std::vector<Value*> args(n);
+            for (size_t i = 0; i < n; ++i)
```
mhyee, line 222: Counting from `n-1` to `0` is probably better, if we want the top of the stack to be the back of the deque.

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+        case Opcode::put_:
+            x = top();
+            for (size_t i = 0; i < bc.immediate.i - 1; ++i)
+                set(i, at(i + 1));
+            set(bc.immediate.i, x);
+            break;
+
+        // TODO implement!
+        // (silently ignored)
```
mhyee, line 340: Silently ignoring things makes me nervous. Can we print debugging stuff to stderr?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+        case Opcode::set_names_:
+        case Opcode::names_:
+        case Opcode::force_:
+
+        // Invalid opcodes:
+        case Opcode::label:
+        case Opcode::invalid_:
+        case Opcode::num_of:
+
+        // Opcodes handled elsewhere
```
mhyee, line 362: Can you elaborate on where these are handled?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+        }
+        // Create mergepoints
+        for (auto m : incom) {
+            if (std::get<1>(m).size() > 1) {
+                mergepoint[std::get<0>(m)] = MachineState();
+            }
+        }
+    }
+
+    bool doMerge(Opcode* trg) {
```
mhyee, line 418: What is this function doing?

rir/src/compiler/pir_compiler.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+using namespace rir::pir;
+typedef rir::pir::Function Function;
+typedef rir::Opcode Opcode;
+typedef rir::BC BC;
+typedef rir::RList RList;
+
+class TheCompiler;
+
+class CodeCompiler {
```
mhyee, line 26: Just in general, I had a harder time following this file, compared to the others. Maybe I just got tired, because this is the last file I'm reviewing.

mhyee (commented):
You submit a monster PR, I submit a monster review. :)

Overall, I'd say it looks pretty good, and wasn't too difficult to follow. I know that a number of my comments will have been addressed or made obsolete by #86, but I guess everything else could be incorporated into the refactoring. (I still need to review the optimizations.)

Also, I ended up reviewing the verifier, before realizing there was a newer version.

I think most of my comments can be grouped into a few categories:

* Being rusty and confused about C++.
* Complaining about naming convention and code style.
* Asking for more documentation. This is a general comment, and I didn't leave a remark at every place I thought needed more documentation...

I also had a hard time understanding what the RIR to PIR compiler was doing. I don't know if scheduling a codewalk would be worth the time and effort, but at the very least, it should have some more documentation.

rir/src/utils/Bitset.h
o-: so this does a bit more. It maps an enum into the bitset. So the idea is that you can do:

```
enum class Bla {
  one,
  two
};
BitSet<uint8_t, Bla> bs;
bs = bs | Bla::two;    // ie. implicit for :  bs = bs | BitSet<uint8_t, Bla>(Bla::two)
```

Which will convert the enum `two` to a bitset, with the `(uint8_t)Bla::two`-th bit set. Ie. it's a safe way of doing: `bs = bs | (1 << (uint8_t)Bla::two)`

rir/src/compiler/pir/pir_impl.h
o-: not sure if I get the question.

This is just the "all pir stuff" header file. So you should not include in other headers, since that gives cyclic dependency.

Why is type not there, probably since other included headers depend on type. but it could be added.

rir/src/compiler/pir/instruction_list.h
o-: ok, currently it is enforced in the sense, that when you add an instruction to instruction.h you need to pass an Instruction tag to the Instruction constructor. Is that too weak? will improve the doku

rir/src/compiler/pir/type.h
o-: nice, didn't think of that one

rir/src/compiler/pir/type.h
o-: we could also remove it. It's just super useful for debugging uninitialized values...

rir/src/compiler/pir/type.cpp
o-: `>=` is overloaded for sub-type check. so this means "if I am a vector of any sort"

rir/src/compiler/pir/type.h
o-: I guess I was worried about perf, since this will be in the most inner loop of all analysis that looks at types of values

rir/src/compiler/pir/type.h
o-: suggestions?

rir/src/compiler/pir/instruction.h
o-: I agree it's messed up, but it was the best solution I found. I'll look at it again

rir/src/compiler/pir/instruction.h
o-: if you declare an instruction with environment (ie `ENV != EnvAccess::None`) then you need to set the environment (ie. invoke a constructor that takes env as an argument) and not this constructor.

rir/src/compiler/pir/instruction.h
o-: well, it's the same since None is 0 :) but i will make it consistent

rir/src/compiler/pir/instruction.h
o-: awesome :)

rir/src/compiler/pir/instruction.h
o-: ok, my hope was that basically the template parameters would give you all the documentation about an instruction you want ...

rir/src/compiler/pir/instruction.h
o-: believe me I tried everything I could think of for those damn constructors :)

rir/src/compiler/pir/bb.cpp
o-: No, the problem is if you remove an instruction, you can't delete it right away, since some analysis might still have a pointer to the instruction (since normally for analysis the results are just `map<Instruction*, AnalysisResult>`).

So when you delete an instruction (e.g. `bb->replace(...)`) I put it on the deleted list and between compiler passes you can call gc. And if not, then the destructor takes the rest.

rir/src/compiler/util/builder.h
o-: So you can write

`LdVar* ld = builder(new LdVar(...`

and not:

`LdVar* ld = (LdVar*) builder(new LdVar(...`

rir/src/compiler/util/visitor.h
o-: since the visitor is used everywhere, I think it is well justified to be a bit clever here

rir/src/compiler/util/visitor.h
o-: https://xkcd.com/221/

rir/src/compiler/util/visitor.h
o-: good point

rir/src/compiler/util/cfg.cpp
o-: I don't know :) I just put "static, extern, inline" in that order until it compiles

o-:
Hey, thanks a lot for your comments, very helpful.

I will try to go over everything, but will leave the pir_compiler.cpp, since guido is working on that.

I agree that the actual translation is the ugliest part and while I put quite some time into the design of pir, the rir2pir translation was rather rushed. (unfortunately it was also easy enough, that it did not break, even though it was poorly written, so I did not have the incentive to improve it)

rir/src/compiler/pir/bb.h
o-: that's clang format

rir/src/compiler/util/visitor.h
o-: once we need it...

PR #80, opened 2018-02-27
PIR optimizations

o-:
Based on pir_1 branch this adds 3 optimization pass commits

### DIFF
--- a/rir/src/compiler/analysis/abstract_value.h
+++ b/rir/src/compiler/analysis/abstract_value.h
@@ -1,0 +1,221 @@
+#ifndef PIR_ABSTRACT_VALUE_H
+#define PIR_ABSTRACT_VALUE_H
+
+#include "../pir/pir.h"
+
+#include <algorithm>
+#include <set>
+#include <unordered_map>
+
+namespace rir {
+namespace pir {
+
+/*
+ * Captures an abstract PIR value.
+ *
+ * Vals is the set of potential candidates. If we don't can't tell what the
+ * possible values are, then we set "unknown" (ie. we taint the value). This is
+ * the top element of our lattice.
+ *
+ */
+struct AbstractValue {
+    bool unknown = false;
+
+    struct ValOrig : public std::pair<Value*, Instruction*> {
+        ValOrig(Value* v, Instruction* o)
+            : std::pair<Value*, Instruction*>(v, o) {}
+        Value* val() const { return first; }
+        Instruction* orig() const { return second; }
+    };
+
+    std::set<ValOrig> vals;
+
+    PirType type = PirType::bottom();
+
+    AbstractValue();
+    AbstractValue(Value* v, Instruction* origin);
+
+    static AbstractValue tainted() {
+        AbstractValue v;
+        v.taint();
+        return v;
+    }
+
+    void taint() {
+        vals.clear();
+        unknown = true;
+        type = PirType::any();
+    }
+
+    bool isUnknown() const { return unknown; }
+
+    bool singleValue() {
+        if (unknown)
+            return false;
+        return vals.size() == 1;
+    }
+
+    bool merge(const AbstractValue& other);
+
+    void print(std::ostream& out = std::cout);
+};
+
+static Value* UnknownParent = (Value*)-1;
+static Value* UninitializedParent = nullptr;
+static Function* UnknownFunction = (Function*)-1;
+
+/*
+ * An AbstracEnvironment is a static approximation of an R runtime Envrionment.
+ *
+ * A key notion is, when an environment leaks. A leaked environment describes
+ * an environment, that is visible to an unknown context. This means, that it
+ * can be inspected and manipulated by code we can't statically analyze.
+ *
+ * Typically an analysis will need to mark an environment leaked, when we call
+ * a (statically) unknown function. The reason is that the callee can always
+ * inspect our environment through sys.parent().
+ *
+ * For inter-procedural analysis we can additionally keep track of closures.
+ */
+template <class AV>
+struct AbstractEnvironment {
+    std::unordered_map<SEXP, AV> entries;
+    std::unordered_map<Value*, Function*> functionPointers;
+
+    Value* parentEnv = UninitializedParent;
+
+    bool leaked = false;
+    bool tainted = false;
+
+    void taint() {
+        tainted = true;
+        for (auto& e : entries) {
+            e.second.taint();
+        }
+    }
+
+    void set(SEXP n, Value* v, Instruction* origin) {
+        entries[n] = AV(v, origin);
+    }
+
+    void print(std::ostream& out = std::cout) {
+        for (auto e : entries) {
+            SEXP name = std::get<0>(e);
+            out << "   " << CHAR(PRINTNAME(name)) << " -> ";
+            AV v = std::get<1>(e);
+            v.print(out);
+            out << "\n";
+        }
+        out << "\n";
+    }
+
+    const AV& get(SEXP e) const {
+        static AV t = AV::tainted();
+        if (entries.count(e))
+            return entries.at(e);
+        return t;
+    }
+
+    bool merge(AbstractEnvironment& other) {
+        bool changed = false;
+        if (!leaked && other.leaked)
+            changed = leaked = true;
+        if (!tainted && other.tainted)
+            changed = tainted = true;
+
+        std::set<SEXP> keys;
+        for (auto e : entries)
+            keys.insert(std::get<0>(e));
+        for (auto e : other.entries)
+            keys.insert(std::get<0>(e));
+        for (auto n : keys) {
+            // if this is not the first incomming edge and it has more entries
+            // we are in trouble.
+            if (!entries.count(n)) {
+                entries[n].taint();
+                changed = true;
+            } else if (entries[n].merge(other.get(n))) {
+                changed = true;
+            }
+        }
+
+        std::set<Value*> fps;
+        for (auto e : functionPointers)
+            fps.insert(std::get<0>(e));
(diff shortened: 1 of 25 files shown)


o-:
tbr

rir/src/compiler/transform/bb.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <unordered_map>
+
+namespace rir {
+namespace pir {
+
+BB* BBTransform::clone(size_t* id_counter, BB* src, Code* target) {
+    std::vector<BB*> bbs;
+
+    // Copy instructions
+    std::unordered_map<Value*, Value*> relocation_table;
```
mhyee, line 14: What is this mapping? Instructions to instructions?

rir/src/compiler/transform/bb.cpp
```diff
@@ -0,0 +1,100 @@
...
+        BB* theClone = BB::cloneInstrs(bb, *id_counter, target);
+        assert(bb->size() == theClone->size());
+        if (bb->id >= bbs.size())
+            bbs.resize(bb->id + 5);
+        bbs[bb->id] = theClone;
+        for (size_t i = 0; i < bb->size(); ++i)
+            relocation_table[bb->at(i)] = theClone->at(i);
+    });
+
+    // Fixup CFG
```
mhyee, line 26: Please elaborate what "fixup CFG" means. I'm assuming that at this point, we've copied all the basic blocks and their instructions... but the new basic blocks still point to the old basic blocks as successors. So this is to update all the successors in the new basic blocks we just copied.

rir/src/compiler/transform/bb.cpp
```diff
@@ -0,0 +1,100 @@
...
+        if (bb->next0 != nullptr)
+            return;
+
+        assert(bb->next1 == nullptr);
+        if (Deopt::Cast(bb->last()))
+            return;
+
+        Return* ret = Return::Cast(bb->last());
+        assert(ret);
+        assert(!found);
```
mhyee, line 91: This seems like a weird thing to assert, because there's no code that could have assigned something to `found`.

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -0,0 +1,154 @@
...
+#define PIR_GENERIC_STATIC_ANALYSIS
+
+#include "../pir/bb.h"
+#include "../pir/function.h"
+#include "../pir/instruction.h"
+#include "../util/cfg.h"
+#include "../util/visitor.h"
+#include "R/r.h"
+
+#include <functional>
```
mhyee, line 11: Are any of these includes used in this file?

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -0,0 +1,154 @@
...
+    }
+    StaticAnalysis(BB* entry, const AbstractState& initialState)
+        : entry(entry), cfg(entry) {
+        mergepoint.resize(cfg.size());
+        mergepoint[entry->id].push_back(initialState);
+    }
+
+    virtual void apply(AbstractState&, Instruction*) const = 0;
+
+    typedef std::function<void(const AbstractState&, Instruction*)> collect_res;
```
mhyee, line 58: Can we use TitleCase for types, so `CollectRes`?

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -0,0 +1,154 @@
...
+
+    template <bool before = true>
+    void collect(collect_res collect) {
```
mhyee, line 61: Ugh, don't like how the function and parameter have the same name.

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -0,0 +1,154 @@
...
+ * implemented, which supplies the implementation for every instruction. Apply
+ * is supposed to modify the abstract state, but not (!) the analysis itself
+ * (that is why it is marked const). The reason is, that after we reached a
+ * fixed-point, it should be possible to reconstruct the state of the analysis
+ * at every instruction. To do so, a dominating state is loaded from
+ * "mergepoint" and then "apply" is used to seek to the desired instruction
+ * pointer (see "collect" for an example).
+ *
+ */
+template <class AbstractState>
```
mhyee, line 36: What is the "type" of `AbstractState`? What operations should it support?

rir/src/compiler/analysis/generic_static_analysis.h
mhyee: And what does `collect` do?

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -0,0 +1,154 @@
...
+class StaticAnalysis {
```
mhyee, line 37: Just in general, I'm having a harder time following this file. Maybe I just need to look at a concrete static analysis?

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -0,0 +1,221 @@
+#ifndef PIR_ABSTRACT_VALUE_H
+#define PIR_ABSTRACT_VALUE_H
+
+#include "../pir/pir.h"
+
+#include <algorithm>
+#include <set>
```
mhyee, line 7: `unordered_set`

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -0,0 +1,221 @@
...
+            return false;
+        return vals.size() == 1;
+    }
+
+    bool merge(const AbstractValue& other);
+
+    void print(std::ostream& out = std::cout);
+};
+
+static Value* UnknownParent = (Value*)-1;
```
mhyee, line 63: I've actually never seen this convention before, and it looks pretty ugly... Why can't we use `nullptr`?

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -0,0 +1,221 @@
...
+static Value* UninitializedParent = nullptr;
+static Function* UnknownFunction = (Function*)-1;
+
+/*
+ * An AbstracEnvironment is a static approximation of an R runtime Envrionment.
```
mhyee, line 68: Typo: AbstracEnvironment -> Abstrac**t**Environment

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -0,0 +1,221 @@
...
+        if (!tainted && other.tainted)
+            changed = tainted = true;
+
+        std::set<SEXP> keys;
+        for (auto e : entries)
+            keys.insert(std::get<0>(e));
+        for (auto e : other.entries)
+            keys.insert(std::get<0>(e));
+        for (auto n : keys) {
+            // if this is not the first incomming edge and it has more entries
```
mhyee, line 132: Typo: incom**m**ing -> incoming

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -0,0 +1,221 @@
+#ifndef PIR_ABSTRACT_VALUE_H
+#define PIR_ABSTRACT_VALUE_H
+
+#include "../pir/pir.h"
+
+#include <algorithm>
```
mhyee, line 6: Should probably move this to the .cpp file (I don't think it's needed here).

rir/src/compiler/analysis/scope.h
```diff
@@ -0,0 +1,35 @@
...
+ * Heavyweight scope analysis.
+ *
+ * This analysis soundly traces all stores and gives a static approximation for
+ * all loads (`LdVar`).
+ *
+ */
+class ScopeAnalysis {
+  public:
+    typedef AbstractEnvironment<AbstractValue> AbstractEnv;
+    typedef AbstractEnvironmentSet<AbstractEnv, AbstractValue> AbstractState;
```
mhyee, line 25: I think I've reached [semantic satiation](https://en.wikipedia.org/wiki/Semantic_satiation) when it comes to "abstract"...

rir/src/compiler/analysis/scope.h
```diff
@@ -0,0 +1,35 @@
+#ifndef PIR_SCOPE_ANALYSIS_H
+#define PIR_SCOPE_ANALYSIS_H
+
+#include "../analysis/generic_static_analysis.h"
+#include "../pir/pir.h"
+#include "abstract_value.h"
+
+#include <algorithm>
+#include <set>
```
mhyee, line 9: `unordered_set`

rir/src/compiler/analysis/scope.h
```diff
@@ -0,0 +1,35 @@
+#ifndef PIR_SCOPE_ANALYSIS_H
+#define PIR_SCOPE_ANALYSIS_H
+
+#include "../analysis/generic_static_analysis.h"
+#include "../pir/pir.h"
+#include "abstract_value.h"
+
+#include <algorithm>
```
mhyee, line 8: Don't think this is needed here.

rir/src/compiler/analysis/abstract_value.h
mhyee: (And there are a whole bunch of other occurrences in this PR that I didn't comment on.)

mhyee (commented):
I think I'm just going to submit this review in its current state. I've already spent enough time on this, and then paged it out and back in, and then tried to go over it too quickly.

The analysis framework is quite involved, and I wasn't able to follow it completely. I hoped looking at the actual concrete analyses would help, but those were also quite involved.

I think this is fine for now (I left some very superficial comments), but later, when I have more time, I'd like to sit down and get a better understanding of what's going on.

PR #84, opened 2018-03-06
Environment handling in RIR

JanJecmen:
This PR introduces two main changes in preparation for PIR and some minor code improvements.
1) Environments are now created on the callee side of a call. This is accomplished by using two classes called proxies to hold either the work that is already done, or data to make it possible to perform the work at a later time.
2) Some initial versions of environment manipulation instructions: making new and getting and setting current env.

I tried to make it nice, but I fear I failed... comments on how to achieve the delayed creation in a better way are welcome.

### DIFF
--- a/rir/src/ir/insns.h
+++ b/rir/src/ir/insns.h
@@ -27,7 +27,7 @@ DEF_INSTR(get_env_, 0, 0, 1, 1)
 /**
  * set_env_:: make tos environment the current env
  */
-DEF_INSTR(set_env_, 0, 1, 0, 1)
+DEF_INSTR(set_env_, 0, 1, 0, 0)
 
 /**
  * ldfun_:: take immediate CP index of symbol, find function bound to that name
--- a/rir/src/ir/insns.h
+++ b/rir/src/ir/insns.h
@@ -10,12 +10,28 @@
 DEF_INSTR(invalid_, 0, 0, 0, 0)
 
 /**
-  * nop_:: do nothing.
-  */
+ * nop_:: do nothing.
+ */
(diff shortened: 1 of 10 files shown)


rir/src/ir/insns.h
```diff
@@ -10,12 +10,28 @@
...
+
+/**
+ * get_env_:: push current env to tos
+ */
+DEF_INSTR(get_env_, 0, 0, 1, 1)
+
+/**
+ * set_env_:: make tos environment the current env
+ */
+DEF_INSTR(set_env_, 0, 1, 0, 1)
```
mhyee: I could be mistaken, but should the last `1` be a `0` because `set_env_` is impure?

mhyee (commented):
Don't really have comments -- the code looks reasonable to me.

I do have a question: maybe I missed something, but why do we need to delay environment creation?

rir/src/ir/insns.h
JanJecmen: You are probably right.
Maybe I should go over all the instructions and see about their purity, but I am not sure I understand the semantics exactly. Another thing is that it's used only in rir dataflow analysis, inlining and cleanup, and I think those will go away with pir.

JanJecmen:
This is in preparation for pir, where the environment is created explicitly, manipulated just as any other value, and possibly eliminated. In that case we don't want to create it in the caller.

rir/src/ir/insns.h
mhyee: For now, let's just ignore the purity of the other instructions, especially if we're not sure about the semantics. And I agree; it'll probably go away with PIR.

PR #86, opened 2018-03-14
[wip] Refactoring

o-:


rir/src/compiler/pir/module.h
```diff
@@ -3,19 +3,26 @@
 
 #include <iostream>
 #include <vector>
+#include "../../runtime/Function.h"
```
o-: should not be necessary!

o-:
I like the direction, thanks a lot!

Update: Long term we should think of some good way of handling header dependencies. Short term, whatever works. Maybe avoid super complicated dependencies.


The rest of my comment can be ignored, I leave it here for documentation.

One question: why pulling out the `stack_machine`? It seems to me that we only need it in the compiler, so keeping it in the cpp file seems better to me.

Include dependency hell is a real thing in C++. In my PR I tried to only include "pir/pir.h" in headers (and "pir/pir_impl.h" in .cpp files). The former contains only forward declarations. So no cross-dependencies between header files.

The only exceptions I made was when I wanted to inherit (in which case you need the declaration for the superclass) and if I wanted to use templates (in which case you are better of to provide the whole implementation in the .h file).

But maybe that was a stupid policy to try to adhere to in the first place :) At least I just saw that the google style guide argues against forward declarations (https://google.github.io/styleguide/cppguide.html#Forward_Declarations). Now I am not sure anymore. It's probably also not the most important discussion.

o-:
somehow this PR does not run on travis. I expect it is because i created it for you, that might confuse it.

Maybe you'll have to close it and create a new one.

charig:
I will close this PR and open it again to try making it work in Travis.

PR #87, opened 2018-03-19
Cleanup monster

o-:
This PR addresses the excelent review by @mhyee for the tbr #79 .

It is best viewed by individual commit


o-:
most commits should be mostly uncontroversial.

5c7c09a  changes the instruction implementation quite a bit. In particular for the arguments, instead of two arrays/vectors with value and types, it is now one array/vector of `std::pair(Value*, PirType)`. I find that this simplifies the api's in many places. In particular the eachArg iterators are nicer.

7c2ebc5 does not address the bit about the performance optimization (avoiding to put the next bb on the queue, if there is only one todo), since the visitor is performance critical.

f7c4bfb rewrites the cfg analysis quite a bit, it is largely a simplification of the approach.

rir/src/utils/EnumSet.h
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    void reset(const Element& e) {
+        assert(boundscheck(e));
+        set_.reset(static_cast<size_t>(e));
+    }
+
+    bool includes(const EnumSet& s) const { return (s.set_ & set_) == s.set_; }
+
+    void operator=(const EnumSet& o) { set_ = o.set_; }
```
mhyee: Shouldn't the default assignment operator be good enough?

rir/src/utils/EnumSet.h
```diff
@@ -0,0 +1,118 @@
...
+
+    bool operator==(const Element& t) const { return singleton(t) == set_; }
+
+    bool operator==(const EnumSet& s) const { return set_ == s.set_; }
+
+    EnumSet operator|(const EnumSet& s) const { return EnumSet(s.set_ | set_); }
+
+    EnumSet operator|(const Element& t) const { return *this | EnumSet(t); }
```
mhyee, line 55: What's the convention for using `singleton(t)` vs `EnumSet(t)`?

rir/src/utils/EnumSet.h
```diff
@@ -0,0 +1,118 @@
...
+
+    bool empty() const { return set_ == Store(); }
```
mhyee: `return set_.none();`

rir/src/compiler/pir/type.h
```diff
@@ -300,46 +297,34 @@ inline std::ostream& operator<<(std::ostream& out, RType t) {
 
 inline std::ostream& operator<<(std::ostream& out, PirType t) {
     if (!t.isRType()) {
-        if (t.t_.n.size() == 0) {
+        if (t.t_.n.empty() == 0) {
```
mhyee: Should just be `t.t_.n.empty()`

rir/src/compiler/pir/singleton_values.h
```diff
@@ -13,8 +13,10 @@ namespace pir {
 
 template <typename T>
 class SingletonValue : public Value {
```
mhyee, line 15: I think we still want to delete the copy constructor and assignment operators?

rir/src/compiler/pir/instruction.h
```diff
@@ -728,25 +684,35 @@ class VLI(MkEnv, Effect::None, EnvAccess::Capture) {
   public:
     std::vector<SEXP> varName;
 
-    typedef std::function<void(SEXP name, Value* val)> local_it;
+    typedef std::function<void(SEXP name, Value* val)> LocalVarIt;
+    typedef std::function<void(SEXP name, InstrArg&)> MutblLocalVarIt;
```
mhyee: I think I prefer `MutableLocalVarIt`. `Var` and `It` are pretty common, `Mutbl` less so.

rir/src/compiler/pir/instruction.h
```diff
@@ -137,16 +136,19 @@ class Instruction : public Value {
     }
 };
 
+// EnvAccess specifies if an instruction has an environment argument
+// (ie. EnvAccess > None), and if yes, what kind of interactions with that
+// environment can happen.
 enum class EnvAccess : uint8_t {
     None,
+    // No usage, besides looking at the reference
```
mhyee: Which access is this comment referring to? I would have thought `Capture`, but "No usage" sounds like `None`.

rir/src/compiler/util/visitor.h
```diff
@@ -143,20 +144,24 @@ class Visitor {
...
+    static bool isDone(std::vector<bool>& done, BB* bb) {
         return bb->id < done.size() && done[bb->id];
     }
 
     template <bool STABLE>
-    static void q(std::deque<BB*>& todo, BB* bb) {
+    static void enqueue(std::deque<BB*>& todo, BB* bb) {
         // For analysis random search is faster
-        if (STABLE || ((uintptr_t)bb & (1 << 5))) {
+        if (coin(gen)) {
```
mhyee: Don't we still want to keep `STABLE`?

Or maybe it's easier to just specialize the template.

o-:
So I quickly went through this with @mhyee yesterday. Does anyone want to have a second look before I merge?

mhyee:
I have a review coming later tonight. Less than 10 comments.

o-:
awesome!

I just rebased to see if the checks now pass. I hope that does not confuse your review draft

mhyee (approved):
Looks good! It's a significant improvement over the previous version. I just had a few minor comments.

rir/src/compiler/util/visitor.h
o-: good catch

PR #89, opened 2018-03-19
[travis] disable cache

o-:
something breaks with the cache on...


o-:
the failing linux test was due to the curl test being stupid.
the failing osx was due to gcc update, clearing the cache fixed it.
so I think we can keep caching enabled.

PR #92, opened 2018-03-21
unify the implementation of all BB Visitors

o-:
There are 3 parameters for BB visitors:

1. performing an action (ie. unconditionally run on all BB's) or a
   predicate (ie. run until the first action fails).
2. rely on BB ids not to change during the visit
3. stable or random visit order

This commit abstracts the visit algorithm and makes those 3 things
template parameters.

(1) is reflected in the 'run' vs. 'check' api and implemented
by a different 'ActionKind' parameter to 'runGeneric'.

(2) is realized by the 'Marker' parameter and the two implementations
'IDMarker' and 'PointerMarker'. Finally, (3) is given by the
'STABLE' bool.

To avoid clients having to deal with the template parameters, there
are predefined versions 'Visitor', 'BreadthFirstVisitor', and
'UnstableIDsVisitor'.


o-:
So this is one of the examples, where I personally like the results, but I had to use some tricks to make it work, and it might actually be a regression in ease of understanding... What do you think?

rir/src/compiler/util/visitor.h
```diff
@@ -12,26 +12,83 @@
...
+};
+
+template <>
+struct PredicateWrapper<BBActionPredicate> {
+    const BBActionPredicate action;
+    bool operator()(BB* bb) const { return action(bb); }
+};
+
+/*
+ * Helpers to remember which BB has already been visited. There is a fast
```
mhyee, line 44: Would be useful to point out that IDMarker is the fast version and PointerMarker is the slow version.

mhyee (approved):
I'm happy to have a simpler API even if it means the implementation is slightly more complicated.

PR #94, opened 2018-03-27
Refactoring

charig:


rir/src/compiler/pir/module.h
```diff
@@ -3,19 +3,31 @@
...
+      : srcFunction(src), srcCode(src->body()), dstFunction(dst) {}
+  IRTransformation(rir::Function* src, rir::Code* srcCode, pir::Function* dst)
+      : srcFunction(src), srcCode(srcCode), dstFunction(dst) {}
+};
+
 
 class Module {
   public:
-    std::vector<Function*> function;
+    std::vector<IRTransformation*> functions;
```
o-, line 26: I don't understand this. Why would the module contain IRTransformations? What if I perform two transformations, would it be twice in the list?

Also why keep srcFunction and srcCode, srcCode is just `srcFunction->code()` not?

rir/src/compiler/pir/stack_machine.h
```diff
@@ -0,0 +1,48 @@
...
+#include "runtime/Function.h"
+#include "../util/builder.h"
+#include <queue>
+
+namespace rir {
+namespace pir {
+
+typedef std::pair<BB*, Value*> ReturnSite;
+
+class StackMachine {
```
o-, line 15: so what I somehow don't like about this api, that there are no invariants enforced on this object. There are very natural invariants (like pc is between start and end), but it is very unclear when they start to hold (since you set all the pieces individually) and they are not checked.

I know that the original MachineState did not have invariants either. But, somehow since you lifted the importance of this class (e.g. by moving it out of the compiler into it's own existence) it seems to demand more scrutiny.

rir/src/compiler/pir/stack_machine.h
```diff
@@ -0,0 +1,48 @@
...
+      size_t stack_size();
+      Value* front();
+      Value* top();
+      void push(Value*);
+      Value* pop();
+      BC getCurrentBC();
+      
+    private:
+      std::deque<Value*> stack;
+      BC currentBC;
```
o-, line 37: I would drop this and decode from pc on demand. Keeping currentBC and pc in sync is an unnecessary source of errors.

rir/src/compiler/pir/stack_machine.h
```diff
@@ -0,0 +1,48 @@
...
+      Opcode* pc;
+      pir::BB* entry = nullptr;
+      Value* at(size_t);
+      void set(size_t n, Value* v);
+      void advancePC();
+      void decodeCurrentBytecode();
+      void consumeBC(Opcode* end);
```
o-, line 44: the name is confusing, since it seems to suggest that this function consumes `end`.

I would prefer if you check from the outside. Eg. in rir2pir `do {... ; machine.advance(); } while (machine.pc() != end);`

rir/src/compiler/pir/stack_machine.h
```diff
@@ -0,0 +1,48 @@
...
+typedef std::pair<BB*, Value*> ReturnSite;
+
+class StackMachine {
+    public:
+      StackMachine() : stack(), currentBC(), pc(), entry() {}
+
+      Opcode* getPC();
+      void setPC(Opcode*);
+      pir::BB* getEntry();
+      void setEntry(pir::BB*);
```
o-, line 22: it seems to me this (and `setPc`) should only be called once, can we enforce this?

(e.g. put it into constructor, which might need some wrangling, or at least have an `init` method that can only be called once)

rir/src/compiler/translations/pir_translator.cpp
```diff
@@ -0,0 +1,7 @@
+#include "pir_translator.h"
+
+namespace rir {
+    bool PirTranslator::getVerbose() { return verbose;}
```
o-, line 4: I guess we disagree on that one, but I kind of dislike setters and getters that do nothing. I don't see the point. But since you seem to like them, let's just go with it ;)

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+pir::Function* Rir2Pir::compileFunction(SEXP function2Compile) {
+    assert(isValidClosureSEXP(function2Compile));
+    DispatchTable* tbl = DispatchTable::unpack(BODY(function2Compile));
+    auto formals = RList(FORMALS(function2Compile));
+
+    std::vector<SEXP> fmls;
+    for (auto it = formals.begin(); it != formals.end(); ++it) {
+        fmls.push_back(it.tag());
+    }
+    return this->compileFunction(tbl->first(), fmls);
```
o-, line 45: `this` is not neccessary

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+pir::Function* Rir2Pir::compileFunction(Function* function2Compile,
+                                        std::vector<SEXP> fmls) {
+    pir::Function* pirFunction = this->compileInnerFunction(function2Compile, fmls);
+    this->optimizeFunction(pirFunction);
```
o-, line 51: `this` is not necessary, (more instances below)

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -0,0 +1,58 @@
...
+        : PirTranslator(), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Module* module)
+        : PirTranslator(module), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Builder* builder)
+        : PirTranslator(), mergepoint(), state(), builder(builder) {}
+    pir::Function* compileFunction(SEXP);
+    pir::Function* compileFunction(Function*, std::vector<SEXP>);
+    pir::Value* translateCode(rir::Function*, rir::Code*);
+    void optimizeFunction(pir::Function*);
+    pir::Module* compileModule(SEXP f);
```
o-, line 41: not implemented, also not necessary. the module is just a collection of all compiled functions

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -0,0 +1,58 @@
...
+  public:
+    Rir2Pir()
+        : PirTranslator(), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Module* module)
+        : PirTranslator(module), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Builder* builder)
+        : PirTranslator(), mergepoint(), state(), builder(builder) {}
+    pir::Function* compileFunction(SEXP);
+    pir::Function* compileFunction(Function*, std::vector<SEXP>);
+    pir::Value* translateCode(rir::Function*, rir::Code*);
```
o-, line 39: shouldn't this be private?

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -0,0 +1,58 @@
...
+        m(pc);
+        return true;
+    }
+};
+
+class Rir2Pir : public PirTranslator {
+  public:
+    Rir2Pir()
+        : PirTranslator(), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Module* module)
```
o-, line 33: unused

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -0,0 +1,58 @@
...
+        : PirTranslator(module), mergepoint(), state(), builder(nullptr) {}
+    Rir2Pir(pir::Builder* builder)
```
o-, line 35: shouldn't this be private/protected. that's not public api, but just used internally for inner functions

rir/src/compiler/pir/module.h
o-: shouldn't the over structure be more like a map:

    closure (SEXP)  ----->  compiledCode

and then `compiledCode` is one struct per closure that contains all the code:

    struct CompileCode {
        rir::Function* src;
        pir::Function* current;
        std::vector<pir::Function*> steps;
    }

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+pir::Function* Rir2Pir::compileInnerFunction(Function* function2Compile,
+                                        std::vector<SEXP> fmls) {
+    pir::Function* pirFunction = new pir::Function(fmls);
+    IRTransformation* rir2Pir =
+        new IRTransformation(function2Compile, pirFunction);
+    this->builder = new Builder(pirFunction, Env::theContext());
+
+    this->module->functions.push_back(rir2Pir);
+    this->translateCode(function2Compile, rir2Pir->srcCode);
+    if (Verify::apply(pirFunction))
```
o-, line 64: please at least assert

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -0,0 +1,58 @@
...
+    void optimizeFunction(pir::Function*);
+    pir::Module* compileModule(SEXP f);
+    
+    /*static pir::IRTransformation* declare(SEXP&);
+    static pir::IRTransformation* declare(rir::Function*);
+    static pir::IRTransformation* declare(rir::Function*, rir::Code*);*/
+  private:
+    std::unordered_map<Opcode*, pir::StackMachine> mergepoint;
+    pir::StackMachine state;
+    pir::Builder* builder;
```
o-, line 49: I think the builder does not belong here

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+        return pirFunction;
+    return nullptr;
+}
+
+pir::Value* Rir2Pir::translateCode(rir::Function* srcFunction, rir::Code* srcCode) {
+    this->recoverCFG(srcCode);
+    
+    std::deque<StackMachine> worklist;
+    pir::Builder* builder = this->getBuilder();
```
o-, line 73: this is very confusing.

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+    this->optimizeFunction(pirFunction);
+    return pirFunction;
+}
+
+pir::Function* Rir2Pir::compileInnerFunction(Function* function2Compile,
+                                        std::vector<SEXP> fmls) {
+    pir::Function* pirFunction = new pir::Function(fmls);
+    IRTransformation* rir2Pir =
+        new IRTransformation(function2Compile, pirFunction);
+    this->builder = new Builder(pirFunction, Env::theContext());
```
o-, line 60: 1. overriding the builder here and not restoring it is very confusing (and probably wrong), since you recursively call this function
2. I would really not keep the builder in the class, since there are multiple ones.
3. The builder is never freed.

you should easily get what you want by just:
```
Builder builder(pirFunction, Env::theContext()); // allocates builder on the stack
translateCode(function2Compile, rir2Pir->srcCode, builder); // passes the builder to translate code
````
where translateCode would have a `Builder& builder` builder reference as the last argument

File: rir/src/compiler/translations/rir_2_pir.h [hunk 17]
```diff
@@ -0,0 +1,58 @@
...
+        }
+        m(pc);
+        return true;
+    }
+};
+
+class Rir2Pir : public PirTranslator {
+  public:
+    Rir2Pir()
+        : PirTranslator(), mergepoint(), state(), builder(nullptr) {}
```
o-, line 32: `PirTranslator(), mergepoint(), state(),` not needed

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+            state.push((*builder)(new MkFunCls(innerF, builder->env)));
+
+            matched = true;
+            state.setPC(next);
+        });
+
+        if (!matched) {
+            int size = state.stack_size();
+            state.runCurrentBC(builder, srcFunction, &results);
```
o-, line 205: passing pointers to internal state is better avoided in C++ (if the state outlives the compiler this will create a dangling pointer). if needed, pass references.

rir/src/compiler/pir/stack_machine.h
```diff
@@ -0,0 +1,48 @@
...
+      bool doMerge(Opcode*, Builder*, StackMachine*);
+
+      void runCurrentBC(Builder*, rir::Function*, std::vector<ReturnSite>*);
```
o-, line 25: I am not really convinced this shold be here and not in the Rir2Pir. The decision to move it here seems to cause a lot of design pain points. For example passing the local variable `results` as a pointer smells.

rir/src/compiler/translations/rir_2_pir.cpp
```diff
@@ -0,0 +1,351 @@
...
+                state.setPC(trg);
+                continue;
+            }
+
+            // Conditional jump
+            switch (bc.bc) {
+            case Opcode::brtrue_:
+            case Opcode::brfalse_: {
+                Value* v = state.pop();
+                (*builder)(new Branch(v));
```
o-, line 114: this is very confusing (`()` operator on a pointer).

another reason to make builder a reference, not a pointer :).

also the idea with the `operator()` is that you make the object name a verb. Like `build(...)`.

I am also fine with not using the `()` operator for this case.

o- (commented):
Hey, thanks for this. I think overall it got cleaner.

There are many C++ gotchas in there, that's why I added a lot of comments.

Mostly I don't agree with the split of work between StackMachine and Rir2Pir. Before this, StackMachine was just a dumb data container. Now it is somehow in-between data container and partial functionality. I find it confusing.

Then there is some resource allocation and bookkeeping problem with the builder.

I still havn't debugged the test failure, I'll look into that next.

rir/src/compiler/pir/stack_machine.cpp
```diff
@@ -0,0 +1,397 @@
...
+        case Opcode::beginloop_:
+        case Opcode::endcontext_:
+        case Opcode::ldddvar_:
+        case Opcode::int3_:
+            std::cerr << "Cannot compile Function. Unsupported bc\n";
+            bc.print();
+            assert(false);
+            break;
+    }
+    this->consumeBC(src->body()->endCode());
```
o-, line 340: ok, found your bug! when you are compiling promises, then this compares the pc from the promise to the end of the function. (ie. in Rir2Pir srcCode vs. srcFunction)

rir/src/compiler/translations/pir_translator.h
```diff
@@ -0,0 +1,26 @@
...
+#include "runtime/Function.h"
+
+namespace rir {
+
+class PirTranslator {
+  public:
+    pir::Module* module;
+    virtual pir::Function* compileFunction(SEXP) = 0;
+    //virtual pir::Function* declare(const std::vector<SEXP>& a) = 0;
+    PirTranslator() : module(new pir::Module) , verbose(false) {}
```
o-, line 15: So this get's called for example on StackMachine:134, which will cause a module to be allocated and then never freed.

If you call the compiler once and then there are multiple modules created, then that's probably not good

PR #96, opened 2018-03-28
Do not assume we know about absent values for tainted envs

o-:
Our environment analysis did not correctly handle tainted envs.

For example in this case, with two environments:

    [a -> 1](tainted)  ->   [b -> 1]

when we load 'a' we know that it comes from the first env. But if
we load 'b', the first environment is tainted, so we do not
know if 'b' is really absent. therefore it could come from the
first, or second env.


o-:
simple bugfix, tbr

PR #97, opened 2018-03-28
 Closure conversion, sort of ;) 

o-:
When we know that a load will hit the outer environment (called
theContext, or ? in pir), we do not want to keep the local
environment alive. For example in `function() a`, we know that
`a` is gonna be loaded from the closure environment.
    
So this optimization transforms:

        env     e0.0  = MkEnv            (parent=?)
        val^?   %0.1  = LdVar            (a, e0.0)

to

        val^?   %0.0  = LdVar            (a, ?)

### DIFF
--- a/rir/src/compiler/analysis/abstract_value.cpp
+++ b/rir/src/compiler/analysis/abstract_value.cpp
@@ -49,5 +49,24 @@ void AbstractPirValue::print(std::ostream& out) {
     }
     out << ") : " << type;
 }
+
+AbstractLoad AbstractREnvironmentHierarchy::get(Value* env, SEXP e) const {
+    while (env != UnknownParent) {
+        if (this->count(env) == 0)
+            return AbstractLoad(env, AbstractPirValue::tainted());
+        auto aenv = this->at(env);
+        const AbstractPirValue& res = aenv.get(e);
+        if (!res.isUnknown())
+            return AbstractLoad(env, res);
+        // Tainted environment, we can't bet on absent values being
+        // actually absent.
+        if (aenv.tainted)
+            return AbstractLoad(env, AbstractPirValue::tainted());
+        env = at(env).parentEnv;
+        if (env == UnknownParent && Env::parentEnv(env))
+            env = Env::parentEnv(env);
+    }
+    return AbstractLoad(env, AbstractPirValue::tainted());
+    }
 }
 }
--- a/rir/src/compiler/analysis/abstract_value.h
+++ b/rir/src/compiler/analysis/abstract_value.h
@@ -98,7 +98,8 @@ struct AbstractPirValue {
 
 static Value* UnknownParent = (Value*)-1;
 static Value* UninitializedParent = nullptr;
-static Function* UnknownFunction = (Function*)-1;
+class MkFunCls;
+static MkFunCls* UnknownFunction = (MkFunCls*)-1;
 
 /*
  * An AbstractREnvironment is a static approximation of an R runtime Envrionment
@@ -115,7 +116,7 @@ static Function* UnknownFunction = (Function*)-1;
  */
 struct AbstractREnvironment {
     std::unordered_map<SEXP, AbstractPirValue> entries;
-    std::unordered_map<Value*, Function*> functionPointers;
+    std::unordered_map<Value*, MkFunCls*> mkClosures;
 
     Value* parentEnv = UninitializedParent;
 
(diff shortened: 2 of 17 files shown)


rir/src/compiler/analysis/scope.cpp
```diff
@@ -72,8 +76,13 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
     if (mk) {
         Value* parentEnv = mk->env();
         // If we know the caller, we can fill in the parent env
-        if (parentEnv == Env::theContext() && invocation)
-            parentEnv = invocation->env();
```
o-, line 76: this is actually a bugfix. In

`mkEnv(parent=parent?)` we currently plug in the env of the caller as the parent, instead of the env of the declaration site :(

o-:
ok, I actually extended this a bit, since I realized that there where some issues with the existing approach. Now I keep track of the existing R environments that belong to the closures, which helps us to keep things apart.

o-:
So to extend a bit on the second commit here, the idea is that when you compile a closure, we keep track of the the closure environment (the outer environment, where the closure was declared). So for example `function() a <<- 1` will now compile to:
```
  real$   %0.0  = LdConst          [1] 1                                             
  void            StVar            (a, %0.0, <environment: R_GlobalEnv>)             
  void            Return           (%0.0)             
```

given this closure was declared at the global environment `<environment: R_GlobalEnv>`.

o-:
rebase on top of #100

rir/src/compiler/translations/pir_translator.cpp
```diff
@@ -0,0 +1,9 @@
+#include "pir_translator.h"
+
+namespace rir {
+namespace pir {
+
+bool PirTranslator::isVerbose() { return verbose; }
+void PirTranslator::setVerbose(bool isVerbose) { verbose = isVerbose; }
```
JanJecmen: Make `verbose` public? This looks like Java :) (just my personal opinion)

rir/src/compiler/analysis/abstract_value.cpp
```diff
@@ -49,5 +49,24 @@ void AbstractPirValue::print(std::ostream& out) {
...
+            return AbstractLoad(env, AbstractPirValue::tainted());
+        auto aenv = this->at(env);
+        const AbstractPirValue& res = aenv.get(e);
+        if (!res.isUnknown())
+            return AbstractLoad(env, res);
+        // Tainted environment, we can't bet on absent values being
+        // actually absent.
+        if (aenv.tainted)
+            return AbstractLoad(env, AbstractPirValue::tainted());
+        env = (*this).at(env).parentEnv;
```
JanJecmen: `this->at(...)`? Or even just `aenv`?

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -175,15 +176,15 @@ struct AbstractREnvironment {
...
-        for (auto e : other.functionPointers)
+        for (auto e : other.mkClosures)
             fps.insert(std::get<0>(e));
         for (auto n : fps) {
-            if (functionPointers[n] != UnknownFunction &&
-                (other.functionPointers.count(n) == 0 ||
-                 functionPointers[n] != other.functionPointers.at(n))) {
-                functionPointers[n] = UnknownFunction;
+            if (mkClosures[n] != UnknownFunction &&
+                (other.mkClosures.count(n) == 0 ||
```
JanJecmen, line 185: What does this mean? Here we merge two aenvs, and one of them has a function that is known, and the other doesn't have it. Why is the result unknown and not the one we know?

JanJecmen (approved):

rir/src/compiler/analysis/abstract_value.h
o-: UnknownFunction is the top. The lattice is

    Top: UnknownFunction
                 |
      MkClosureFun*
                 |
     Bottom:  Absent in the hashmap

o-: oh, I see what you mean, yeah from that lattice it should be possible to set to the other

o-: But the lattice is a bit different:

           UnknownFunction
             /            \
       MkClosureFun*       Absent
             \            /
               Bottom

But the implementation does not distinguish between bottom and absent. That's why we have to be conservative here.

An example would be:

    if (x)
      (function()1)();
    bla...

at `bla...` the closure created in the then branch does not necessarily exist and we should set it to Unknown.

rir/src/compiler/translations/pir_translator.cpp
o-: I agree, but that's from #100 , so I am gonna leave it here. Feel free to change in another PR

PR #99, opened 2018-03-28
Rename fun cls

o-:
I would like to rename pir::Function to pir::Closure in #97 , but then I realized that this will probably mess with @charig and @JanJecmen 's ongoing work. So I keep the renaming here in a separate PR, such that we can wait with this until you are ready.


charig:
A first pass of my work has finished. We should decide (from what I perceive between @JanJecmen @mhyee and @o-) if we would like to go on that direction or not. If yes, we should integrate the code and then I can start a second pass.

o-:
rebased

PR #100, opened 2018-03-30
refactoring

o-:
fixed version of #94


charig:
Thanks very much for this refactoring of my refactoting @o-. I like it. If we finally integrate it to the main repo I can try another pass so that the "general API" not only work from rir->pir but also from pir>pir (optimizations) and from pir->rir.

charig (approved):
I am only not sure of versioned functions. In my big picture each optimization should be a new compilation pass and the module should have an ordered list of "translations" (rir->pir;pir->opt1;opt1->opt2;...;optn->rir).
But, IMO,  this is alerady great step towards a cleaner API.

o-:
I am still not sure what the gain of such an api would be? I mean I can see that for debugging it would be nice to save intermediate steps. But we certainly can't afford to copy the code for every step (when not debugging). Also, as you have seen, we use pointers everywhere in the ir, so every time you copy, you also need to rewrite all pointers. It is fairly expensive.

PR #101, opened 2018-04-02
add support for statically resolved calls

o-:
When the call target is statically known, let's use a different call
instruction that allows us to directly link to that known closure.

This commit also fixes a bug in the compiler, where we assumed that
every 'static_call_stack' calls a builting (which is wrong).

Also we add support for speculative static calls for callsites which
have been observed to be monomorphic.

### DIFF
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -15,14 +15,15 @@ class TheScopeAnalysis : public StaticAnalysis<AbstractREnvironmentHierarchy> {
 
     static constexpr size_t maxDepth = 5;
     size_t depth;
-    Call* invocation = nullptr;
+    Value* staticClosureEnv = nullptr;
 
     TheScopeAnalysis(Closure* origin, const std::vector<SEXP>& args, BB* bb)
         : Super(bb), origin(origin), args(args), depth(0) {}
-    TheScopeAnalysis(Closure* origin, const std::vector<SEXP>& args, BB* bb,
-                     const AS& initialState, Call* invocation, size_t depth)
+    TheScopeAnalysis(Closure* origin, const std::vector<SEXP>& args,
+                     Value* staticClosureEnv, BB* bb, const AS& initialState,
+                     size_t depth)
         : Super(bb, initialState), origin(origin), args(args), depth(depth),
-          invocation(invocation) {}
+          staticClosureEnv(staticClosureEnv) {}
 
     void apply(AS& envs, Instruction* i) const override;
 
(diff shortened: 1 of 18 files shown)


o-:
Hey, would appreciate a quick peek here.
^^ @charig @JanJecmen  ?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -69,19 +70,15 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
...
         // If we know the caller, we can fill in the parent env
-        if (parentEnv == Env::notClosed() && invocation) {
-            Value* cls = invocation->cls();
-            if (envs[invocation->env()].mkClosures.count(cls)) {
-                auto mkCls = envs[invocation->env()].mkClosures.at(cls);
-                parentEnv = mkCls->env();
-            }
+        if (parentEnv == Env::notClosed() &&
+            staticClosureEnv != Env::notClosed()) {
+            parentEnv = staticClosureEnv;
```
charig, line 81: Is it possible to have a parentEnv and a staticEnv?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -96,18 +93,33 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
...
+        } else if (StaticCall::Cast(i)) {
+            auto call = StaticCall::Cast(i);
+            Closure* trg = call->cls();
+            if (trg->argNames.size() == call->nCallArgs()) {
+                TheScopeAnalysis nextFun(trg, trg->argNames, trg->closureEnv(),
+                                         trg->entry, envs, depth + 1);
                 nextFun();
                 envs.merge(nextFun.result());
                 handled = true;
             }
```
charig, line 120: Not sure but perhaps move the if (trg->argnames).... that is repeated in both branches outside?

rir/src/compiler/opt/inline.cpp
```diff
@@ -24,22 +24,38 @@ class TheInliner {
...
+                if (Call::Cast(*it)) {
+                    Call* call = Call::Cast(*it);
+                    auto cls = MkFunCls::Cast(call->cls());
+                    if (!cls)
+                        continue;
+                    inlinee = cls->fun;
+                    if (inlinee->argNames.size() != call->nCallArgs())
+                        continue;
+                    assert(cls->fun->closureEnv() == Env::notClosed());
+                    staticEnv = cls->env();
```
charig, line 43: Is it also static in this case? If not I would just call it env.

charig (approved):
I just could not fully get the ifPromise(&lambda) pattern but the PR seems fine. Anyway, I still do not fully understand the system so I did my best but I may have missed some things.

o-:
regarding the `ifPromise`:
the core idea is that, it used to be, that the `promise` member in `MkArg` is never null. Now I wanted it to be null sometimes. To avoid npe, I made the member private and put it behind this "monadic" api. The idea is that you pass a lambda, and when the promise is not null, the lambda will be executed. That is a very nice pattern, that is much more robust, than to expect all the clients doing a null check.

o-:
That being said however, maybe it was a bad idea to allow the promise to be null in the first place. I will try to see if it is possible to not abuse `MkArg` here, and either have `StaticCall` accept non-promised arguments, or add a second instruction to create "fake promises".

o-:
I have recently started to have this policy of:
* public variables are never null
* public functions never return null

and it has served me very well. I can highly recommend as a coding style.

rir/src/compiler/analysis/scope.cpp
o-: yes, in which case the parent env has precedence.

o-:
superseeded by #109

PR #102, opened 2018-04-03
add a test

o-:


charig (approved):

o-:
well, it breaks on mac and we shoud find out why

charig:
Yes, I saw when approving. I will have a look tomorrow.

JanJecmen:
So on my machine, when I run it in debugger, it loops in the scope analysis forever. when I pause randomly, often I land in `__builtin_operator_new` or `_delete` in some std hash table (set / map), or somewhere else in there like hash computation or rehashing. When it runs, it just spins the cpu, memory is not leaking.

Since it is mac only, is it possible that it is caused by using libc++ instead of the gnu libstdc++?

charig:
I went one hour through it. The main problem is in the scope analysis as Jan said. But debugging it seems more like some domain subtle problem. Essentially the *done* var inside the 
```
do {
...
} while (!done)
``` 
ends up always in false. And after each pass of the visitor basic blocks 2 and 3 are always changed.

So I will continue debugging tomorrow but I wrote down here in case @o- have some idea why running that block in that particular tests may throw different results in Mac and Unix...

o-:
tbr, bugfix

PR #105, opened 2018-04-06
Make optimizations part of the compiler 

charig:


rir/src/compiler/translations/pir_translator.cpp
```diff
@@ -1,9 +1,11 @@
 #include "pir_translator.h"
+#include "rir_compiler.h"
 
 namespace rir {
 namespace pir {
 
-bool PirTranslator::isVerbose() { return verbose; }
-void PirTranslator::setVerbose(bool isVerbose) { verbose = isVerbose; }
+void PirTranslator::apply(Closure* function) { 
+    this->applyTranslation(function);
```
o-: I don't understand the purpose of this delegation. Why not call all of them apply, or applyTranslation

(no `this` required)

rir/src/compiler/translations/pir_translator.h
```diff
@@ -2,22 +2,24 @@
...
+class RirCompiler;
 class PirTranslator {
   public:
-    PirTranslator(bool verbose) : verbose(verbose) {}
+    PirTranslator(std::string name)
+        : name(name) {}
 
-    pir::Closure* compileClosure(SEXP);
+    void apply(Closure* function);
+    std::string& getName() { return this->name; }
```
o-: no `this`

rir/src/compiler/opt/cleanup.h
```diff
@@ -1,13 +1,19 @@
...
 class Closure;
-class Cleanup {
+class Cleanup : public PirTranslator {
   public:
-    static void apply(Closure* function);
+    Cleanup()
+        : PirTranslator("cleanup") {};
+
+  protected:
+    void applyTranslation(Closure* function);
```
o-: please always mark implementations of `virtual` functions with `override` or `final`

rir/src/compiler/opt/delay_env.h
```diff
@@ -1,19 +1,25 @@
...
+
 class Closure;
-class DelayEnv {
+class DelayEnv : public PirTranslator {
   public:
-    static void apply(Closure*);
+    DelayEnv() : PirTranslator("Delay Environment"){};
+
+  protected:
+    void applyTranslation(Closure* function);
```
o-: missing `override`, there are more instances below

rir/src/compiler/opt/elide_env.h
```diff
@@ -1,6 +1,8 @@
 #ifndef PIR_ELIDE_ENV_H
 #define PIR_ELIDE_ENV_H
 
+#include "../translations/rir_compiler.h"
```
o-: that's the wrong include. `pir_translator.h` should be enough, not? (more instances of this below and above)

rir/src/compiler/translations/pir_translator.h
o-: also returning a string reference in a getter is very bad style.

the idea with a reference is that you can modify in place. Ie. a client can change the string with:
`translator->getName() = "another name"` which completely defeats the idea of making `name` private in the first place :)

So either return `const std::string&` which disallows modification of the returned reference. Or return `std::string` (which semantically copies the value) and rely on the C++ compiler optimizations to avoid unnecessary copying.

In this case I personally would opt for the second option.

as a rule of thumb, returning (or passing) non-const references is almost always unwarranted.

rir/src/compiler/translations/rir_compiler.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace rir {
+namespace pir {
+
+class RirCompiler {
+  public:
+    RirCompiler(Module* module) : module(module) {}
+    virtual Closure* compileClosure(SEXP) = 0;
+    
+    Module* getModule() { return module; }
```
o-: why? I removed this on purpose. is there a legitimate use for this method? The module is owned by the client of the compiler. So the client should hold on to it. If a client gets the module from the compiler, this is most likely a sign of a bad client.

rir/src/compiler/translations/rir_2_pir_compiler.h
```diff
@@ -0,0 +1,47 @@
...
+    Rir2PirCompiler(Module* module) : RirCompiler(module) {
+        translations.push_back(new ForceDominance());
+        translations.push_back(new ScopeResolution());
+        translations.push_back(new Cleanup());
+        translations.push_back(new DelayInstr());
+        translations.push_back(new ElideEnv());
+        translations.push_back(new DelayEnv());
+    }
+
+    Closure* compileClosure(SEXP);
```
o-: mark override or final

rir/src/compiler/translations/rir_2_pir_compiler.h
```diff
@@ -0,0 +1,47 @@
+#ifndef RIR_2_PIR_COMPILER_H
+#define RIR_2_PIR_COMPILER_H
+
+#include "../opt/cleanup.h"
+#include "../opt/delay_env.h"
+#include "../opt/delay_instr.h"
+#include "../opt/elide_env.h"
+#include "../opt/force_dominance.h"
+#include "../opt/inline.h"
+#include "../opt/scope_resolution.h"
```
o-: I don't like those dependencies here. Can you move the constructor implementation to the cpp file?

rir/src/compiler/translations/rir_2_pir.h
```diff
@@ -1,37 +1,19 @@
 #ifndef RIR_2_PIR_H
```
o-, line 1: maybe we should move the rir_2_pir.[h|cpp] as well as the pir/stackmachine.[h|cpp] to a `rir_2_pir_compiler` subfolder?

rir/src/compiler/translations/rir_2_pir_compiler.h
```diff
@@ -0,0 +1,47 @@
...
+#include "rir_compiler.h"
+
+namespace rir {
+namespace pir {
+
+class Rir2PirCompiler : public RirCompiler {
+  public:
+    Rir2PirCompiler(Module* module) : RirCompiler(module) {
+        translations.push_back(new ForceDominance());
```
o-: personally I would have used lambdas here. Let me show you an alternative implementation:

```
class Rir2PirCompiler {
  typedef std::function<Closure*(Closure*)> PirTranslatorFunction;
  std::vector<PirTranslatorFunction> translations;
  ...
    // register a translation:
    translations.push_back([](Closure* c){return ForceDominance::apply(c);});
  ...
    // apply translation:
    Closure* cur = ...;
    for (translate : translations) {
        cur = translate(cur);
    }
```

This approach give you much less coupling! You can get away without all the translators implementing the exact same interface. You don't have dependencies in the header files. Your don't need virtual methods. And you don't need to instantiate the translators, like here, but can instead use static methods, like before.

o- (commented):
Very nice refactoring. Thanks.

Read the comment on rir_2_pir_compiler.h:19 first. If you decide to change the api in the suggested way, other comments might become obsolete. But feel free to pick the design you like more.

Other than the comments, this lgtm

o-:
@charig another idea, on how to completely statically combine translations. (not sure if we actually want this, it's neat but also ugly):

```
#include <iostream>

struct Closure {}; 

template<class... Translations>
struct TranslationChain {
    static void apply(Closure* c) {
        doApply<Translations...>(c);
    }   

  private:
    template<class T1, class T2, class... Tn> 
    static void doApply(Closure* c) {
        T1::apply(c);
        T2::apply(c);
        doApply<Tn...>(c);
    }   
    template<class T>
    static void doApply(Closure* c) {
        T::apply(c);
    }   
};

struct TranslationA { static void apply(Closure* c) { std::cout << "Apply A\n"; } };
struct TranslationB { static void apply(Closure* c) { std::cout << "Apply B\n"; } };
struct TranslationC { static void apply(Closure* c) { std::cout << "Apply C\n"; } };

int main() {
    TranslationChain<TranslationA, TranslationB, TranslationC> translate;
    Closure c;
    translate.apply(&c);
}
```

rir/src/compiler/translations/pir_translator.cpp
charig: Good catch. Completely nonsense what I did.

rir/src/compiler/translations/pir_translator.h
charig: I Went with the second option. Thanks

charig:
@o-: I added a new version with most of your suggestions. I just did not change with the important stuff you mentioned. We already discuss it by chat. The only thing I can say is that I see the OO model may be a little more flexible at this stage in which we may still perform changes and extend the model. And perhaps we can later move to something like what you suggest when the compiler becomes more mature and we already understood all the use cases, advantages, and drwabacks. But I have no strong reasons for pushinng my alternative instead of a static functional lambda application for this particular case and I am certainly not sure which is the best. So if you want to go for your alternative, lets do it, it is ok for me.

o-:
@charig sure, as I said, this looks good to me, go ahead

PR #108, opened 2018-04-13
PIR to RIR translation

JanJecmen:
This is WIP. The ideas for the first pass are:
- just get something working
- register allocation is the dumbest thing possible (phi and inputs are in the same slot, other that that, everything gets its own slot)
- use `ldloc_` and `stloc_` for every instruction arguments and result (there are some optimizations to this like load/store is copied directly, store/load is removed if the store is not needed for later)
- a lot is still missing, most notably function calls

Also, a lot of it is clang format...


rir/src/compiler/pir/bb.h
```diff
@@ -67,6 +67,7 @@ class BB {
     void append(Instruction* i);
 
     Instrs::iterator insert(Instrs::iterator it, Instruction* i);
+    Instrs::iterator insert(Instrs::iterator it, Instrs is);
     void replace(Instrs::iterator it, Instruction* i);
```
charig: This last two are a little weird. What is Instrs? Just a list of instructions?

JanJecmen: Yes, it's just a `std::vector<Instruction*>`, but this is a leftover piece of code that is not used currently.. I removed it for now.

o-:
is this up for review, or still wip?

JanJecmen:
~both, kind of :)~
wip

o-:
UPDATE: I might have changed my mind on this. I can see how it makes some things easier...

hey, so first question: is `PirCopy` necessary?

it seems to me that you could do the following instead:
1. create a list of all `Phi`
2. whenever you are done compiling a BB `bb`, you do the following:

```
for (phi : listOfPhis) {
  if (phi->input.find(bb))
    // do whatever Tag::PirCopy: case does in the pir2rir compiler
```

But, probably I am missing something?

rir/src/compiler/pir/instruction.h
```diff
@@ -773,10 +784,21 @@ typedef SEXP (*CCODE)(SEXP, SEXP, SEXP, SEXP);
 
 class VLI(CallBuiltin, Effect::Any, EnvAccess::Write) {
   public:
+    constexpr static size_t callArgOffset = 1;
```
o-: I don't think that's necessary here. Every argument to the builtin instruction are arguments for the builtin.

Call has a callArgOffset, because the first instruction argument is the closure. So the first call argument starts at index 1. but here the builtin is an immediate argument (part of the instruction) and not an ir-level input to the instruction. (like in callEager)

rir/src/compiler/pir/instruction_list.h
```diff
@@ -61,6 +61,7 @@
     V(ForSeqSize)                                                              \
     V(Deopt)                                                                   \
     V(Force)                                                                   \
-    V(CastType)
+    V(CastType)                                                                \
+    V(PirCopy)
```
o-, line 65: see comment bellow, it would be great if we can do it without this instruction. The reason is, I am afraid somebody will use it earlier in the compiler and suddenly we have aliasing to deal with.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+            case Tag::Return: {
+                Return* ret = Return::Cast(instr);
+                load(it, ret->arg(0).val());
+                cs << BC::ret();
+
+                // this is the end of this BB
+                return;
+            }
+            case Tag::MkArg: {
+                break;
```
o-: maybe `assert(uses are all call instruction)`

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+                auto call = Call::Cast(instr);
+                auto cls = call->cls();
+                loadEnv(it, call->env());
+                cs << BC::setEnv()
+                   << BC::ldloc(alloc[cls]); // TODO: can use load(it, cls) for
+                                             // the closure?
+
+                std::vector<FunIdxT> callArgs;
+                call->eachCallArg([&](Value* arg) {
+                    auto mkarg = MkArg::Cast(arg);
```
o-: can we add a todo that we are ignoring `mkarg->arg<0>()` (the potentially eagerly computed value slot). It's fine to ignore it for now, but we should probably add it at some point later.

rir/src/interpreter/interp.cpp
```diff
@@ -119,6 +119,9 @@ static void jit(SEXP cls, Context* ctx) {
 
 void tryOptimizeClosure(DispatchTable* table, size_t offset, Context* ctx) {
 
+    // TODO: needs to change for PIR
```
o-: hmm, not super happy about dead code. I would suggest we remove this rir2rir optimization stuff and focus on rir2pir2rir optimization.

rir/src/ir/insns.h
```diff
@@ -80,6 +85,11 @@ DEF_INSTR(stvar2_, 1, 1, 0, 1)
  */
 DEF_INSTR(stloc_, 1, 1, 0, 1)
 
+/**
+ * copyloc_:: copy one local into another
+ */
+DEF_INSTR(copyloc_, 2, 0, 0, 1)
```
o-: can we give this another name? `movloc` maybe. I find `copying` a bit a heavy word for the operation.

rir/src/optimization/localize.h
```diff
@@ -36,6 +36,8 @@ class Localizer : public InstructionDispatcher::Receiver {
         SEXP sym = (*ins).immediateConst();
         auto v = analysis[ins][sym];
 
+// Disable as part of PIR work... ldarg_ reused for pir, since this will go away
+#if 0
```
o-: again, I think we should just merge #82 or something similar, instead  of keeping dead code

rir/src/runtime/Function.h
```diff
@@ -61,6 +61,7 @@ struct Function {
         envChanged = false;
         deopt = false;
         markOpt = false;
+        isPirCompiled = false;
```
o-: is this neccessary? isn't the slot/signature mechanism enough?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+                break;
+            }
+            case Tag::LdVarSuper: {
+                assert(false && "not yet implemented.");
+                break;
+            }
+            case Tag::StVar: {
+                auto stvar = StVar::Cast(instr);
+                loadEnv(it, stvar->env());
+                cs << BC::setEnv() << BC::stvar(stvar->varName);
```
o-: can we at least keep track of the last setEnv and avoid if unnecessary?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+            }
+            case Tag::Force: {
+                auto force = Force::Cast(instr);
+                load(it, force->arg(0).val());
+                cs << BC::force();
+                store(it, force);
+                break;
+            }
+            case Tag::CastType: {
+                // TODO: what should it do?
```
o-: nop should be fine

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+                assert(false && "not yet implemented.");
+                break;
+            }
+            case Tag::StaticEagerCall: {
+                assert(false && "not yet implemented.");
+                break;
+            }
+            case Tag::CallBuiltin: {
+                auto blt = CallBuiltin::Cast(instr);
+                // TODO: is it needed to set the environment?
```
o-: i would think so

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+            }
+            case Tag::CallSafeBuiltin: {
+                assert(false && "not yet implemented.");
+                break;
+            }
+            case Tag::MkEnv: {
+                auto mkenv = MkEnv::Cast(instr);
+                loadEnv(it, mkenv->parent());
+                cs << BC::makeEnv() << BC::dup() << BC::setEnv();
+                // bind all args
```
o-: maybe we should have an instruction that does this in one go? not for now, but in the future...

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+                mkenv->eachLocalVar([&](SEXP name, Value* val) {
+                    load(it, val);
+                    cs << BC::stvar(name);
+                });
+                store(it, mkenv);
+                break;
+            }
+            case Tag::Phi: {
+                auto phi = Phi::Cast(instr);
+                load(it, phi->arg(0).val());
```
o-: I don't understand this. Why can you take the input 0? because you know that all of the inputs are in the same local value?

o- (commented):
yeah, finally we can run stuff. did a first pass. thanks for the PR.

o-:
small note: can you squash the commits? Eg.:

```
git checkout master
git checkout -b p2r-squash
git merge --squash p2r
git add rir && git commit
```

rir/src/compiler/util/visitor.h
```diff
@@ -199,7 +199,7 @@ class Visitor
 class BreadthFirstVisitor
     : public VisitorImplementation<true, VisitorHelpers::IDMarker> {};
 class UnstableIDsVisitor
-    : public VisitorImplementation<false, VisitorHelpers::PointerMarker> {};
+    : public VisitorImplementation<true, VisitorHelpers::PointerMarker> {};
```
mhyee, line 202: Why is this changed to stable ordering?

rir/src/api.cpp
```diff
@@ -114,17 +116,58 @@ REXPORT SEXP rir_analysis_liveness(SEXP what) {
...
+
+    // compile back to rir
+    auto table = DispatchTable::unpack(BODY(what));
+    size_t offset = 0;
+    auto oldFun = table->at(offset);
+    pir::Pir2RirCompiler p2r;
+    auto fun = p2r(m, oldFun);
+    p(fun->container());
+
+    // TODO: put instead into a new table slot...
```
mhyee: I think we need to figure out exactly what we want to do here. I was under the impression that in this case, the new version should be in the same table slot (because the calling convention or interface is the same), but in the linked list of versions. In other words, we use a different slot if the caller needs to do something different.

But part of this reasoning was because we had the old optimization/deoptimization system, which has been removed (#82). So maybe we should rethink this.

rir/src/api.cpp
```diff
@@ -114,17 +116,58 @@ REXPORT SEXP rir_analysis_liveness(SEXP what) {
...
+
+    // patch the closure
```
mhyee: I think this is just asking for a method that inserts a new version into an existing slot and updates all the pointers/fields.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+    Alloc(Code* code) {
+        BreadthFirstVisitor::run(code->entry, [&](Instruction* instr) {
+            Phi* phi = Phi::Cast(instr);
+            if (phi) {
+                auto slot = allocateLocal(phi);
+                phi->eachArg([&](Value* arg) {
+                    // here the arg must not yet be allocated
+                    if (shouldAllocate(arg))
+                        allocateLocal(arg, slot);
```
mhyee, line 41: I don't understand, why is the same slot being used for all the phi's arguments?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+    void pushPromise(SEXP ast) {
+        defaultArg.push(false);
+        push(ast);
+    }
+    void pushBody(SEXP ast) {
+        defaultArg.push(false);
+        push(ast);
+    }
+
+    FunIdxT pop(size_t localsCnt) {
```
mhyee: Hmm, with this signature, it almost looks as if the function is popping `localsCnt` times, rather than popping once and passing `localsCnt` to `finalize`. (This actually confused me for a few seconds.) Is there a better name for this that isn't too awkward? Maybe we shouldn't call the functions push/pop?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -0,0 +1,601 @@
...
+            next = next->next0;
+        cs << BC::br(bbLabels[next]);
+    });
+
+    return alloc.slots();
+}
+
+void Pir2Rir::removePhis(Code* code) {
+
+    // For each Phi, insert copies
```
mhyee, line 548: Can you elaborate here? I'm not sure what's going on.

mhyee (commented):
I still need to look at `Pir2Rir::compile` in more detail, but for now, here's some questions and comments.

rir/src/compiler/util/visitor.h
JanJecmen: Oh, this was when I was playing with the reg. allocation.. It's used only for renumbering the BBs after optimizations, and I wanted to have the IDs stable. Not needed now I guess.

rir/src/api.cpp
JanJecmen: I agree we should think about it. For PIR, the caller indeed needs to do something different, namely it should not allocate an environment for the callee. I am working now on supporting the slots mechanism.

rir/src/api.cpp
JanJecmen: Actually, I am not sure about these... 

- `envChanged` and `envLeaked` were used for rir optimizations that went away, right @o-? Should we still keep them?
- `invocationCount` for a new version should start from 0, I guess?
- `signature`: I think this still needs discussion, maybe new slot = different signature?

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: That is the out-of-SSA transition: you remove the `Phi` by assigning it to a register, and making sure that all inputs are in that same register. Then when executing, the `Phi` is a `noop` because you know that regardless of which path you took, the value is in the `Phi`'s register.

rir/src/compiler/util/visitor.h
o-: uhm, this should probably be stable, since, as you say it's used for renumbering... this looks like a bug, gonna look into it

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: Yeah, sorry, I was looking at that for a while and then left it like this... Renamed the method to `finalizeCode`

rir/src/api.cpp
o-: > I think we need to figure out exactly what we want to do here. I was under the impression that in this case, the new version should be in the same table slot (because the calling convention or interface is the same)

hmm, ok, I guess there are two ways to look at it. you could say, that in both cases the env is set up by the callee. just in pir optimized it is explicit and in not pir optimized it's implicit.

but I find the more natural way is to say we have two cc, one is: "env is caller setup", thats what R and unoptimized rir does. And then we have "env is callee setup", which is what optimized rir does.

rir/src/api.cpp
o-: > Should we still keep them?

no

> invocationCount for a new version should start from 0, I guess?

do we count per version or per function? if the former, then probably 0, yeah

rir/src/compiler/translations/pir_2_rir.cpp
o-: ah, yeah I was also struggling with the name here, but then forgot, after I understood what it does :)  Boissinot etal calls this "converting from SSA to CSSA (conventional ssa)"

o-: btw. (but that's for later), the copies should be parallel, to avoid introducing artificial interference.

I think we should have the following:

every BB has an "out-copy" and an "in-phi" table. Then the conversion would look like this:

```
BB0:
  a0 = 0
  b0 = 0

BB1:
  goto BB3

BB2:
  a2 = 2
  b2 = 2
  goto BB3

BB3:
  x = phi(a0/BB1, b2/BB2)
  y = phi(b0/BB1, a2/BB2)
```

in this case we would say:

```
BB1-out:
    a1o, b1o = a0, b0

BB2-out:
    a2o, b20 = a2, b2

BB3-in:
    x, y = a1o | b2o , b1o | a2o
```

o-: then, basically the removePhis pass would (instead of inserting copy instructions) populate the in-/out- tables

JanJecmen:
I addressed the comments and started the refactoring to use dispatch table slots. It is ugly with the slots now, 0 is plain RIR, 1 is PIR optimized RIR, and the offsets are just magic numbers in the code...
Gonna continue on that.

I will squash later before merge I guess? Is that OK @o-?

rir/src/api.cpp
mhyee: > I think this still needs discussion, maybe new slot = different signature?

I think this was the original plan. Each slot is described by a signature, if two versions have the same signature then they have the same slot.

o-:
sorry, are you waiting for feedback on this one? it looks good to me

JanJecmen:
Thanks, any further comments we can fix in a new pr...

PR #109, opened 2018-04-16
add support for statically resolved and eager calls

o-:
This commit adds support for eager calls

When the call target is statically known, let's use a different call
instruction that allows us to directly link to that known closure.

This commit also fixes a bug in the compiler, where we assumed that
every 'static_call_stack' calls a builting (which is wrong).

Also we add support for speculative static calls for callsites which
have been observed to be monomorphic.

@charig: this is mostly the same PR as #101 , but I added a new `StaticEagerCall`.

The previous PR did this to "simulate" eager arguments:

```
a = mkArg(noExpression, aValue)
StaticCall(cls, a)
```

(Ie. create a mkArg, with no promise expression). Leading to stupid code, since now all the mkArg might actually be "fake"-mkArgs (ie. promise expression == nullptr), and therefore all those `ifPromise` lambdas. That was just stupid.

This PR instead adds a new call, that expects a value instead of a promise:

```
StaticEagerCall(cls, aValue)
```

This blows up the code a bit, but makes it much easier to understand.

Additionally, this PR also fixes the following bug in the inliner. We used to mark the fact that we inlined a MkArg, by changing it's type. For example:

```
f:
   prom a0 = MkArg(...)
   Call(g, a0)

g:
   val^ a1 = LdArg(0)
   val a2 = force a
```

As you can see, the `LdArg` converts the promise from type `prom` (a promise as a value) to type `val^` (a promise that should be evaluated). When we inline, we just changed the type of `MkArg` from `prom` to `val^`. This breaks if the `MkArg` flows into two calls:

```
f:
   prom a0 = MkArg(...)
   Call(g, a0)
   Call(x, a0)   

g:
   val^ a1 = LdArg(0)
   val a2 = force a
```

If we inline and change the type of `a0` then the second call `Call(x, a0)` breaks. Therefore we now insert an explicit cast:

```
f:
   prom a0 = MkArg(...)
   val^ a1 = CastType(a0)
   val a2 = force(a1)          // inlined g
   Call(x, a0)
```

### DIFF
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
@@ -21,9 +21,11 @@ Rir2PirCompiler::Rir2PirCompiler(Module* module) : RirCompiler(module) {
     translations.push_back(new ForceDominance());
     translations.push_back(new ScopeResolution());
     translations.push_back(new Cleanup());
+    translations.push_back(new Cleanup());
     translations.push_back(new DelayInstr());
     translations.push_back(new ElideEnv());
     translations.push_back(new DelayEnv());
+    translations.push_back(new Cleanup());
 }
 
 Closure* Rir2PirCompiler::compileClosure(SEXP closure) {
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
@@ -46,8 +46,8 @@ Closure* Rir2PirCompiler::compileFunction(rir::Function* srcFunction,
 
 Closure* Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
                                          const std::vector<SEXP>& args,
-                                         Value* closureEnv) {
-    Closure* pirFunction = module->declare(srcFunction, args);
+                                         Env* closureEnv) {
+    Closure* pirFunction = module->declare(srcFunction, args, closureEnv);
 
     Builder builder(pirFunction, closureEnv);
 
@@ -102,7 +102,7 @@ void Rir2PirCompiler::printAfterPass(const std::string& pass,
 void Rir2PirCompiler::applyOptimizations(Closure* f,
                                          const std::string& category) {
     size_t passnr = 0;
-    for (auto translation : this->translations) {
+    for (auto& translation : this->translations) {
         translation->apply(f);
         if (isVerbose())
             printAfterPass(translation->getName(), category, f, passnr++);
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
@@ -47,23 +47,25 @@ Closure* Rir2PirCompiler::compileFunction(rir::Function* srcFunction,
 Closure* Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
                                          const std::vector<SEXP>& args,
                                          Env* closureEnv) {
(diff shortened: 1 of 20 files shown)


rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -21,9 +21,11 @@ Rir2PirCompiler::Rir2PirCompiler(Module* module) : RirCompiler(module) {
     translations.push_back(new ForceDominance());
     translations.push_back(new ScopeResolution());
     translations.push_back(new Cleanup());
+    translations.push_back(new Cleanup());
```
charig, line 24: Two consecutive cleanups are needed?

charig (approved):
I only got lost in instruction.h I still do not understand that part very well. The rest seems fine!

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
o-: yeah, cleanup is a bit stupid right now. If you have 3 dead instructions that depend on each other it removes one per run :/

o-:
thanks. @charig 
which part of instruction.h was not clear?

basically that file creates one class per instruction. The instructions are mostly uniform.

There are fix length instructions, which have a fixed number of arguments and there are varlength instructions, that have a variable number of arguments.

Every instruction inherits from InstructionDescription, which is a brief summary of what the instruction does, that is used by the various analysis/optimization passes.

The inheritance hierarchy is a bit fucked up, but it's idiomatic C++. The problem is that inheriting from a template class, gives you unrelated classes. For example:

```
template <int X>
class templateClass {};
class A : public templateClass<1> {};
class B : public templateClass<2> {};
```

here `A` and `B` are completely unrelated, because `templateClass` is invariant (neither covariant nor contravariant).

The trick is to patch it through:

```
class CommonInterface {};
template <int X>
class templateClass {} : public CommonInterface;
class A : public templateClass<1> {};
class B : public templateClass<2> {};
```

now additionally, sometimes you want a template superclass to inject something that depends on the baseclass type. For example to generate the `Cast` static methods. This is how you do it:

```
template<class BASE_CLASS>
class templateClass {
   static BASE_CLASS* Cast() { reinterpret_cast<BASE_CLASS*>(this); };
};
class SomeClass : public templateClass<SomeClass> {};
```

Now in instruction.h I combined both those patterns, and some more hacks :)

charig:
@o- The class is full of defines, then I got scared and I hadn't read that class too much :). Now it is much more clear with your explanation thanks for that. I realize I had miss this code:
```
#define FLI(type, nargs, io, env)                                              \
    type:                                                                      \
  public                                                                       \
    FixedLenInstruction<Tag::type, type, nargs, io, env>
```

So previously I saw the FLI and VLI everywhere but did not know where they came from. 

Regarding types, yes, thanks also for that. Apparently they are the same as in Java.

JanJecmen:
BTW, the last piece is callled CRTP: https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern

o-:
@JanJecmen thanks for the link.

@charig  The other thing I did here, that is a bit strange is, that I give every instruction it's tag. This allows us to safely downcast. Because we can check the tag, figure out which instruction it is and then downcast. You might say, c++ does this already, and this is true. We could use: `dynamic_cast<LdConst*>(instruction)` instead of `LdConst::Cast(instruction)`. But it is slow! see https://stackoverflow.com/questions/5134975/what-can-make-c-rtti-undesirable for an explanation (llvm does the same thing).

Currently our instruction still has virtual methods. Thus we only get the speed benefit, but we still have the space overhead for every instruction. I plan to replace this at some point, but it's probably not top prio.

charig:
@o- Thanks for the link!

PR #112, opened 2018-04-17
do not assert when trying to pir compile a non rir compiled closure

o-:


JanJecmen:
So here I am not sure what we want to do with this API:

- require something RIR compiled, change it inplace or do nothing
- just get anything, possibly compile it to RIR, then give back a new thing
- remove it completely and use eg. the `rir.markOptimize` to invoke PIR

Just a thought...

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -29,7 +30,14 @@ Rir2PirCompiler::Rir2PirCompiler(Module* module) : RirCompiler(module) {
 }
 
 Closure* Rir2PirCompiler::compileClosure(SEXP closure) {
-    assert(isValidClosureSEXP(closure));
+    assert(TYPEOF(closure) == CLOSXP);
+    if (!isValidClosureSEXP(closure)) {
+        Protect p;
```
JanJecmen: The `Protect` is not needed here, no? Neither `BODY` nor `SET_BODY` trigger GC (I hope), and it is destroyed immediately after.

JanJecmen (approved):

o-:
good point. Somehow this was one of the asserts triggered in the no-segv.R errors. But maybe we should rather make sure, that nobody passes a not rir compiled closure?

o-:
@JanJecmen how about this version instead?

PR #113, opened 2018-04-17
fix compileWithGuess

o-:
looking up in R_GlobalEnv is actually quite wrong. For example:

    f <- function() 1
    g <- function() {
      f <- function() 42
      f()
    }

will return 1, instead of 42.

### DIFF
--- a/rir/src/ir/Compiler.cpp
+++ b/rir/src/ir/Compiler.cpp
@@ -36,9 +36,10 @@ class Context {
       public:
         CodeStream cs;
         std::stack<LoopContext> loops;
+        SEXP env;
         CodeContext* parent;
-        CodeContext(SEXP ast, FunctionWriter& fun, CodeContext* p)
-            : cs(fun, ast), parent(p) {}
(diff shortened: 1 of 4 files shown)


rir/src/ir/Compiler.cpp
```diff
@@ -1084,10 +1087,13 @@ SEXP findClosure(SEXP sym, SEXP rho) {
 
 // try to look up a closure
 bool compileWithGuess(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
+    if (!ctx.env())
+        return false;
+
     RList args(args_);
     CodeStream& cs = ctx.cs();
 
-    SEXP cls = findClosure(fun, R_GlobalEnv);  // or current env of the closure?
```
JanJecmen, line 1090: Thanks, my bad... Those are the comments that nobody ever deals with after merging

JanJecmen (approved):
I like that the compiler sets all closure fields now!

rir/src/ir/Compiler.cpp
o-: yeah :) well at least the comment gave me the right hint, what to fix

PR #114, opened 2018-04-19
make instruction description static lookup instead of virtual call

o-:
@mhyee hah, that was fun


mhyee (approved):
I think this is actually a bit cleaner! At least, there's slightly less template and virtual magic, but a bit more macro and static magic.

PR #115, opened 2018-04-19
try if this works on osx

o-:


charig:
Not sure if it is ok but I attach a patch with which I manage to make it work:

charig (changes requested):
Not sure if it is ok, but I managed to make it work on OSX with the following patch:
```
diff --git a/rir/src/compiler/pir_tests.cpp b/rir/src/compiler/pir_tests.cpp
index 1ff3845..0bcc042 100644
--- a/rir/src/compiler/pir_tests.cpp
+++ b/rir/src/compiler/pir_tests.cpp
@@ -90,9 +90,10 @@ bool test42(const std::string& input) {
     return true;
 };

-class NullBuffer : public std::ostream {
+class NullBuffer : public std::ostream, std::streambuf {
   public:
-    int overflow(int c) { return (c == traits_type::eof()) ? '\0' : c; }
+    NullBuffer() : std::ostream(this) {}
+    int overflow(int c) { return (c == std::ostream::traits_type::eof()) ? '\0' : c; }
 };

 bool verify(Module* m) {
@@ -106,6 +107,7 @@ bool verify(Module* m) {
     // TODO: find fix for osx
     NullBuffer nb;
     m->print(nb);
+    //m->print(std::cout);

     return true;
 }
```

o-:
ah, awesome, thanks :) ehm, you can directly push to my branch if you want, else I will do it later

charig (approved):

o-:
@charig thanks a lot

PR #116, opened 2018-04-25
Another refac?

charig:


o-:
uhm, this has no code :)
except for an unrelated change to gnur

charig:
Sorry. Wrong branch. The proper PR is [here](https://github.com/reactorlabs/rir/pull/117)

PR #117, opened 2018-04-25
Compilers are also translators

charig:


o-:
can you factor out the submodule update, that seems unrelated. Also, you need to push to the gnur repo first, so the commit you refer to with the submodule actually exists, otherwise travis fails.

o-:
hmm, there is some problem on my machine
```
'Query::noEnv(f)' failed
R: ../../rir/src/compiler/pir_tests.cpp:86: bool (anonymous namespace)::test42(const std::string &): Assertion `false' failed.
```

charig:
Sorry I do not know why I did such a mess. The cleaned commit is in a new [PR](https://github.com/reactorlabs/rir/pull/119)

PR #118, opened 2018-04-26
OMG Bugs Bugs Bugs

o-:
* Merge of abstract value unknown with non-unknown, did not go to top.
* Merge of two environments where a variable is missing in one,
  did not set this variable to unknown, depending on the merge order.
* Dead store elimination was oblivious to unknown loads (that could
  potentially make them observable).
* Static analysis (and a bunch of others) used the size of the cfg
  to size a BB id indexed vector. Unfortunately there is absolutely
  no connection between the number of BB and the largest id...

### DIFF
--- a/rir/src/compiler/analysis/scope.h
+++ b/rir/src/compiler/analysis/scope.h
@@ -19,9 +19,14 @@ namespace pir {
  *
  */
 class ScopeAnalysis {
-  public:
-    std::unordered_map<Instruction*, AbstractLoad> loads;
     std::unordered_set<Instruction*> observedStores;
+    std::unordered_set<Value*> allStoresObserved;
+
+  public:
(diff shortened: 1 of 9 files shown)


charig (approved):

rir/src/compiler/analysis/scope.h
```diff
@@ -19,9 +19,15 @@ namespace pir {
  *
  */
 class ScopeAnalysis {
+    std::unordered_set<Instruction*> observedStores;
+    std::unordered_set<Value*> allStoresObserved;
+
   public:
+    bool deadStore(Instruction* i) {
+        return !allStoresObserved.count(i->env()) &&
+               observedStores.count(i) == 0;
```
mhyee: `!observedStores.count(i)`

mhyee (approved):

PR #119, opened 2018-04-26
Try to unify all translators into a cleaner api: One Compiler that ju…

charig:


o-:
hmm, so somehow it still fails on my machine. I have to actually add more Cleanup passes, for it to work..

o-:
another strange thing, that I am not sure why it happens:
```
$ ~/Documents/rir/external/custom-r/bin/R  -f /tmp/r-test.K6YFmP -d valgrind --debugger-args=--leak-check=full
...
==9118== 544 (80 direct, 464 indirect) bytes in 1 blocks are definitely lost in loss record 166 of 1,889
==9118==    at 0x4C2E1CA: operator new(unsigned long) (vg_replace_malloc.c:334)
==9118==    by 0x124505A2: rir::pir::Builder::createBB() (builder.cpp:7)
==9118==    by 0x12450658: rir::pir::Builder::Builder(rir::pir::Closure*, rir::pir::Value*) (builder.cpp:11)
==9118==    by 0x12424FC3: rir::pir::PirCompiler::compile(rir::pir::IRCode) (pir_compiler.cpp:55)
==9118==    by 0x12429880: rir::pir::Rir2Pir::translate(rir::Function*, rir::Code*)::$_1::operator()(rir::Opcode*) const (rir_2_pir.cpp:191)
==9118==    by 0x12429491: std::_Function_handler<void (rir::Opcode*), rir::pir::Rir2Pir::translate(rir::Function*, rir::Code*)::$_1>::_M_invoke(std::_Any_data const&, rir::Opcode*&&) (std_function.h:316)
==9118==    by 0x124354C7: std::function<void (rir::Opcode*)>::operator()(rir::Opcode*) const (std_function.h:706)
==9118==    by 0x124291E7: (anonymous namespace)::Matcher<4ul>::operator()(rir::Opcode*, rir::Opcode*, std::function<void (rir::Opcode*)>) const (rir_2_pir.cpp:40)
==9118==    by 0x1242868A: rir::pir::Rir2Pir::translate(rir::Function*, rir::Code*) (rir_2_pir.cpp:168)
==9118==    by 0x12439A86: rir::pir::StackMachine::runCurrentBC(rir::pir::Rir2Pir&, rir::pir::Builder&) (stack_machine.cpp:152)
==9118==    by 0x124286EC: rir::pir::Rir2Pir::translate(rir::Function*, rir::Code*) (rir_2_pir.cpp:201)
==9118==    by 0x12427B2B: rir::pir::Rir2Pir::apply(rir::pir::IRCode) (rir_2_pir.cpp:51)
==9118== 
==9118== LEAK SUMMARY:
==9118==    definitely lost: 80 bytes in 1 blocks
==9118==    indirectly lost: 464 bytes in 8 blocks
==9118==      possibly lost: 0 bytes in 0 blocks
==9118==    still reachable: 28,377,208 bytes in 13,305 blocks
==9118==         suppressed: 0 bytes in 0 blocks
```

rir/src/compiler/translations/ir_translator.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+class IRTranslator {
+  public:
+    IRTranslator(RirCompiler& cmp, std::string name)
+        : cmp(cmp), name(name) {}
+
+    virtual ~IRTranslator() {}
+    
+    virtual void apply(IRCode) = 0;
```
o-, line 32: What is the idea with this api? wouldn't it be more reasonable to use overloading, ie. `apply(Closure*)` and `apply(rir::Function*)`

rir/src/compiler/translations/ir_translator.h
```diff
@@ -0,0 +1,43 @@
...
+#include "../pir/closure.h"
+#include <string>
+
+
+namespace rir {
+namespace pir {
+
+class RirCompiler;
+struct RirInput;
+typedef Closure PirInput;
```
o-, line 14: What is the idea with the typedef? Why not use Closure. I would not have guessed that Closure is the PirInput.

rir/src/compiler/translations/ir_translator.h
```diff
@@ -0,0 +1,43 @@
...
+
+union IRCode
```
o-, line 16: So unions are a bit dangerous. Because they are untagged, they do not give you any guarantees. So usually you want to encapsulate them in a struct, that remembers the kind.

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
```diff
@@ -142,23 +143,33 @@ void StackMachine::runCurrentBC(Rir2Pir& rir2pir, Builder& insert) {
...
             rir::Code* promiseCode = srcFunction->codeAt(argi);
             Promise* prom = insert.function->createProm();
+            RirCompiler& cmp = rir2pir.compiler();
             {
+                Builder* oldBuilder = cmp.getBuilder();
                 Builder promiseBuilder(insert.function, prom);
-                Rir2Pir compiler(rir2pir.compiler(), promiseBuilder,
-                                 srcFunction, promiseCode);
-                compiler.translate();
+                cmp.setBuilder(&promiseBuilder);
```
o-, line 150: sorry, but I really do not agree with this! I think swapping the builder back and forth like this is really brittle. I just see a lot more ways to get it wrong and no benefit. Think of it in terms of invariants. Can you formulate the  invariant of the compiler object? It's something like: "While a Rir2Pir object that points to a RirCompiler object is executing it's translate function, the builder field of the compiler object points to a builder that corresponds to the the code being compiled by the translate function of the rir2pir object, but temporarily it can also point to another builder, but will be set back, before the outer translate function regains control."

o- (commented):
hey, sorry, I am not really through with reading the code. Here are just some preliminary comments, since i need to leave.

 My worry for this pr is a bit, that I don't see the direction. maybe you can explain a bit? because usually when I see a refactoring, i think of it in terms of some future change it will facilitate. but I am not sure what the future usecase of the abstractions you propose are.

rir/src/compiler/translations/ir_translator.h
mhyee: Yeah, I agree with @o-. I understand the desire to have some sort of common API, but I find it more natural to use overloading than to use an `IRCode` union. (And I don't think it's a good idea to create some sort of common superclass or interface for `Closure*` and `rir::Function*`.)

rir/src/compiler/opt/force_dominance.h
```diff
@@ -15,11 +15,11 @@ namespace pir {
  *
  */
 class Closure;
-class ForceDominance : public PirTranslator {
+class ForceDominance : public IRTranslator {
   public:
-    ForceDominance() : PirTranslator("force dominance") {};
+    ForceDominance(RirCompiler& compiler) : IRTranslator(compiler, "force dominance") {};
```
mhyee, line 20: While we're at it: `"Force Dominance"` because all the other names use title case.

rir/src/compiler/opt/scope_resolution.h
```diff
@@ -13,12 +13,11 @@ namespace pir {
  * environment, to pir SSA variables.
  *
  */
-class Closure;
-class ScopeResolution : public PirTranslator {
+class ScopeResolution : public IRTranslator {
   public:
-    ScopeResolution() : PirTranslator("scope resulotion") {};
+    ScopeResolution(RirCompiler& compiler) : IRTranslator(compiler, "scope resulotion") {};
```
mhyee, line 18: `"scope resulotion"` -> `"Scope Resolution"`

rir/src/compiler/translations/ir_translator.h
```diff
@@ -0,0 +1,43 @@
...
+        : cmp(cmp), name(name) {}
+
+    virtual ~IRTranslator() {}
+    
+    virtual void apply(IRCode) = 0;
+    std::string getName() { return this->name; }
+    RirCompiler& compiler() { return cmp; }
+
+  protected:
+    RirCompiler& cmp;
```
mhyee, line 37: Maybe I'm missing something, but why does `IRTranslator` need a reference to a `RirCompiler`?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -45,21 +46,28 @@ struct Matcher {
...
+    translate(srcFunction, srcFunction->body());
+    //input.pirClosure 
+} 
+
+Value* Rir2Pir::translate(rir::Function* srcFunction, rir::Code* srcCode) {
     assert(!done);
     done = true;
 
-    recoverCFG(srcCode);
+    Builder* insert = cmp.getBuilder();
```
mhyee, line 59: `Builder& insert = *cmp.getBuilder()`. Yeah, it looks a bit weird, but it lets us avoid writing `(*insert)(...)` which I think looks weirder.

mhyee (commented):
I'm still getting familiar with the code base, so these comments might be a bit superficial.

I agree that there's some elegance to having a common API for the RIR->PIR and PIR->RIR compilers, and also the PIR->PIR optimizers. But is it natural to unify the API? Does it make things simpler and more extensible?

One counterpoint is that the `IRCode` union seems really awkward. But that's probably something we could work around, if we decide that we do want a unified API.

Another counterpoint is that there's really only three kinds of translators (RIR->PIR, PIR->RIR, PIR->PIR) and two kinds of IR (RIR and PIR). Maybe if we had even more kinds of IR and more layers of translations, it would make sense to do the refactor.

charig:
@o- @mhyee: Thanks for your comments. I wrote in the description that this refactoring was more controversial. So I agree mostly with all what you said. There some ugly parts that were mainly because I wanted to change as less as possible the current implementation such as why we need the compiler in the translator, etc. But let me explain as brief as possible my main idea: 

What I really wanted to introduce was the idea that we should have one compiler composing translations. And that the first translation (rir 2 pir) and the latter translation (pir 2 rir) should be just one more step. Nowadays they are kindof separated and the introduction of the pir 2 rir will make it worst. I am not sure even if this is the proper approach. But that is why I decided to unify the api. Anyway, I agree that we do not have an explosion of IRs and that maybe overloading is the best alternative here. On the other hand we have already talked about native code.

So conclusion...we can kill this PR but it may be interesting to discuss the main idea.

mhyee:
Maybe we should postpone this discussion until #108 is finished and merged? I do think it's worth discussing (thanks for bringing it up), and if we start seriously considering native code, then we should definitely come back to this.

o-:
archiving this for now

PR #120, opened 2018-05-09
A new Register allocator

o-:
This patch adds a new liveness interval algorithm, a naive algo for
sinking ssa variables to stack slots and a coloring register alloc
for the remaining variables.

Liveranges are stored in the following form:

    Instruction -> BB -> {live : bool, start : pos, end : pos}

Two variables do not interfere if: for all BB: either they are not
both live, or their start-end intervals do not overlap.

Stack slots are assigned for variables which are produced and
consumed exactly once in the same BB. Additionally trivial phis,
which are produced at the end of all input BBs and the consumed
by a phi at the beginning of the successor block are sunk to stack
slots.

The approach is fairly naive and should be improved.

The remaining registers are allocated in a linear pass on the
dominator tree. According to Hack and Goos, this should be optimal.

UPDATE: we have established offline, that our `removePhi` transformation is in fact moving to CSSA. I leave the below comment of reference. The reason why it works is that `x1` from below will be split into `x1'` which goes into the phi and the old `x1` which goes into `y = x1+x2`.

> Caveat: we precolor all phi, such that phi and all inputs are
> coalesced into the same input. Also as of yet, we do not convert
> to CSSA. I conjecture that there should be a program that breaks
> this approach. Something like:
> 
>     x1       x2         x3
> 
>       \      /         /
>          y = x1+x2
>             \        /
>            x4 = phi(x1,x2,x3)
> 
> But so far I haven't been able to construct it, or understand
> why it couldn't happen. More investigation is necessary and
> probably another reading of Boissinot etal.
> 

Unfortunately some other things got lumped into this PR:

* The verifier was not run after every optimization
* Delay instruction did delay phis (which is never a good idea)
* Scope resolution was broken for `LdFun`. This instruction
  performs an implicit `ChkClosure` test. If we replace a `LdFun`
  with something that is not of static type `closure`, we need
  to add that runtime check.
* Scope resolution stupidly places phis. Changed it to move
  phi's up the cfg to the point where the inputs actually come
  from different predecessors.

### DIFF
--- a/rir/src/compiler/analysis/abstract_value.cpp
+++ b/rir/src/compiler/analysis/abstract_value.cpp
@@ -54,6 +54,24 @@ void AbstractPirValue::print(std::ostream& out) {
     out << ") : " << type;
 }
 
+MkFunCls* AbstractREnvironmentHierarchy::findClosure(Value* env, Value* fun) {
+    for (;;) {
+        if (Force::Cast(fun)) {
+            fun = Force::Cast(fun)->arg<0>().val();
+        } else if (ChkClosure::Cast(fun)) {
+            fun = ChkClosure::Cast(fun)->arg<0>().val();
+        } else {
+            break;
+        }
+    }
+    while (env && env != AbstractREnvironment::UnknownParent) {
+        if ((*this)[env].mkClosures.count(fun))
+            return (*this)[env].mkClosures.at(fun);
+        env = (*this)[env].parentEnv;
+    }
+    return AbstractREnvironment::UnknownClosure;
+}
+
 AbstractLoad AbstractREnvironmentHierarchy::get(Value* env, SEXP e) const {
     while (env != AbstractREnvironment::UnknownParent) {
         if (this->count(env) == 0)
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -9,8 +9,9 @@ using namespace rir::pir;
 class TheVerifier {
   public:
     Closure* f;
+    CFG cfg;
 
-    TheVerifier(Closure* f) : f(f) {}
+    TheVerifier(Closure* f) : f(f), cfg(f->entry) {}
 
     bool ok = true;
 
@@ -26,7 +27,8 @@ class TheVerifier {
         }
 
         for (auto p : f->promises) {
-            verify(p);
+            if (p)
+                verify(p);
             if (!ok) {
                 std::cerr << "Verification of promise failed\n";
                 p->print(std::cerr);
@@ -34,7 +36,8 @@ class TheVerifier {
             }
         }
         for (auto p : f->defaultArgs) {
-            verify(p);
+            if (p)
+                verify(p);
             if (!ok) {
                 std::cerr << "Verification of argument failed\n";
                 p->print(std::cerr);
@@ -50,6 +53,16 @@ class TheVerifier {
             if (!bb->next0 && !bb->next1) {
                 std::cerr << "bb" << bb->id << " has no successor\n";
                 ok = false;
+            }
+            if (cfg.predecessors[bb->id].size() > 1) {
+                for (auto in : cfg.predecessors[bb->id]) {
+                    if (in->next1) {
+                        std::cerr << "BB " << in->id << " merges into "
+                                  << bb->id << " and branches into "
+                                  << in->next1->id << " at the same time.\n";
+                        ok = false;
+                    }
+                }
             }
         } else {
             Instruction* last = bb->last();
@@ -94,7 +107,15 @@ class TheVerifier {
             Instruction* iv = Instruction::Cast(v);
             if (iv) {
                 if (phi) {
-                    // TODO: what is a good condition for phi inputs?
+                    if (!cfg.transitivePredecessors[i->bb()->id].count(
+                            iv->bb())) {
+                        std::cerr << "Error at instruction '";
+                        i->print(std::cerr);
+                        std::cerr << "': input '";
+                        iv->printRef(std::cerr);
+                        std::cerr << "' does not come from a predecessor.\n";
+                        ok = false;
+                    }
                 } else if ((iv->bb() == i->bb() &&
                             bb->indexOf(iv) > bb->indexOf(i)) ||
                            (iv->bb() != i->bb() &&
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -54,6 +54,27 @@ class TheVerifier {
                 std::cerr << "bb" << bb->id << " has no successor\n";
                 ok = false;
             }
+            // This check verifies that our graph is in edge-split format.
+            // Currently we do not rely on this property, however we should
+            // make it a conscious decision if we want to violate it.
+            // This basically rules out graphs of the following form:
+            //
+            //   A       B
+            //     \   /   \
+            //       C       D
+            //
+            // or
+            //     _
+            //  | / \
+            //  A __/
+            //  |
+            //
+            // The nice property about edge-split graphs is, that merge-points
+            // are always dominated by *both* inputs, therefore local code
+            // motion can push instructions to both input blocks.
+            //
+            // In the above example, we can't push an instruction from C to A
+            // and B, without worrying about D.
             if (cfg.predecessors[bb->id].size() > 1) {
                 for (auto in : cfg.predecessors[bb->id]) {
                     if (in->next1) {
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -54,27 +54,28 @@ class TheVerifier {
                 std::cerr << "bb" << bb->id << " has no successor\n";
                 ok = false;
             }
-            // This check verifies that our graph is in edge-split format.
-            // Currently we do not rely on this property, however we should
-            // make it a conscious decision if we want to violate it.
-            // This basically rules out graphs of the following form:
-            //
-            //   A       B
-            //     \   /   \
-            //       C       D
-            //
-            // or
-            //     _
-            //  | / \
-            //  A __/
-            //  |
-            //
-            // The nice property about edge-split graphs is, that merge-points
-            // are always dominated by *both* inputs, therefore local code
-            // motion can push instructions to both input blocks.
-            //
-            // In the above example, we can't push an instruction from C to A
-            // and B, without worrying about D.
+            /* This check verifies that our graph is in edge-split format.
+               Currently we do not rely on this property, however we should
+               make it a conscious decision if we want to violate it.
+               This basically rules out graphs of the following form:
+
+                 A       B
+                   \   /   \
+                     C       D
+
+               or
+                   _
+                | / \
+                A __/
+                |
+
+               The nice property about edge-split graphs is, that merge-points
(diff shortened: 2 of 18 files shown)


rir/src/compiler/analysis/abstract_value.cpp
```diff
@@ -54,6 +54,24 @@ void AbstractPirValue::print(std::ostream& out) {
     out << ") : " << type;
 }
 
+MkFunCls* AbstractREnvironmentHierarchy::findClosure(Value* env, Value* fun) {
+    for (;;) {
```
mhyee, line 58: Is `while (true)` too much to type? 😄

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -24,6 +24,7 @@ class TheCleanup {
                 Instruction* i = *ip;
                 auto next = ip + 1;
                 Force* force = Force::Cast(i);
+                ChkClosure* chkcls = ChkClosure::Cast(i);
```
mhyee, line 27: What's the difference between `chkcls` and `closure`?

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -36,6 +37,12 @@ class TheCleanup {
                         force->replaceUsesWith(arg);
                         next = bb->remove(ip);
                     }
+                } else if (chkcls) {
+                    Value* arg = chkcls->arg<0>().val();
+                    if (PirType(RType::closure).isSuper(arg->type)) {
```
mhyee: And this is almost identical to the case on line 52, except that it uses `PirType(RType::closure)` instead of `PirType::val()`.

rir/src/compiler/util/builder.cpp
```diff
@@ -4,14 +4,14 @@
...
+BB* Builder::createBB() { return new BB(code, ++code->maxBBId); }
 
 Builder::Builder(Closure* fun, Value* closureEnv)
     : function(fun), code(fun), env(nullptr), bb(fun->entry) {
     bb = function->entry = createBB();
-    std::vector<Value*> args;
-    for (size_t i = 0; i < fun->argNames.size(); ++i)
-        args.push_back(this->operator()(new LdArg(i)));
+    std::vector<Value*> args(fun->argNames.size());
+    for (long i = fun->argNames.size() - 1; i >= 0; --i)
```
JanJecmen, line 13: Why `long`? Just curious...

rir/src/compiler/util/visitor.h
```diff
@@ -187,19 +193,67 @@ class VisitorImplementation {
...
+                }
+                done.set(next);
+            };
+
+            apply(cur->next0);
+            apply(cur->next1);
+
+            action(cur);
+        }
+        assert(todo.empty());
```
JanJecmen: `&& delayedTodo.empty()`?
Anyway, what's the reason for the assert? The loop only exits if both stacks are empty. Is it just for code readability?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -70,52 +72,74 @@ class TheScopeResolution {
...
-                            if (!aval.isSingleValue() && !aval.isUnknown()) {
-                                auto phi = new Phi;
+                        // This load can have multiple values. We need a phi
+                        // to distinguish them. (LdFun case is not yet
+                        // handled here)
+                        if (!aval.isSingleValue() && !aval.isUnknown() &&
+                            !ldfun) {
+                            auto hasAllInputs = [&](BB* load) {
+                                bool success = true;
                                 aval.eachSource([&](ValOrig& src) {
```
JanJecmen: Just an idea: why not have something like the visitor check that would run only until first fail?
Also, unrelated, I looked at what `ValOrig` is, maybe it could inherit from `std::pair`? Would eliminate the need for `operator<`/`==`

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -26,73 +28,499 @@ namespace pir {
...
+          bbsSize(code->maxBBId + 1) {
+        computeLiveness();
+        computeStackAllocation();
+        computeAllocation();
+    }
+
+    // Run backwards analysis to compute livenessintervals
+    void computeLiveness(bool verbose = false) {
+        // temp list of live out sets for every BB
+        std::unordered_map<BB*, std::set<Value*>> liveAtEnd(bbsSize);
```
JanJecmen, line 105: Could be just a `vector` since the keys are BBs

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -26,73 +28,499 @@ namespace pir {
...
+            todo.erase(todo.begin());
+
+            // keep track of currently live variables
+            std::set<Value*> accumulated;
+            std::map<BB*, std::set<Value*>> accumulatedPhiInput;
+
+            // Mark all (backwards) incoming live variables
+            for (auto v : liveAtEnd[bb]) {
+                auto& liveRange = livenessInterval[v][bb->id];
+                assert(livenessInterval.count(v));
```
JanJecmen: This assert always passes since the prev line inserts `v` if missing

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -26,73 +28,499 @@ namespace pir {
...
+                }
+                std::cout << "\n";
+            }
+        }
+    }
+
+    void computeStackAllocation() {
+        Visitor::run(code->entry, [&](BB* bb) {
+            {
+                // Phi at the beginning of BB, all inputs at the end of
```
JanJecmen: Is this now the case? that all phis are at the beginning? Also, they are still sequential, right? No parallel semantics yet

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -26,73 +28,499 @@ namespace pir {
...
+                        ++check;
+                        --newStackSize;
+                    }
                 });
+
+                if (!foundAll)
+                    return;
+
+                // pop args from stack, discarding all unmatched values
+                // in the process
```
JanJecmen: I don't understand this, how can we discard the values that are not matched?
I mean, if the stack contains `[... A B C]` and the instruction expects `A` and `C`, I'd think we cannot allocate `A` to stack, but this looks like we do and we throw away `B`

JanJecmen (approved):
Hi, nice one!
As a side note, we should probably rename `removePhis` to something like `convertToCSSA` because the phis are not actually removed there...

rir/src/compiler/util/builder.cpp
o-: for size_t (aka unsigned) the `i >= 0` is true by definition and optimized away

rir/src/compiler/translations/pir_2_rir.cpp
o-: the advantage is that i can distinguish between absent/present, to figure out when a bb is visited the first time vs. subsequent visits, which need merging

rir/src/compiler/util/builder.cpp
JanJecmen: Yeah, I know, only I've never seen long, always int or long long  :)

rir/src/compiler/translations/pir_2_rir.cpp
o-: no. but if they happen to be at the beginning, then it works

rir/src/compiler/translations/pir_2_rir.cpp
o-: what we do here is: the stack contains `xxx A B C` and they are all only used once. then we resize the stack to `xxx`, mark `A, C` to be on the stack and put `B` in a register

rir/src/compiler/util/builder.cpp
o-: ah, ok, not sure I guess int wold do it here

rir/src/compiler/translations/pir_2_rir.cpp
o-: ups, thanks

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: Oh right, this is not what's actually on the stack... The B will get stored to a local by its producer in this case. Thanks for the explanation!

rir/src/compiler/util/builder.cpp
JanJecmen: The only difference I found is that long is at least 32 bits while int can be 16 on some hw...
Doesn't matter really, it just caught my eye

mhyee (commented):
I'm not sure if I can finish my review this week (still need to go over `scope_resolution.cpp` and `pir_2_rir.cpp`), so here's some quick comments for now.

rir/src/compiler/opt/cleanup.cpp
o-: closure creates a closure, chcls throws an error if the input is not a closure

rir/src/compiler/analysis/abstract_value.cpp
o-: i find it aesthetically pleasing :)

rir/src/compiler/opt/cleanup.cpp
mhyee: Wait, does `ChkClosure::Cast(i)` have side effects? What's the difference between line 27 and line 29?

o-: oh, sorry, I didn't actually look at the code. Yeah, so the difference is that line 53 is wrong and I didn't see it, when I added line 40 :) So closure should just go away

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -51,6 +54,16 @@ class TheVerifier {
                 std::cerr << "bb" << bb->id << " has no successor\n";
                 ok = false;
             }
+            if (cfg.predecessors[bb->id].size() > 1) {
+                for (auto in : cfg.predecessors[bb->id]) {
+                    if (in->next1) {
+                        std::cerr << "BB " << in->id << " merges into "
+                                  << bb->id << " and branches into "
+                                  << in->next1->id << " at the same time.\n";
```
charig, line 84: I am not fully sure how we are treating self pointing edges. If they are consider predecessors, successors or both. I think "theoretically" they are both. And in that I think this is not an invalid property. It is exactly what happens with *critical edges*. They have two predecessors (one is themselves) and they have two successors (one is themselves). Perhaps, it would be good to add a comment with the underlying logical property that the if is veryfing when there are some subtleties?

o-: yeah, I think we should not allow those. I believe that we do not generate any and would like to keep it that way. They are nasty edge-cases for many transformations and easily avoidable.

o-: but I was very sneaky and put it into this unrelated PR :)

o-: but if there is the wish to discuss further, I can remove this check

rir/src/compiler/opt/scope_resolution.cpp
o-: regarding inheriting from pair: yes, but it would make naming the fields more cumbersome (ie. would need functions, which means additional `()` for the clients. not sure what's better

PR #129, opened 2018-05-15
Cleanup

JanJecmen:
This adds dead code removal from RIR cleanup pass - anything after `br_`, `ret_` and `return_` is dead until the next label or the end of the codestream.

Also fixes the pass to not dispatch on instructions that are marked to be deleted.

Returns are propagated now back if they are the target of a `br_`.

To make this work, handling of `ret_` in pir2rir needed to be changed.

Lastly, changed how we translate branches back to RIR: instead of guessing, now always use `brfalse_` and possibly change to `brtrue_` in RIR cleanup:
```
   ...			   ...
   brfalse_ 0		   brtrue_ 1
   br_ 1		   ->	   
0:			0:
   ...			   ...
```

### DIFF
--- a/rir/src/ir/cleanup.h
+++ b/rir/src/ir/cleanup.h
@@ -90,11 +90,54 @@ class BCCleanup : public InstructionDispatcher::Receiver {
     }
 
     void br_(CodeEditor::Iterator ins) override {
-        // remove br_ x; x: ...
-        if (ins + 1 != code_.end()) {
-            if (code_.target(*ins) == ins + 1) {
+        auto next = removeDead(ins + 1);
+        if (next == code_.end())
+            return;
+        // next is a label here
+        auto target = code_.target(*ins);
+        if (target == next) {
+            ins.asCursor(code_).remove();
+            return;
+        }
+        auto cont = target + 1;
+        if (cont == code_.end())
+            return;
+        // if the continuation is return, propagate to the prev block
+        if ((*cont).isReturn()) {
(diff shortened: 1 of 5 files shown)


rir/src/ir/cleanup.h
```diff
@@ -129,20 +172,25 @@ class BCCleanup : public InstructionDispatcher::Receiver {
...
-    // }
+    void ret_(CodeEditor::Iterator ins) override { removeDead(ins + 1); }
+
+    void return_(CodeEditor::Iterator ins) override { removeDead(ins + 1); }
 
     void run() {
-        for (auto i = code_.begin(); i != code_.end(); ++i)
-            dispatcher.dispatch(i);
+        for (auto i = code_.begin(); i != code_.end(); ++i) {
+            if (!i.deleted())
```
o-: maybe you should use a cursor here?

I seem to remember that our idea was, that the iterator always shows you the original code and the cursor shows you the edited code. therefore if you use a cursor, it should skip the deleted BCs

o-: ah, but the api of the dispatcher wants an iterator.

I wonder why there could be deleted instructions here? is it that you run it multiple times without comiting the changes?

o-: ah, now I understand, you remove more than one instruction, and then the dispatcher catches up, with instructions you already removed.

ok, that sounds reasonable then. maybe add a comment.

o- (approved):
lgtm

PR #130, opened 2018-05-15
remove unused deoptimizer

JanJecmen:
What about this? @o-


o-:
given we might want to have deoptimization at some point, wouldn't it be good to keep at least the deopt table? it's probably not exactly what we need, but might be a start? what do you think?

JanJecmen:
I just remembered a comment that we should get rid of the deoptimizer :)
I agree we want to have deopt for sure. I personally don't like that it's in C, but it can be used as a starting point. So let's just close this and rewrite in the future.
Or maybe just remove the `guard_env_` instruction?
And also, do we want to get rid of the FRAME_CHANGED/LEAKED bits?

PR #132, opened 2018-05-21
P2R2

JanJecmen:
I tried to deal with the ordering of arguments -- they were popped from stack, stored in the instructions, then when allocating looked at in the reverse order. Not sure if this is the best way of going about it, but I changed the instructions to have the args such that TOS arg is last.

Also I added some missing instructions to the PIR to RIR compiler. Only deopt is missing now, but we need to test the heck out of it, because at several places I don't know what should happen.

For now I open this just for discussion I guess.

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -748,15 +748,19 @@ class VLI(Call, Effect::Any, EnvAccess::Leak), public CallInstructionI {
 // specified as `cls_`, args passed as promises.
 class VLI(StaticCall, Effect::Any, EnvAccess::Leak), public CallInstructionI {
     Closure* cls_;
+    SEXP origin_;
 
   public:
     constexpr static size_t callArgOffset = 1;
 
     Closure* cls() { return cls_; }
+    SEXP origin() { return origin_; }
     size_t nCallArgs() override { return nargs() - callArgOffset; }
 
-    StaticCall(Value * e, Closure * cls, const std::vector<Value*>& args)
-        : VarLenInstruction(PirType::valOrLazy(), e), cls_(cls) {
+    StaticCall(Value * e, Closure * cls, const std::vector<Value*>& args,
+               unsigned src, SEXP origin)
+        : VarLenInstruction(PirType::valOrLazy(), e), cls_(cls),
+          origin_(origin) {
         for (unsigned i = 0; i < args.size(); ++i)
             this->pushArg(args[i], RType::prom);
     }
@@ -775,17 +779,22 @@ class VLI(StaticCall, Effect::Any, EnvAccess::Leak), public CallInstructionI {
 class VLI(StaticEagerCall, Effect::Any, EnvAccess::Leak),
     public CallInstructionI {
     Closure* cls_;
+    SEXP origin_;
 
   public:
     constexpr static size_t callArgOffset = 1;
 
     Closure* cls() { return cls_; }
+    SEXP origin() { return origin_; }
     size_t nCallArgs() override { return nargs() - callArgOffset; }
 
-    StaticEagerCall(Value * e, Closure * cls, const std::vector<Value*>& args)
-        : VarLenInstruction(PirType::valOrLazy(), e), cls_(cls) {
+    StaticEagerCall(Value * e, Closure * cls, const std::vector<Value*>& args,
+                    unsigned src, SEXP origin)
+        : VarLenInstruction(PirType::valOrLazy(), e), cls_(cls),
+          origin_(origin) {
         for (unsigned i = 0; i < args.size(); ++i)
             this->pushArg(args[i], PirType::val());
+        srcIdx = src;
     }
 
     void eachCallArg(ArgumentValueIterator it) override {
@@ -840,10 +849,12 @@ class VLI(CallBuiltin, Effect::Any, EnvAccess::Write) {
 
 class VLI(CallSafeBuiltin, Effect::None, EnvAccess::None) {
   public:
+    SEXP blt;
     const CCODE builtin;
     int builtinId;
 
-    CallSafeBuiltin(SEXP builtin, const std::vector<Value*>& args);
+    CallSafeBuiltin(SEXP builtin, const std::vector<Value*>& args,
+                    unsigned src);
 
     void printArgs(std::ostream& out) override;
 };
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -587,20 +587,20 @@ class FLI(AsTest, 1, Effect::None, EnvAccess::None) {
 
 class FLI(Subassign1_1D, 3, Effect::None, EnvAccess::None) {
   public:
-    Subassign1_1D(Value* value, Value* index, Value* vec)
+    Subassign1_1D(Value* vec, Value* index, Value* val)
         : FixedLenInstruction(
               PirType::val(),
               {{PirType::val(), PirType::val(), PirType::val()}},
-              {{value, index, vec}}) {}
+              {{vec, index, val}}) {}
 };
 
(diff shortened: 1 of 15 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -157,6 +157,15 @@ class Instruction : public Value {
             it(arg(i));
     }
 
+    void eachStackArg(Instruction::ArgumentInstructionIterator it) const {
```
o-: can we call this: `reverseEachCallArg`.  Also, can we maybe remove the `Instruction` variants of the iterators again. Somehow I think they cause more confustion than they are worth.

Also, as a clarification:

for calls, the `eachCallArg` is skipping the env, since it is not a call arg. We should keep that convention, that every iterator that skips the env, should be called `...CallArg....`

rir/src/compiler/pir/instruction.h
```diff
@@ -578,20 +587,21 @@ class FLI(AsTest, 1, Effect::None, EnvAccess::None) {
 
 class FLI(Subassign1_1D, 3, Effect::None, EnvAccess::None) {
   public:
-    Subassign1_1D(Value* vec, Value* index, Value* value)
+    Subassign1_1D(Value* value, Value* index, Value* vec)
```
o-: maybe add a todo, that we should fix the order in rir

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -284,7 +285,7 @@ class SSAAllocator {
                 size_t newStackSize = stack.size();
```
o-, line 285: as you mentioned on slack, we should exclude Lazy calls here

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -929,11 +990,36 @@ size_t Pir2Rir::compile(Context& ctx, Code* code) {
                 break;
             }
             case Tag::StaticCall: {
-                assert(false && "not yet implemented.");
+                auto call = StaticCall::Cast(instr);
+                setEnv(it, call->env());
+
+                compiler.compile(call->cls(), call->origin());
+                // push callee - this stores it into constant pool - discuss!
+                cs << BC::push(call->origin());
```
o-: I think that's fine. we do the same in rir land

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -946,32 +1032,30 @@ size_t Pir2Rir::compile(Context& ctx, Code* code) {
...
+                cs << BC::makeEnv() << BC::setEnv();
                 currentEnv = mkenv;
                 // bind all args
                 // TODO: maybe later have a separate instruction for this?
                 mkenv->eachLocalVar([&](SEXP name, Value* val) {
                     load(it, val);
                     cs << BC::stvar(name);
                 });
+                cs << BC::getEnv();
                 store(it, mkenv);
```
o-: can't you do the store before you set all variables

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -946,32 +1032,30 @@ size_t Pir2Rir::compile(Context& ctx, Code* code) {
...
                 break;
             }
             case Tag::Phi: {
-                auto phi = Phi::Cast(instr);
-                phi->eachArg([&](BB*, Value* arg) {
-                    assert(((alloc.onStack(phi) && alloc.onStack(arg)) ||
-                            (alloc[phi] == alloc[arg])) &&
-                           "Phi inputs must all be allocated in 1 slot");
```
o-: why is this no longer true?

o- (commented):
hey, thanks a lot, that's quite some cool stuff

rir/src/compiler/translations/pir_2_rir.cpp
o-: and we should also exclude implicit arguments. There are actually a bunch of them. From the top of my head:
* env() for the call instructions
* env() for the st-/ldvar

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -929,11 +990,36 @@ size_t Pir2Rir::compile(Context& ctx, Code* code) {
...
+                auto call = StaticCall::Cast(instr);
+                setEnv(it, call->env());
+
+                compiler.compile(call->cls(), call->origin());
+                // push callee - this stores it into constant pool - discuss!
+                cs << BC::push(call->origin());
+
+                std::vector<FunIdxT> callArgs;
+                call->eachCallArg([&](Value* arg) {
+                    // TODO: for now, ignore the eager value in MkArg
```
o-: actually if all of them have a non-missing eager arg, then you could do callStack instead...

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: It is but it is checked by `SSAAllocator::verify()`

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: Good idea, thanks!
Question, though, do you think this should restore the old environment, or leave the newly created one as current env?

rir/src/compiler/translations/pir_2_rir.cpp
o-: uhm, ok, but still, we probably want to run the verifier at the end of compilation, whereas the allocator verify is for debugging

rir/src/compiler/pir/instruction.h
JanJecmen: But then we have two things named `...CallArg...`:
* For calls it is actually the callargs that we pass to a callee;
* For all instructions when allocating to stack they are args to that instruction that it expects on the stack :(

rir/src/compiler/pir/instruction.h
JanJecmen: I changed it in rir

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: So I went for virtual dispatch for now, where the default is that it skips env(), and instructions that don't want to skip it / want to skip something else can override..

o-: that part is not yet pushed, I guess. Or did I miss it?

rir/src/compiler/pir/instruction.h
o-: I see, that's how you where thinking about it.

the problem is that there might not be a 1:1 mapping between pir and rir instructions. So I am not convinced that pir instructions should know about the arguments needed by some rir instruction they might eventually be translated into. So I think pir instructions should only know about themselves, and the intelligence (how to map, reverse map, skip some args) should be in the function/class/module that does the translation.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -554,6 +575,76 @@ class SSAAllocator {
...
+                while (!args.empty()) {
+                    action(args.back());
+                    args.pop_back();
+                }
+            };
+            switch (i->tag) {
+            case Tag::MkArg: {
+                // just the eager value, if it's not missing
+                auto mkarg = MkArg::Cast(i);
+                if (mkarg->hasEagerArg())
```
o-: You could instead of `hasEagerArg` and `eagerArg` have the following api:
```
InstrArg::ifEager(std::function<void(Value*)> maybe) {
  if (val() != Missing)
    maybe(val());
}
```
And then you can write:

`MkArg::Cast(i)->arg<0>.ifEager(action);`

o-:
awesom, lgtm

Issue #133, opened 2018-05-23
Make optimizations configurable


charig:
Implement some kind of configuration support at startup so clients can customize easily which optimizations are going to be performed by rir.

PR #134, opened 2018-05-23
Add support to rir so that the optimizations are configurable via an INI file

charig:
Implements #133

### DIFF
--- a/rir/src/utils/configurations.cpp
+++ b/rir/src/utils/configurations.cpp
@@ -1,5 +1,5 @@
-#include <sstream>
 #include <iostream>
+#include <sstream>
 
 #include "configurations.h"
 
@@ -12,11 +12,12 @@
 #include "../compiler/opt/scope_resolution.h"
 
 namespace rir {
-    
-void Configurations::parseINIFile() {    
+
+void Configurations::parseINIFile() {
     INIReader reader(".pir/configuration.ini");
     if (reader.ParseError() < 0) {
-        std::cout << "Can't load optimizations.ini file, we resort to default optimizations'\n";
+        //std::cout << "Can't load optimizations.ini file, we resort to default "
+        //             "optimizations'\n";
         return defaultOptimizations();
     }
     read(reader, "globalValueNumber");
@@ -28,11 +29,11 @@ void Configurations::parseINIFile() {
     string cleanups = reader.Get("optimizations", "cleanup", "UNKNOWN");
     std::stringstream data(cleanups);
     std::string order;
-    while(std::getline(data, order, ','))
-    {
-        optimizations.insert(new Optimization(new pir::Cleanup(), std::stoi( order )));
+    while (std::getline(data, order, ',')) {
+        optimizations.insert(
+            new Optimization(new pir::Cleanup(), std::stoi(order)));
     }
-}    
+}
 
 void Configurations::defaultOptimizations() {
     optimizations.insert(new Optimization(new pir::ForceDominance(), 1));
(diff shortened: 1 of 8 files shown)


rir/src/utils/configurations.h
```diff
@@ -0,0 +1,39 @@
...
+struct Optimization {
+    Optimization(pir::PirTranslator* translator, short order) : translator(translator), order(order) {}
+    ~Optimization() { delete translator; }
+    
+    pir::PirTranslator* translator;
+    short order;
+
+    bool operator < (const Optimization& anotherOptimization) const {
+        if (order == anotherOptimization.order) {
+            return anotherOptimization.translator->getName().compare("cleanup");
```
o-: `compare` behaves like c `strcmp`, see [here](http://www.cplusplus.com/reference/string/string/compare/).
Therefore if `a` is shorter than `b`, this will return `-1`, which is `true` (everything except `0` is).

you probably want to say:  `a < b`  iff  `a.compare(b) < 0`

rir/src/utils/configurations.h
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+            return order <= anotherOptimization.order;
```
o-: needs to be `<`

rir/src/utils/configurations.cpp
```diff
@@ -0,0 +1,67 @@
...
+    optimizations.insert(new Optimization(new pir::DelayInstr(), 5));
+    optimizations.insert(new Optimization(new pir::ElideEnv(), 6));
+    optimizations.insert(new Optimization(new pir::DelayEnv(), 7));
+    optimizations.insert(new Optimization(new pir::Cleanup(), 8));
+}
+
+void Configurations::read(INIReader& reader, string optimizationName) {
+    short order = reader.GetInteger("optimizations", optimizationName, 0);
+    if (order) {
+        if (!optimizationName.compare("globalValueNumber")){
```
o-: `optimizationName == "globalValueNumber"` is more idiomatic

rir/src/utils/configurations.h
```diff
@@ -1,39 +1,40 @@
...
-    
+
     pir::PirTranslator* translator;
     short order;
 
-    bool operator < (const Optimization& anotherOptimization) const {
+    bool operator<(const Optimization& anotherOptimization) const {
         if (order == anotherOptimization.order) {
-            return anotherOptimization.translator->getName().compare("cleanup");
+            return anotherOptimization.translator->getName() == "cleanup";
```
o-: but this is still not an order. 
`!((1,"a") < (1, "b"))` and `!((1,"b") < (1, "a"))`, but `(1,"a") != (1, "b")`

charig: Indeed. 
My idea was that we would not set up in the config file the same order for different opt but I agree that it would be better to provide a proper order.

rir/src/utils/configurations.h
```diff
@@ -16,7 +16,14 @@ struct Optimization {
 
     bool operator<(const Optimization& anotherOptimization) const {
         if (order == anotherOptimization.order) {
-            return anotherOptimization.translator->getName() == "cleanup";
+            if (anotherOptimization.translator->getName() == "cleanup") {
+                return true;
```
o-: I believe it is still wrong. For example now `(1, "cleanup") < (1, "cleanup")` :)

sorry, but the set really relies on on the order being an order. otherwise it will start overriding elements, those are really nasty bugs.

best thing here is to rely on somebody else being more clever. (ie. delegate to the stdlib). For example if you look at [tuple](http://en.cppreference.com/w/cpp/utility/tuple/operator_cmp) then you see that the generic implementation is actually kind of tricky:

```
(bool)(std::get<0>(lhs) < std::get<0>(rhs)) || (!(bool)(std::get<0>(rhs) < std::get<0>(lhs)) && lhstail < rhstail), where lhstail is lhs without its first element, and rhstail is rhs without its first element. For two empty tuples, 
```

so to use the tuple, try this:
```
auto T = [](const Optimization& o) {
  return std::tuple<unsigned, unsigned, string>(o.order, o.getName() == "cleanup" ? 0 : 1, o.getName());
};
return T(*this) < T(other);
```

charig: So i guess just adding this condition should make it:
```
if (anotherOptimization.translator->getName() == "cleanup" && this->translator->getName() == "cleanup") return false;
```

But I followed your suggestion and push a new version based on tuple comparison.

PR #136, opened 2018-05-24
Revert "Add support to rir so that the optimizations are configurable…

o-:
… via an INI file (#134)"

This reverts commit 80c23ff6d7a7f84cf65752b5f5fad25035a3f0d4.

optimizations set is not sorted. order appears more or less random.
this breaks the pre-push-hook on my machine, therefore reverting.


o-:
@charig tbr

PR #140, opened 2018-05-24
Fix segv

o-:
new compiler, new problems.

this fixes some issues encountered with gcc 8.1.1.

kind of surprising that we didn't see them before...

(The change in Sexp.h is just to avoid some warnings)


o-:
tbr

Issue #139, opened 2018-05-24
`for` loop


JanJecmen:
This is broken still: `function(x) for (i in 1:x) cat(i, "\n")`

Somewhere along the way, the `phi` for the sequence index is lost. Consequently, a constant is always incremented which leads to an infinite loop.

Original RIR:
```
   guard_fun_  for == 0x7fa38e0351b8
   guard_fun_  : == 0x7fa38e03ae00
   push_  5 # [1] 1
   ldvar_  6 # x
   colon_
   set_shared_
   for_seq_size_
   push_  7 # [1] 0
1:
   inc_
   dup2_
   lt_
   brtrue_  0
   pull_  2
   pull_  1
   extract2_1_
   stvar_  8 # i
   ldfun_  9 # cat
   call_ [ 0 1 ]
        -> cat(i, "\n")
           Prof : [ 0, <0> ]

   pop_
   br_  1
0:
   pop_
   pop_
   pop_
   push_  12 # NULL
   invisible_
   ret_
------------------------
@0
   ldvar_  8 # i
   ret_
------------------------
@1
   push_  10 # [1] "\n"
   ret_
```

PIR (sans promises):
```
BB0
  val^    %0.0  = LdArg            0
  real$   %0.1  = LdConst          [1] 1
  val     %0.2  = Force            %0.0
  val     %0.3  = Colon            %0.1, %0.2
  int$    %0.4  = ForSeqSize       %0.3
  int$    %0.5  = LdConst          [1] 0
  env     e0.6  = MkEnv            parent=<environment: R_GlobalEnv>, x=%0.0
  goto BB1
BB1
  int$    %1.0  = Inc              %0.5
  lgl     %1.1  = Lt               %0.4, %1.0
  t       %1.2  = AsTest           %1.1
  void            Branch           %1.2 -> BB2 (if true) | BB3 (if false)
BB3
  val$    %3.0  = Extract2_1D      %0.3, %1.0
  void            StVar            i, %3.0, e0.6
  cls     %3.2  = LdFun            cat, e0.6
  prom    %3.3  = MkArg            missing, Prom(0), e0.6
  str$    %3.4  = LdConst          [1] "\n"
  prom    %3.5  = MkArg            %3.4, Prom(1), e0.6
  val^    %3.6  = Call             e0.6, %3.2, %3.3, %3.5                             ; env leak
  goto BB1
BB2
  nil     %2.0  = LdConst          NULL
  void            Return           %2.0
======= Liveness ========
[...]

PR #141, opened 2018-05-24
don't finalize function after compiling inlined eager promise expression

o-:
Pir2Rir::translate would translate + finalize code. But translate() could
be called multiple times, when compiling expressions for eager promises.

This caused the things at the end of translate (like removing
unnecessary phi's) to be run too early.

Fixed the issue by making translation and finalization two separate steps.
Also, translation is now const, meaning it does not alter the rir2pir object
itself, but rather just reads the input code and writes to the output
builder. This should make it harder to introduce a similar bug in the
future.

Also removed the rir_inlined_promise_2_pir, since it is just making the
thing more convoluted, for no good reason.


o-:
fixes #139

charig (approved):

PR #143, opened 2018-05-25
Refactor call & support for calls without envs and eager args

o-:
this heavily refactors all the machinery behind the call bytecodes. The main concept is, that all the arguments related to a call site are put in a stack allocated `CallsiteContext` struct, to avoid super-long argument lists. Also we use the type of those structs to statically advertise, if we have yet to search for a callee (`DispatchContext`) or not, or if we already popped eager args off the stack (the `Final...` variants).

As a new feature this version has the `dispatch` method, that selects slot 1, only if possible. Also we now support eager arguments for slot 1.

It is implemented the following way: `rirEvalCode` takes `SEXP env` and `CallsiteContext* callCtx`, either of them can be `null`.

* If `env` is null, the callee must set `env` before using any `env`y bytecode, like `stvar`. Pir optimized code does so.
* If `callCtx` is null, the callee must not use `ldarg_`. Slot 0 code does so.
* When `ldarg_` accesses an arg, it will query `callCtx->hasEagerArgs` and either create a promise on demand, or load the value with `callCtx->eagerArg(i)`, which reads it from the stack. The caller must pop the arguments *after* the function returns.


rir/src/interpreter/interp.cpp
```diff
@@ -24,6 +22,112 @@ extern Rboolean R_Visible;
...
+                           getFixedupAst(), callee);
+    }
+};
+
+struct FinalDispatchContext : public DispatchContext {
+    FinalCallContext setCallee(SEXP callee) const;
+    DispatchContext fixupAst() const = delete;
+    FinalDispatchContext popArgs(Context*) const = delete;
+    explicit FinalDispatchContext(const DispatchContext& other)
+        : DispatchContext(other) {}
```
JanJecmen, line 105: Maybe add `DispatchContext(const DispatchContext& other) = default;`

JanJecmen: And the same for `CallContext`

rir/src/interpreter/interp.cpp
```diff
@@ -24,6 +22,112 @@ extern Rboolean R_Visible;
...
+struct DispatchContext {
+    R_bcstack_t* eagerArgs;
+    Function* caller;
+    SEXP callerEnv;
+    CallSite* callSite;
+    unsigned nargs;
+    SEXP ast;
+
+    DispatchContext(Code* c, unsigned id, R_bcstack_t* eagerArgs,
+                    unsigned nargs, SEXP callerEnv, Context* ctx)
```
JanJecmen: * `nargs` are also stored in the `CallSite`, could get rid of it in all the parameter lists
* `id` could maybe be called `callSiteId`?

rir/src/interpreter/interp.cpp
```diff
@@ -1442,55 +1529,93 @@ SEXP evalRirCode(Code* c, Context* ctx, EnvironmentProxy* ep) {
...
+            int ttt = R_PPStackTop;
+
             Immediate id = readImmediate();
             advanceImmediate();
             Immediate n = readImmediate();
             advanceImmediate();
-            ostack_push(ctx, doDispatch(c, false, n, id, ep, ctx));
+            DispatchContext call(c, id, n, getenv(), ctx);
+            res = doDispatch(call, ctx);
+            ostack_pop(ctx);
```
JanJecmen: What is this pop? Also, in `insns.h` we have the descriptions wrong - there is no 3rd immediate:
```
/**
 * dispatch_:: similar to call, but receiver is tos and 3rd immediate
 *             is selector
 */
DEF_INSTR(dispatch_, 2, 1, 1, 0)

/**
 * dispatch_stack_:: similar to dispatch, but all args on stack
 */
DEF_INSTR(dispatch_stack_, 2, -1, 1, 0)
```

rir/src/interpreter/interp.cpp
```diff
@@ -1442,55 +1529,93 @@ SEXP evalRirCode(Code* c, Context* ctx, EnvironmentProxy* ep) {
         }
 
         INSTRUCTION(call_) {
+            auto lll = ostack_length(ctx);
+            int ttt = R_PPStackTop;
+
             Immediate id = readImmediate();
             advanceImmediate();
-            Immediate n = readImmediate();
+            Immediate nargs = readImmediate();
```
JanJecmen: Now that I think of it, why do we have `nargs` as immediate? Every call knows its callsite where the number of args is stored too. There is a comment in `CallSite` saying the `nargs` are duplicated there, so why not keep only those? I tried to find a place where the immediate is needed, and as far as I can tell, at every place where it's decoded, the callsite id is also available and thus `cs->nargs`

rir/src/interpreter/interp.cpp
```diff
@@ -1195,6 +1275,11 @@ SEXP evalRirCode(Code* c, Context* ctx, EnvironmentProxy* ep) {
 
     R_Visible = TRUE;
 
+    auto getenv = [&env]() {
```
JanJecmen: Maybe be explicit here: `auto getenv = [&env]() -> SEXP { ... };`
Or `auto getenv = [&]() -> SEXP { ... };` even?
Not sure what's best readable

rir/src/interpreter/interp.cpp
JanJecmen: I think I understand now:

* `call_` has callee on tos, pops it after `doCall`
* `call_stack_` has callee on stack under n args, pops it after `doCall` (which pops n args)
* `static_call_stack_` doesn't have callee on stack at all, only pops args in `doCall`
* `dispatch_` has obj on tos, pops it after `doDispatch`
* `dispatch_stack_` has obj on stack, but it's regarded as the first of the n args, so it doesn't pop it because that already happened in `doDispatch`

We should probably add comments I think.

rir/src/interpreter/interp.cpp
```diff
@@ -475,286 +526,312 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
-            createArgsListStack(ctxt_.caller, cs, ep, false, globalContext());
-        ostack_popn(globalContext(), cs->nargs);
-        // Also patch the node stack top in the context -- arguments need
-        // to be removed in case of non-local return
-        RCNTXT* cntxt = findClosureReturnContext();
-        cntxt->nodestack -= cs->nargs;
-    } else {
-        argslist =
-            createArgsList(ctxt_.caller, call, cs, ep, false, globalContext());
+    CallSite* cs = call.callSite;
```
JanJecmen: Just put `call.callSite` as an arg to `profileCall` to be consistent with `doCall`

rir/src/interpreter/interp.cpp
```diff
@@ -475,286 +526,312 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
-    // closure call. From now, they are also protected during gc
-    // (contexts are considered roots by the collector).
-    RCNTXT* cntxt = findClosureReturnContext();
-    cntxt->promargs = argslist;
-    cntxt->cloenv = env;
+    // ===============================================
+    // Then try S3
+    const char* generic = CHAR(PRINTNAME(selector));
+    SEXP rho1 = Rf_NewEnvironment(R_NilValue, R_NilValue, call.callerEnv);
+    ostack_push(ctx, rho1);
```
JanJecmen: Is this just to protect the env? (I know this is not from this PR, but I noticed now)

rir/src/interpreter/interp.cpp
```diff
@@ -475,286 +526,312 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
-        // the promise args list
-        SET_PRVALUE(CAR(actuals), obj);
+    Function* fun = table->at(slot);
+    fun->registerInvocation();
+
+    SEXP arglist = actuals;
+
+    SEXP env = R_NilValue;
+    if (!arglist && needsEnv) {
+        arglist = createLegacyLazyArgsList(call, ctx);
```
JanJecmen, line 689: I think this should be `PROTECT`ed

rir/src/interpreter/interp.cpp
```diff
@@ -475,286 +526,312 @@ INLINE SEXP fixupAST(SEXP call, Context* ctx, size_t nargs) {
...
+
+    switch (TYPEOF(fcall.callee)) {
+    case SPECIALSXP:
+        return legacySpecialCall(fcall, ctx);
+    case BUILTINSXP:
+        // force all promises in the args list. We created a lazy legacy args
+        // list, since we did not know the callee yet. So now we know and need
+        // to make it eager.
+        for (SEXP a = actuals; a != R_NilValue; a = CDR(a))
+            SETCAR(a, Rf_eval(CAR(a), call.callerEnv));
```
JanJecmen: This `eval` call just forces the promise, not? Why not force it ourselves here?
Or even `Rf_evalList(actuals, call.callerEnv, call.ast, 0);`?

JanJecmen (approved):
👍 this is awesome!

rir/src/interpreter/interp.h
```diff
@@ -5,15 +5,16 @@
...
 #include "interp_data.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-SEXP evalRirCode(Code* c, Context* ctx, rir::EnvironmentProxy* ep);
+struct DispatchContext;
+SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
+                 const DispatchContext* = nullptr);
```
JanJecmen: I would maybe prefer being explicit with the nullptr dispatch context, since one needs to look at the declaration to see it

JanJecmen: I would maybe prefer being explicit with the nullptr dispatch context, since one needs to look at the declaration to see it

mhyee: I think I agree. I count only five calls to `evalRirCode`, of which three use the default argument. So I don't think explicitly passing `nullptr` is that inconvenient.

Either way, please give a name to the `DispatchContext*` argument, e.g. `callCtxt`.

rir/src/interpreter/interp.cpp
```diff
@@ -24,6 +22,112 @@ extern Rboolean R_Visible;
...
+    DispatchContext(Code* c, unsigned id, unsigned nargs, SEXP callerEnv,
+                    Context* ctx)
+        : DispatchContext(c, id, nullptr, nargs, callerEnv, ctx) {}
+
+    DispatchContext(R_bcstack_t* eagerArgs, Function* caller, SEXP callerEnv,
+                    CallSite* callSite, unsigned nargs, SEXP ast)
+        : eagerArgs(eagerArgs), caller(caller), callerEnv(callerEnv),
+          callSite(callSite), nargs(nargs), ast(ast) {}
+
+    bool hasEagerArgs() const { return eagerArgs; }
```
mhyee: I prefer the more explicit `return eagerArgs != nullptr`.

rir/src/interpreter/interp.cpp
```diff
@@ -24,6 +22,112 @@ extern Rboolean R_Visible;
 
 using namespace rir;
 
+struct CallContext;
+struct FinalCallContext;
+struct FinalDispatchContext;
+
+struct DispatchContext {
```
mhyee, line 29: Does it make sense to move some of these struct declarations to the header file?

rir/src/interpreter/interp.cpp
```diff
@@ -24,6 +22,112 @@ extern Rboolean R_Visible;
...
+        return caller->codeAt(callSite->args()[i]);
+    }
+
+    SEXP eagerArg(unsigned i) const {
+        assert(eagerArgs && i < nargs);
+        return ostack_at_cell(eagerArgs + i);
+    }
+
+    SEXP getFixedupAst() const;
+    DispatchContext fixupAst() const {
```
mhyee, line 65: Should this be `virtual`? I'm a bit confused because `CallContext` inherits from `DispatchContext`, but if I do:

```
DispatchContext c = CallContext(...);
c.fixupAst();
```

It'll call `DispatchContext::fixupAst()`.

mhyee (commented):
I don't have much to say, seeing as I'm not too familiar with this part of the implementation. But that will change soon...

rir/src/interpreter/interp.cpp
o-: what's the purpose of that? (since I don't know this idiom)

o-: Ah, sorry I get it, you mean to add `FinalDispatchContext(const FinalDispatchContext& other) = default`, to restore the default that I overrode? I think it's not necessary, as they should in fact not be copied, but passed by reference. Maybe I should rather delete all the copy ctrs.

o-: actually, we do need them. ok, ignore what I said

rir/src/interpreter/interp.cpp
o-: gonna do half of the work here. we can remove the nargs from the immediate arguments in a later PR, since it's kind of out-of-scope

rir/src/interpreter/interp.cpp
o-: no. This is internal to `interp.cpp`. In general we should try not to put things into headers, which are internal to one file.

rir/src/interpreter/interp.cpp
o-: yes, you are right, but it should not be virtual. In my opinion `virtual` should only be used if you actually do have a call site where it is not statically clear, what should be called, which is not the case here. Also, the idea with this struct is that it is basically just a collection of things that would otherwise have to be passed as 5 separate arguments to all those functions. So we are relying on the compiler here, to make passing the struct as efficient, as passing the arguments separately. This will certainly not be the case, if the struct has virtual methods.

JanJecmen: So, your example is slicing the callee member in `CallContext` I believe:
> In C++ programming, object slicing occurs when an object of a subclass type is copied to an object of superclass type: the superclass copy will not have any of the member variables defined in the subclass. (These variables have, in effect, been "sliced off".) More subtly, object slicing can also occur when an object of a subclass type is copied to an object of the same type by the superclass's assignment operator, in which case some of the target object's member variables will retain their original values instead of being copied from the source object.
(https://en.wikipedia.org/wiki/Object_slicing)

But if we use pointers or references, then yes, the base version would be called. The only problem would be if someone passed a `CallContext` into `doDispatch(DispatchContext&,...)`. That doesn't happen now, but enforcing it would need that `virtual` or some other machinery :/

Also maybe related to the `=default` copy ctrs in my comment below.. Originally my idea was just to be more explicit about what we want.

Maybe best would be to say `DispatchContext` copy ctr is default, `CallContext` copy ctr is default, and the final version's ctrs from the non-final versions call those, and the final version's copy ctrs are either default or deleted?

o-: > But if we use pointers or references, then yes, the base version would be called. The only problem would be if someone passed a CallContext into doDispatch(DispatchContext&,...)

even then it is not a problem, since both do the same. You'd just get a FinalDispatchContext, which means you loose access to the callee, but that's it...

mhyee: My concern is that there's too many gotchas here (thanks C++). Even my example of demonstrating a gotcha fell victim to object slicing!

If it's intentionally *not* `virtual`, then I would have liked to see some comment, since the usual pattern is to make it `virtual`. But there's still places where it can go wrong, and I'd prefer it if we could enforce it in the compiler.

If the point is to just pass things around in a struct, would it make more sense to use composition instead of inheritance? Does the code rely on the fact that `CallContext` is a subtype of `DispatchContext`, or can we work around that?

PR #144, opened 2018-05-30
Global value numbering

charig:
The implementation is based on [Rekha Pai's paper](https://dl.acm.org/citation.cfm?id=3034351) which presents a polynomial time algorithm for detecting global redundancies.

The main difficulty of the problem is to find redundancies among join point. For instance:
```
bb1:                                           bb2:
 x1 = ...                                        x2 = ...
 y1= x1 + 1                                  y2 = x2 + 1

                  bb3:
                   x3 = phi(x1,x2)
                   w3 = x3 + 1
```

In this case w3 is redundant but its redundancy depends on both previous bbs.

Anyway, this first wip commit still does not implement the most important case of the transfer function to support this kind of redundancies. Till now it only detects and removes redundancies at the same bb level.

### DIFF
--- a/rir/src/compiler/analysis/abstract_value.h
+++ b/rir/src/compiler/analysis/abstract_value.h
@@ -43,7 +43,7 @@ namespace pir {
 /*
  * Captures an abstract PIR value.
  *
- * Vals is the set of potential candidates. If we don't can't tell what the
+ * Vals is the set of potential candidates. If we can't tell what the
  * possible values are, then we set "unknown" (ie. we taint the value). This is
  * the top element of our lattice.
  *
@@ -210,13 +210,6 @@ struct AbstractREnvironment {
         return changed;
     }
 };
-
-/*
- * AbstractEnvironmentSet is an abstract domain that deals with multiple
- * environments at the same time. This is necessary for inter-procedural
- * analysis, or analyzing a function with multiple environments.
- *
- */
 
 struct AbstractLoad {
     Value* env;
--- a/rir/src/compiler/analysis/abstract_value.h
+++ b/rir/src/compiler/analysis/abstract_value.h
@@ -40,6 +40,16 @@ struct hash<rir::pir::ValOrig> {
 namespace rir {
 namespace pir {
 
+class AbstractValue {
+    virtual bool merge(const AbstractValue& otherValue) = 0;            
+};
+
+template <class AbsValue>
+class AbstractState : public std::unordered_map<Value*, AbsValue> {
+    virtual bool merge(const AbstractState& otherState) = 0;            
+};         
+
+
 /*
  * Captures an abstract PIR value.
  *
--- a/rir/src/compiler/analysis/generic_static_analysis.h
+++ b/rir/src/compiler/analysis/generic_static_analysis.h
@@ -45,16 +45,6 @@ enum class PositioningStyle { BeforeInstruction, AfterInstruction };
  * provided by the subclass that specializes StaticAnalysis.
  */
 
-class AbstractValue {
-    virtual bool merge(const AbstractValue& otherValue) = 0;            
-};
-
-template <class AbsValue>
-class AbstractState : public std::unordered_map<Value*, AbsValue> {
-    virtual bool merge(const AbstractState& otherState) = 0;            
-};         
-
-
 template <class State>
 class StaticAnalysis {
   private:
--- a/rir/src/compiler/analysis/generic_static_analysis.h
+++ b/rir/src/compiler/analysis/generic_static_analysis.h
@@ -32,20 +32,39 @@ enum class PositioningStyle { BeforeInstruction, AfterInstruction };
  * "mergepoint" and then "apply" is used to seek to the desired instruction
  * pointer (see "collect" for an example).
  *
- * AbstractState basically has to have a merge function.
+ * The lattice is indirectly defined by the merge function defined within a
+ * subclass of AbstractValue. It is up to the client to ensure
+ * that the merge function conforms a lattice.
+ * 
+ * AbstractState basically has to have a merge function. Ideally, it fits the
+ * AbstractState abstract class defined below and the client can subclass it. 
+ * For other cases, ad-hoc AbstractStates can be defined as long as they implement 
+ * the merge function.  
+ * 
  * Anything else depends on the requirements of the apply function, which is
  * provided by the subclass that specializes StaticAnalysis.
  */
-template <class AbstractState>
+
+class AbstractValue {
+    virtual bool merge(const AbstractValue& otherValue) = 0;            
+};
+
+template <class AbsValue>
+class AbstractState : public std::unordered_map<Value*, AbsValue> {
+    virtual bool merge(const AbstractState& otherState) = 0;            
+};         
+
+
+template <class State>
 class StaticAnalysis {
   private:
-    std::vector<std::vector<AbstractState>> mergepoint;
-    virtual void apply(AbstractState&, Instruction*) const = 0;
-    AbstractState exitpoint;
+    std::vector<std::vector<State>> mergepoint;
+    virtual void apply(State&, Instruction*) const = 0;
+    State exitpoint;
     bool done = false;
 
   protected:
-    const std::vector<std::vector<AbstractState>>& getMergepoints() {
+    const std::vector<std::vector<State>>& getMergepoints() {
         return mergepoint;
     }
 
@@ -56,25 +75,25 @@ class StaticAnalysis {
         mergepoint.resize(cls->maxBBId + 1);
         mergepoint[entry->id].resize(1);
     }
-    StaticAnalysis(Closure* cls, const AbstractState& initialState)
+    StaticAnalysis(Closure* cls, const State& initialState)
         : entry(cls->entry) {
         mergepoint.resize(cls->maxBBId + 1);
         mergepoint[entry->id].push_back(initialState);
     }
 
-    const AbstractState& result() {
+    const State& result() {
         assert(done);
 
         return exitpoint;
     }
 
     template <PositioningStyle POS>
-    const AbstractState& at(Instruction* i) {
+    const State& at(Instruction* i) {
         assert(done);
 
         BB* bb = i->bb();
         size_t segment = 0;
-        AbstractState state = mergepoint[bb->id][segment];
+        State state = mergepoint[bb->id][segment];
         for (auto j : *bb) {
             if (POS == PositioningStyle::BeforeInstruction && i == j)
                 return state;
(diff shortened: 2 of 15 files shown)


rir/src/compiler/analysis/abstract_value.h
```diff
@@ -211,13 +211,6 @@ struct AbstractREnvironment {
     }
 };
 
-/*
```
o-, line 214: I think this is just in the wrong place, the class is further down

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -32,20 +32,39 @@ enum class PositioningStyle { BeforeInstruction, AfterInstruction };
...
+ * AbstractState abstract class defined below and the client can subclass it. 
+ * For other cases, ad-hoc AbstractStates can be defined as long as they implement 
+ * the merge function.  
+ * 
  * Anything else depends on the requirements of the apply function, which is
  * provided by the subclass that specializes StaticAnalysis.
  */
-template <class AbstractState>
+
+class AbstractValue {
```
o-: can we put this into abstract_value.h with the others?

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+    LAnd* land = LAnd::Cast(instruction);
+    LOr* lor = LOr::Cast(instruction);
+
+    Phi* phi = Phi::Cast(instruction);
+    if (phi) {
+        /* It is a phi node. Create a phi eq-class if it is the first time
+         * Note that the original algorithm does not process phis but creates
+         * the phi equivalance classes at join points. For that it requires to
+         * be in CSSA. We workaround this by creating the phi equivalaence here.
+         * Check if we mantain semantics.
```
o-, line 42: we could also run after converting to cssa?

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+                klass.phis = valuePhiFunc;
+            } else {
+                GREquivalenceClass klass;
+                klass.expressions.push_back(assignment);
+                klass.phis = valuePhiFunc;
+                partition.insert({instruction, klass});
+            }
+        }
+    
+        /*if (add || gte || lte || mul || division || idiv || mod || colon ||
```
o-: ??

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+#include "../pir/pir.h"
+#include "generic_static_analysis.h"
+#include <algorithm>
+
+namespace rir {
+namespace pir {
+
+/*
+ * Polynomial Time Global Value Numbering analysis based on:
+ * https://www.sciencedirect.com/science/article/pii/S1477842415300622
```
o-, line 13: ugh, paywalled

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    PIRAssignment assignment;
+    if (loadConst) {
+        assignment.setConstant(&loadConst->c);
+    } else if (loadArg) {
+        assignment.setIndex(loadArg->id);
+    } else {
+        ValueExpression::Expression expression;
+        if (loadFun || loadVar || superLoadVar) {
+            expression.operand1 = instruction->arg(0).first;
```
o-: `arg(0)` is just the environment. the variable name is `loadVar->varName`. But, probably ldvar should be treated like a function call, since those are really the local variables the `scope_resolution` pass was not able to figure out. So we don't really know what we get back from the instruction.

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+    SEXP* constant;
+    size_t argIndex;
+    ValueExpression::Expression* expression;
+};
+
+struct PIRAssignment {
+    GRPIRAssignment argument;
+    unsigned char type;
+    
+    void setConstant(SEXP* constant) {
```
o-: this should be `SEXP`. It's a bit confusing. R has a `typedef SEXPREC* SEXP`, so `SEXP` is already a pointer

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+        } else if (inc || isObject) {
+            expression.operand1 = instruction->arg(0).first;
+            expression.operation = &instruction->tag;
+        } else if (add || gte || lte || mul || division || idiv || mod || colon || 
+                power || sub || gt || lt || neq || eq || land || lor) {
+            expression.operand1 = instruction->arg(0).first;
+            expression.operand2 = instruction->arg(1).first;
+            expression.operation = &instruction->tag;
```
o-: As a rule of thumb, using `&` in C++ is almost always wrong ;)

So here, you are passing the *address* of the location where the tag is stored to the expression. Let's assume you have two Instructions `A` and `B`. Your memory might look like this:

```
Address   |  Variable    |  Value
0x123     |  A::tag      |  Add
0x124     |  ...         |  ....
0x125     |  B::tag      |  Add
```

Now I assume you wanted to have both instructions in the same class, since they are both `Add`. But what happens here is that you store the address `0x123` and `0x125`. Then later on you compare `operation == expression.operation`, which compares their pointer *address* values. You would have to compare using `*operation` to get the content of the memory word holding the tag value.

*But* this is not the way to go here, since Tag is just a small enum, you should just store `Tag`, the actual tag value, instead of `Tag*`. so tldr is, `Expression::operation` should have type `Tag`.

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+*/
+struct ValueExpression {
+    typedef ValueExpression Expression;
+    ValueExpression() : operand1(), operand2(), operation(nullptr) {};
+    ValueExpression(Value* value) : operand1(value), operand2(), operation(nullptr) {};
+    ValueExpression(Value* value, Value* anotherValue, Tag* operation) : 
+        operand1(value), operand2(anotherValue), operation(operation) {};
+    Value* operand1;
+    Value* operand2;
+    Tag* operation;
```
o-: use Tag, see above

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (phi) {
+        /* It is a phi node. Create a phi eq-class if it is the first time
+         * Note that the original algorithm does not process phis but creates
+         * the phi equivalance classes at join points. For that it requires to
+         * be in CSSA. We workaround this by creating the phi equivalaence here.
+         * Check if we mantain semantics. 
+        */
+        if (!partition.count(instruction)) {
+            GREquivalenceClass klass;
+            klass.phis = new ValuePhiFunction(phi->arg(0).val(), phi->arg(1).val());
```
o-: phi is vararg, it can have any number of arguments

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+    Value* rightPath;
+    //BB* joinPoint;
+    bool operator==(const ValuePhiFunction& anotherPF) {
+        return leftPath == anotherPF.leftPath &&
+                rightPath == anotherPF.rightPath;
+    }        
+};
+
+union GRPIRAssignment {
+    SEXP* constant;
```
o-: `SEXP`

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+    Value* leftPath;
+    Value* rightPath;
+    //BB* joinPoint;
+    bool operator==(const ValuePhiFunction& anotherPF) {
+        return leftPath == anotherPF.leftPath &&
+                rightPath == anotherPF.rightPath;
+    }        
+};
+
+union GRPIRAssignment {
```
o-: make this an inner class of `PIRAssignment`?

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+    SEXP* constant;
+    size_t argIndex;
+    ValueExpression::Expression* expression;
```
o-: just `Expression`, no pointer? See bellow comment on passing pointer to a local variable at `setExpression(&expression)`

rir/src/compiler/analysis/global_redundancies.h
```diff
@@ -0,0 +1,214 @@
...
+    }
+    
+};*/
+
+/*
+ * An abstraction of a set of equivalent expressions.
+ * For more information see page 170. 
+*/
+struct ValueExpression {
+    typedef ValueExpression Expression;
```
o-: why?

rir/src/compiler/analysis/global_redundancies.cpp
```diff
@@ -0,0 +1,122 @@
...
+        } else if (inc || isObject) {
+            expression.operand1 = instruction->arg(0).first;
+            expression.operation = &instruction->tag;
+        } else if (add || gte || lte || mul || division || idiv || mod || colon || 
+                power || sub || gt || lt || neq || eq || land || lor) {
+            expression.operand1 = instruction->arg(0).first;
+            expression.operand2 = instruction->arg(1).first;
+            expression.operation = &instruction->tag;            
+        }
+        assignment.setExpression(&expression);
```
o-: this won't work. You pass a pointer to the local variable `expression`. In C++ `Expression expression` creates a local variable on the stack. When the function returns, the stack frame vanishes and the local variable with it. Passing a pointer of a local variable and then keeping it alive longer than the function, is undefined. In this case it probably just results in a dangling pointer.

The way to go is probably to always pass Expressions by value. Since they should be really small structs (maybe 3 words big), I think it's ok.

rir/src/compiler/opt/global_value_numbering.h
```diff
@@ -0,0 +1,27 @@
+#ifndef PIR_GVN_H
+#define PIR_GVN_H
+
+#include "../translations/pir_translator.h"
+
+namespace rir {
+namespace pir {
+
+/*
+ * This pass searches for dominating force instructions.
```
o-: uhm, c/p comment from another pass

o- (commented):
first round of comments. there are some issues with pointers vs. values in this pr

rir/src/compiler/analysis/global_redundancies.cpp
charig: The paper describes somewhere that to simplify the detection of equivalence it assumes CSSA and this some tricky thing at join points. I think the semantics could be the same without moving to CSSA and processing phi nodes in the trasfer function. But yes, this should be still checked, that is why the comment is there.

rir/src/compiler/analysis/global_redundancies.cpp
charig: Indeed. In the first version, I think it is ok to give support just to two value phis that are the most common. I added a comment to the code.

PR #145, opened 2018-05-30
bugs bugs bugs

o-:
letz squash all the :bug: :beetle: 

77f6ab9

* LdArgSuper did not actually load a value
* Is::tag was hiding Value::tag
* Seq required number().scalar(), but there was no support for casting to that, for now let's require just value.
* The "stack" passed to Deopt was actually a totally wrong pointer
* dominator analysis was actually totally broken. In the case of  `a -> b -> c | a -> c`, the merge block `c` did not know it was dominated by `a`.
* When looking up call sites (the structs that describe the metadata of a call site) inside promises, the pir compiler was using the function pointer instead of the promise. Ie. `callSite(function->body())` instead of `callSite(code)`. This would cause random memory from the function being used as the `CallSite` pointer.
* Phi's where all the inputs are the same, except for a self-reference, should be removed. this removes stupid phis in nested loops.

c36ec37

* out of range error when merging abstract envs with mismatching local vars.
* verifier used cfg of main function when verifying promises
* cast insertion pass was broken, when multiple casts had to be inserted.

4f83f12

* sometimes abstract environments would become `nullptr`, even though the abstract env hierarchy assumed that all env chains end in `UnknownParent`.
* Fix merging two abstract envs, when the parent env is not the same
* fixed super lookup in abstract environments
* fixed replacing StVarSuper with StVar (hint: only possible if we find the variable to assign to)
* fixed updating phi input BB pointer, when deleting some bbs in cleanup phase
* fix cleanup creating non edge-split graphs

6b8e1bb

* fix BB renumbering pass (again). again we used the id marker to renumber ids, which resulted in id's beging assigned twice sometimes
* fixed rir_2_pir not creating edge-split graph sometimes
* fixed constant builtins and specials not being typed as functions in pir


o-:
so I am gonna merge this, tbr. this is all just bug fixes, that we really want to have :)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -275,26 +282,48 @@ void Rir2Pir::finalize(Value* ret, Builder& insert) {
...
+                    if (p == ret)
+                        ret = p->arg(0).val();
+                    p->replaceUsesWith(p->arg(0).val());
+                    it = bb->remove(it);
+                    changed = true;
+                    continue;
+                }
+                // Phi where all inputs are the same value (except the phi
+                // itself), then we can remove it. 
+                Value* allTheSame = p->arg(0).val();
```
JanJecmen, line 307: This took me a while but +1 :)

PR #146, opened 2018-05-30
Fail graciously, when pir compilation fails 

o-:
this is based on top of #143


o-:
ready to review now

rir/src/api.cpp
```diff
@@ -47,7 +50,27 @@ REXPORT SEXP rir_compile(SEXP what, SEXP env = NULL) {
             Rf_error("closure already compiled");
 
         SEXP result = Compiler::compileClosure(what);
+        PROTECT(result);
+
         Rf_copyMostAttrib(what, result);
+
+#ifdef SLOWASSERT
```
JanJecmen: Why `SLOWASSERT`? Or is it just for now to enable it in debug mode?

JanJecmen: And isn't `SLOWASSERT` always defined? It's either an assert or an empty block, but still defined. Or do function-like macros behave differently?

rir/src/compiler/pir/bb.cpp
```diff
@@ -26,6 +28,13 @@ BB::~BB() {
         delete i;
 }
 
+Instruction* BB::last() {
+    if (!instrs.size())
```
JanJecmen: Maybe `instrs.empty()` is more readable?

rir/src/compiler/pir/module.cpp
```diff
@@ -23,6 +23,20 @@ void Module::printEachVersion(std::ostream& out) {
     }
 }
 
+void Module::getAndCreateIfMissing(rir::Function* f, const std::vector<SEXP>& a,
+                                   Env* env, MaybeVal get, MaybeCreate create) {
+    if (functions.count(f))
+        return get(functions.at(f).current());
```
JanJecmen: I don't understand this. Or, at least, it looks weird. `MaybeVal` is `void` and then we return this from a `void` function?

rir/src/compiler/pir/type.cpp
```diff
@@ -30,8 +30,10 @@ PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
     case PROMSXP:
         t_.r.set(RType::prom);
         break;
-    case LANGSXP:
+    case EXPRSXP:
```
JanJecmen, line 33: Is `EXPRSXP` always code? I found here that it can be just a symbol or a vector: https://github.com/hadley/r-internals/blob/master/vectors.md

rir/src/compiler/pir/type.h
```diff
@@ -79,7 +80,6 @@ enum class TypeFlags : uint8_t {
 
     lazy,
     missing,
-    obj,
```
JanJecmen, line 82: What happens to objects now?

rir/src/ir/cleanup.h
```diff
@@ -7,6 +7,12 @@ namespace rir {
 
 class BCCleanup : public InstructionDispatcher::Receiver {
   public:
+    static void apply(CodeEditor& code) {
```
JanJecmen: Used to run `steam`-times. Now only once

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -135,6 +137,12 @@ Value* Rir2Pir::translate(rir::Code* srcCode, Builder& insert) const {
                 insert(new Branch(v));
                 break;
             }
+            case Opcode::beginloop_:
+                if (compiler.isVerbose())
+                    std::cerr
+                        << "Cannot compile Function. Unsupported return bc\n";
```
JanJecmen, line 142: I don't understand, why is `beginloop_` considered a jump?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -236,26 +247,42 @@ Value* Rir2Pir::translate(rir::Code* srcCode, Builder& insert) const {
...
+                },
+                []() {});
         });
 
         if (!matched) {
             int size = state.stack_size();
-            state.runCurrentBC(*this, insert);
+            if (!state.tryRunCurrentBC(*this, insert)) {
+                if (compiler.isVerbose()) {
+                    std::cout << "Abort p2r due to unsupported bc ";
```
JanJecmen: here it's `r2p`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -236,26 +247,42 @@ Value* Rir2Pir::translate(rir::Code* srcCode, Builder& insert) const {
...
-
-            matched = true;
-            state.setPC(next);
+                    matched = true;
+                    state.setPC(next);
+                },
+                []() {});
         });
 
         if (!matched) {
```
JanJecmen: Maybe call this differently? It is matched at the point where ifFunctionLiteral advances the pc, but it can still fail when compiling.

rir/src/compiler/translations/rir_2_pir/stack_machine.h
```diff
@@ -30,7 +30,10 @@ class StackMachine {
 
     typedef std::pair<BB*, Value*> ReturnSite;
     typedef std::function<void(ReturnSite)> ReturnMaybe;
```
JanJecmen: Not used anywhere

rir/src/compiler/translations/rir_2_pir/stack_machine.h
```diff
@@ -30,7 +30,10 @@ class StackMachine {
...
-    void runCurrentBC(const Rir2Pir& cmp, Builder&);
+
+    typedef std::function<void()> Maybe;
```
JanJecmen: Also not used. Or am I missing something?
And a thought, what if we moved all these typedefs to one common place? But on the other hand, it's nice to see them where they are used, even though they are duplicated sometimes... So I don't know..
And a question on `typedef` vs. `using`: which one do you prefer? :)

rir/src/compiler/translations/rir_compiler.h
```diff
@@ -17,7 +17,13 @@ class RirCompiler {
     RirCompiler(const RirCompiler&) = delete;
     void operator=(const RirCompiler&) = delete;
 
-    virtual Closure* compileClosure(SEXP) = 0;
+    typedef std::function<void()> Maybe;
+    typedef std::function<void(Closure*)> MaybeVal;
```
JanJecmen: If I understand correctly the naming convention, this should be called `MaybeClosure`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
```diff
@@ -15,14 +15,38 @@ class Rir2Pir {
...
+                   Maybe fail) const;
+
+    void translate(rir::Code* srcCode, Builder& insert,
+                   MaybeVal success) const {
+        translate(srcCode, insert, success, []() {});
+    }
+
+    template <typename T>
+    T translate(rir::Code* srcCode, Builder& insert,
+                std::function<T(Value*)> success,
```
JanJecmen: Just wondering, can you have templated typedefs? :D

JanJecmen: so looks like you can do templates with `using`:

```
    // ...
    template <typename T>
    using MaybeVal = std::function<T(Value*)>;

    template <typename T>
    T translate(rir::Code* srcCode, Builder& insert,
                MaybeVal<T> success, // ...
```

Just FYI, not saying it's more readable :)

rir/src/api.cpp
```diff
@@ -47,7 +50,27 @@ REXPORT SEXP rir_compile(SEXP what, SEXP env = NULL) {
...
+        pir::Rir2PirCompiler cmp(m.get());
+        // cmp.setVerbose(true);
+        cmp.compileClosure(result,
+                           [&](pir::Closure* c) {
+                               cmp.optimizeModule();
+                               // TODO, next step: enable back-compilation
+                               // pir::Pir2RirCompiler p2r;
+                               // p2r.compile(c, closure);
+                           },
+                           [&]() {});
```
JanJecmen: `[]() {}`?
(only angle brackets are now missing, can we make a template out of this?)

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -40,33 +44,43 @@ Closure* Rir2PirCompiler::compileClosure(SEXP closure) {
...
-Closure* Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
-                                         const std::vector<SEXP>& args,
-                                         Env* closureEnv) {
-    return module->getOrCreate(
-        srcFunction, args, closureEnv, [&](Closure* pirFunction) {
+void Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
+                                     const std::vector<SEXP>& args,
+                                     Env* closureEnv, MaybeVal success,
+                                     Maybe fail) {
+    module->getAndCreateIfMissing(
```
JanJecmen: This looks weird, the `get` part. Maybe have a separate `get` that would just take `Function*` and then `createIfMissing` that would do this?

JanJecmen: Ok I see now, the `get` calls `success` on the result. It took me quite a while to decipher it :)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -74,7 +74,9 @@ void recoverCFG(rir::Code* srcCode, rir::Function* srcFunction,
 namespace rir {
 namespace pir {
 
-Value* Rir2Pir::translate(rir::Code* srcCode, Builder& insert) const {
+void Rir2Pir::translate(rir::Code* srcCode, Builder& insert,
+                        std::function<void(Value*)> success,
+                        std::function<void()> fail) const {
```
JanJecmen: These are `MaybeVal success` and `Maybe fail`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -236,26 +247,42 @@ Value* Rir2Pir::translate(rir::Code* srcCode, Builder& insert) const {
...
 
-            state.push(
-                insert(new MkFunCls(innerF, insert.env, fmls, code, src)));
-
-            matched = true;
-            state.setPC(next);
+                    matched = true;
+                    state.setPC(next);
+                },
+                []() {});
```
JanJecmen: Should this be empty? Don't we want to propagate up that the closure couldn't be compiled?
This just skips the `push; push; push; close` sequence and silently does nothing and continues, if I understand it?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
```diff
@@ -15,14 +15,38 @@ class Rir2Pir {
...
+                                   return true;
+                               },
+                               []() { return false; });
     }
 
-    ~Rir2Pir() { assert(finalized); }
-
   private:
-    Value* translate(rir::Code* srcCode, Builder& insert) const;
+    void translate(rir::Code* srcCode, Builder& insert, MaybeVal success,
```
JanJecmen: Got lost in all the `translate`s for a bit... Can we rename the templated version maybe?

rir/src/compiler/util/cfg.cpp
```diff
@@ -131,21 +64,76 @@ void computeDominanceGraph(DominanceGraph* cfg, BB* start) {
...
+                transitivePredecessors[next->id].push_back(bb);
+            }
+        };
+        apply(bb->next0);
+        apply(bb->next1);
+        if (!bb->next0 && !bb->next1)
+            exits_.push_back(bb);
+    });
+
+    std::function<void(BB*, BB*)> complete = [&](BB* bb, BB* pre1) {
```
JanJecmen, line 54: `auto complete = ...`?

rir/src/compiler/util/cfg.cpp
```diff
@@ -131,21 +64,76 @@ void computeDominanceGraph(DominanceGraph* cfg, BB* start) {
...
 }
 }
 
 namespace rir {
 namespace pir {
 
-CFG::CFG(BB* start) { computeCfg(this, start); }
+CFG::CFG(Code* start) {
+    predecessors_.resize(start->nextBBId);
+    transitivePredecessors.resize(start->nextBBId);
```
JanJecmen: Could the `.resize` calls be just moved to the vector's respective constructors?
`CFG::CFG(Code* start) : predecessors_(start->nextBBId), ... { ... }`

rir/src/compiler/util/cfg.cpp
```diff
@@ -94,20 +32,15 @@ class BBInterSet : public std::unordered_set<BB*> {
 
 // Static Analysis computes the set of all dominating bb's, for every bb
 // reachable from start. Runs until none of the sets grow anymore.
-void computeDominanceGraph(DominanceGraph* cfg, BB* start) {
+void computeDominanceGraph(DominanceGraph* cfg, Code* start) {
```
JanJecmen: To be consistent with `CFG`, maybe put this into the `DominanceGraph` constructor? And also make `dominating` private, since it's queried with methods

rir/src/compiler/util/cfg.h
```diff
@@ -9,14 +9,20 @@ namespace rir {
 namespace pir {
 
 class CFG {
-    typedef std::unordered_set<BB*> BBList;
+    typedef std::vector<BB*> BBList;
```
JanJecmen, line 12: Is this safe? I think it should be, but I remember that I had a bug where indexing with a bb id failed because the index was out of bounds. At least I believe it was because of that.. I attempted to fix it by renumbering promises in the cleanup phase. But now it looks like all Codes always have ids starting from 0 and I can't locate when this changed, so I am wondering if I was wrong before?

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
```diff
@@ -138,47 +140,56 @@ void StackMachine::runCurrentBC(const Rir2Pir& rir2pir, Builder& insert) {
...
+                                                      cs->call, monomorphic));
+
+                    BB* cont = insert.createBB();
+                    fallback->next0 = cont;
+                    asExpected->next0 = cont;
+                    insert.bb = cont;
+                    push(insert(new Phi({r1, r2}, {fallback, asExpected})));
+                },
+                [&]() {
+                    push(insert(new Call(insert.env, pop(), args, cs->call)));
```
JanJecmen: Could do something like
```
auto insertGenericCall = [&]() { push(insert(new Call(insert.env, pop(), args, cs->call))); };
if (monomorphic && ...) {
    rir2pir.compiler.compileClosure(
        monomorphic,
        ...,
        insertGenericCall
    );
} else {
    insertGenericCall();
}
```
but maybe it's too much...

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
```diff
@@ -413,11 +434,9 @@ void StackMachine::runCurrentBC(const Rir2Pir& rir2pir, Builder& insert) {
     case Opcode::endcontext_:
     case Opcode::ldddvar_:
     case Opcode::int3_:
-        std::cerr << "Cannot compile Function. Unsupported bc\n";
-        bc.print();
-        assert(false);
-        break;
+        return false;
```
JanJecmen, line 438: Maybe move the currently unused ones and the ones from pir here as well?

JanJecmen (approved):
I like this... Looks a bit like CPS, only we have to return from the callbacks :)
But on the flip side, the parts where the callbacks are passed around take some brain stretching to understand

o-:
thanks for the thorough review. I agree with your assessment of the callback style. I wanted to avoid using exceptions (because of setjmp/longjmp) and alternatively returning optional values seemed more intrusive. Can you point me to the specific cases where you found it annoying to read?

rir/src/compiler/pir/type.h
o-: well, basically, needs more thinking

JanJecmen:
Mainly the part with `Rir2PirCompiler::compileClosure` calling `module->getAndCreateIfMissing`
And then the `translate` template and overloads calling each other, but that's not related to the callbacks :)

But overall I quite like the style

rir/src/compiler/pir/module.cpp
o-: shorthand for `get(...); return;` ;)

rir/src/compiler/pir/type.cpp
o-: I think so. a symbol or vector can also be an ast. this is what makes the whole R ast quite confusing. E.g. `1` is the number one, but `Promise(1, ...)` or `Expresxp(1, ...)` is the expression that produces `1`.

rir/src/compiler/util/cfg.cpp
o-: no, because it is recursive

rir/src/compiler/util/cfg.h
o-: yeah, I fixed this stuff in #145

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
o-: I think it's still ok to fail hard on the ones we do not expect to see

rir/src/compiler/translations/rir_2_pir/stack_machine.h
mhyee: What's our style guide / consensus on `typedef` vs `using`?

Because of old habits, I tend to use `typedef` more often, but `using` lets you use templates. Also, something like

```
using foobar = void (*)(double);
```

is a bit easier to read than

```
typedef void (*foobar)(double);
```

See also https://stackoverflow.com/a/24916697

PR #148, opened 2018-06-06
small tweaks

JanJecmen:
noticed during reviewing...
Also tried to improve the descriptions of our rir call instructions

### DIFF
--- a/rir/src/compiler/pir/closure.cpp
+++ b/rir/src/compiler/pir/closure.cpp
@@ -27,8 +27,7 @@ Promise* Closure::createProm() {
 
 Closure::~Closure() {
     for (auto p : promises)
-        if (p)
-            delete p;
+        delete p;
     for (auto p : defaultArgs)
         delete p;
 }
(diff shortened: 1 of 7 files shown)


rir/src/compiler/pir/closure.cpp
```diff
@@ -27,8 +27,7 @@ Promise* Closure::createProm() {
 
 Closure::~Closure() {
     for (auto p : promises)
-        if (p)
-            delete p;
+        delete p;
```
o-, line 30: I think I added this on purpose. If a promise get's unused, then the cleanup phase can delete it. This just sets it to null, otherwise all the other promises would need to be renumbered (since we address them by index into this array).

JanJecmen: Ok but just peepholing this piece of code, the two versions are equivalent, no?
Because `delete nullptr;` is valid and is a no-op.

o-: oh, is it. didn't know, alright, then I'm ok

o-:
lgtm

rir/src/compiler/pir/closure.cpp
JanJecmen: from cppreference:
> If expression  evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's  implementation-defined), but the default deallocation functions are  guaranteed to do nothing when handed a null pointer. (until C++14)
If expression evaluates to a null pointer value, no destructors are called, and the deallocation function is not called. (since C++14)

PR #149, opened 2018-06-08
Fix popping of eager arguments and simplify passing of callsite info in interpreter

o-:
The previous solution was massively overengineered and also broken.
In general, if we want to support `ldarg_` for eager arguments
(passed on the stack), we need to pop eager arguments *after*
`doCall`. Everything else is just broken.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -22,34 +22,49 @@ extern Rboolean R_Visible;
 
 using namespace rir;
 
-struct CallContext;
-struct FinalCallContext;
-struct FinalDispatchContext;
+struct CallContext {
+    CallContext(Code* c, unsigned callsiteId, SEXP callee,
+                R_bcstack_t* eagerArgs, SEXP callerEnv, Context* ctx)
+        : eagerArgs(eagerArgs), caller(c->function()), callerEnv(callerEnv),
+          callSite(c->callSite(callsiteId)),
+          ast(cp_pool_at(ctx, callSite->call)), callee_(callee) {
+        assert(!callee || TYPEOF(callee) == CLOSXP ||
+               TYPEOF(callee) == SPECIALSXP || TYPEOF(callee) == BUILTINSXP);
+    }
 
-struct DispatchContext {
+    CallContext(Code* c, unsigned id, SEXP callee, SEXP callerEnv, Context* ctx)
+        : CallContext(c, id, callee, nullptr, callerEnv, ctx) {}
+
+    CallContext(Code* c, unsigned id, R_bcstack_t* eagerArgs, SEXP callerEnv,
+                Context* ctx)
+        : CallContext(c, id, nullptr, eagerArgs, callerEnv, ctx) {}
+
+    CallContext(Code* c, unsigned id, SEXP callerEnv, Context* ctx)
+        : CallContext(c, id, nullptr, nullptr, callerEnv, ctx) {}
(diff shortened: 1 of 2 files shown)


o-:
@JanJecmen @mhyee thanks for your persistence. I looked at the calls once more and realized that the whole idea of popping arguments *before* the call, was just utterly broken. If we want to be able to load args from the stack with `ldarg_` then, of course we need to keep them on stack, until after the function returns.

So, this commit removes most of that crap again. We now have just one `CallsiteInfo` struct. No more static diff between dispatch and call (if callee is nullptr it will crash early enough...) and no more static diff between args popped or not (since we keep them on stack the whole time).

rir/src/interpreter/interp.cpp
```diff
@@ -22,34 +22,47 @@ extern Rboolean R_Visible;
 
 using namespace rir;
 
-struct CallContext;
-struct FinalCallContext;
-struct FinalDispatchContext;
+struct CallsiteInfo {
```
JanJecmen: The name is a bit off I feel, since this struct contains a `CallSite*`. Maybe just `CallInfo` or `CallContext` or `DispatchInfo` or similar?
But this is fine too.

rir/src/interpreter/interp.cpp
```diff
@@ -1566,10 +1555,10 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
             advanceImmediate();
             Immediate n = readImmediate();
             advanceImmediate();
-            CallContext call(c, id, R_BCNodeStackTop - n, ostack_at(ctx, n),
-                             getenv(), ctx);
+            CallsiteInfo call(c, id, ostack_at(ctx, n), R_BCNodeStackTop - n,
+                              getenv(), ctx);
             res = doCall(call, ctx);
-            ostack_pop(ctx); // callee
+            ostack_popn(ctx, n + 1);
```
JanJecmen: The comment a few lines above is outdated now

JanJecmen: and the other `*stack_` versions as well

rir/src/interpreter/interp.cpp
```diff
@@ -1546,7 +1535,7 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
             Immediate id = readImmediate();
             advanceImmediate();
             advanceImmediate(); // nargs, TODO: remove
-            CallContext call(c, id, ostack_at(ctx, 0), getenv(), ctx);
+            CallsiteInfo call(c, id, ostack_at(ctx, 0), getenv(), ctx);
```
JanJecmen: this could be `ostack_top(ctx)`

rir/src/interpreter/interp.cpp
```diff
@@ -1566,10 +1555,10 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
             advanceImmediate();
             Immediate n = readImmediate();
             advanceImmediate();
-            CallContext call(c, id, R_BCNodeStackTop - n, ostack_at(ctx, n),
-                             getenv(), ctx);
+            CallsiteInfo call(c, id, ostack_at(ctx, n), R_BCNodeStackTop - n,
```
JanJecmen: and all the `R_BCNodeStackTop - n` could use `ostack_cell_at` :)

JanJecmen (approved):
better and better with each day :)

PR #150, opened 2018-06-15
fix promise inliner

o-:
when we inline a promise, we did not account for the case where the
promise itself is still used. There are two main cases:

1. The promise is used before being forced, e.g.:
```
    prom %1 = mkarg(...)
              call (prom1)
    val %3  = force(%1)
```
here we can't inline.

2. the promise is forced before being used standalone
```
    prom %1 = mkarg(missing, code, env)
    val %2  = force(%1)
              call(%1)
```
here we can create a new promise with the eager value:
```
    # prom %1 = mkarg(missing, code, env)
    val %2  = force(%1)
    prom %3 = mkarg(%2, code, env)
              call(%3)
```

### DIFF
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -45,9 +45,14 @@ using namespace rir::pir;
 struct ForcedAt : public std::unordered_map<Value*, Force*> {
     static Force* ambiguous() { return (Force*)0x22; }
 
-    void evalAt(Value* val, Force* force) {
+    void forcedAt(MkArg* val, Force* force) {
         if (!count(val))
             (*this)[val] = force;
+    }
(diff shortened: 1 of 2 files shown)


charig:
I already reviewed this code in #151. If #151 subsumes this PR let's close it.

Btw, is this case correct?

```
    # prom %1 = mkarg(missing, code, env)
    val %2  = force(%1)
    prom %3 = mkarg(%2, code, env)
              call(%3)
```

What happens if the promise is not side-effects free? The code of the promise is executed in `call(%3)` 
even if the first parameter of mkarg is a concrete value?

o-:
> What happens if the promise is not side-effects free? The code of the promise is executed in call(%3)
even if the first parameter of mkarg is a concrete value?

not anymore

PR #151, opened 2018-06-21
towards enabling pir2rir

o-:
For every piece of code we compile, assert that it can be successfully
compiled to pir and back to rir. The result is thrown away.

For this change to work a couple of bugs are fixed:

1.
promises can be recursive. We can refer to a promise in a promise.
Therefore it is not possible to compile promises in a fixed order.
We need to compile them on demand.

(Recursive dependencies should not exist)

2.
Support for environment arguments was not very robust. This patch
simplifies the code by abstracting the handling of arguments somewhat.

Before every pir instruction, we always load all arguments; after every
pir instruction we always store the result.

The special cases are:

* Some instructions want the argument implicitly. If an instruction
  takes an environment and does not consume it explicitly, then
  there is a preamble that loads the env and issues a setEnv.

* ~~MkArg is currently a bit of a hack, and it still is in this patch:
  either, we have an eager arg, then the result of mkarg is that
  eager arg. Otherwise it does not have a result.
  Then the call is responsible for consuming all the eager args.~~

The implicit environment loading gets very annoying, if the environment
is not the TOS. Therefore this patch changes the order of vararg
instructions, such that (as in the fixed argument length case), the
environment always comes last.

This allows us to evaluate the arguments in the pir-order, when
translating back to rir; and we always find the env to be tos,
therefore to consume it, we just issue a set_env_

3.
Our guard function was using `==` to compare closure values. That
fails for primitives. Add a new `identical_` bytecode that does
pointer equality.

4.
Currently we don't have deoptimization supprt. We'll need to add
it soon. This patch just aborts instead of deopting.

### DIFF
--- a/rir/src/compiler/opt/cleanup.cpp
+++ b/rir/src/compiler/opt/cleanup.cpp
@@ -23,6 +23,60 @@ class TheCleanup {
             while (ip != bb->end()) {
                 Instruction* i = *ip;
                 auto next = ip + 1;
+
+                // Convert lazy calls to eager calls, if all args are eager.
+                {
+                    auto call = Call::Cast(i);
+                    if (call) {
+                        bool allEager = true;
+                        std::vector<Value*> args;
+                        call->eachCallArg([&](Value* v) {
+                            auto arg = MkArg::Cast(v);
+                            if (arg && arg->eagerArg() == Missing::instance())
+                                allEager = false;
+                            if (allEager)
+                                args.push_back(arg->eagerArg());
+                        });
+                        if (allEager) {
+                            auto eagerCall = new EagerCall(
+                                call->env(), call->cls(), args, call->srcIdx);
+                            call->replaceUsesWith(eagerCall);
+                            bb->replace(ip, eagerCall);
+                        }
+                    }
+                }
+
+                {
+                    auto call = StaticCall::Cast(i);
+                    if (call) {
+                        bool allEager = true;
+                        std::vector<Value*> args;
+                        call->eachCallArg([&](Value* v) {
+                            auto arg = MkArg::Cast(v);
+                            if (arg && arg->eagerArg() == Missing::instance())
+                                allEager = false;
+                            if (allEager)
+                                args.push_back(arg->eagerArg());
+                        });
+                        if (allEager) {
+                            auto eagerCall = new StaticEagerCall(
+                                call->env(), call->cls(), args, call->srcIdx,
+                                call->origin());
+                            call->replaceUsesWith(eagerCall);
+                            bb->replace(ip, eagerCall);
+                        }
+                    }
+                }
+
+                ip = next;
+            }
+        });
+
+        Visitor::run(function->entry, [&](BB* bb) {
+            auto ip = bb->begin();
+            while (ip != bb->end()) {
+                Instruction* i = *ip;
+                auto next = ip + 1;
                 Force* force = Force::Cast(i);
                 ChkClosure* chkcls = ChkClosure::Cast(i);
                 ChkMissing* missing = ChkMissing::Cast(i);
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -42,8 +42,11 @@ using namespace rir::pir;
  * branch) or not. Thus we cannot inline it.
  */
 
-struct ForcedAt : public std::unordered_map<Value*, Force*> {
-    static Force* ambiguous() { return (Force*)0x22; }
+struct ForcedBy : public std::unordered_map<Value*, Force*> {
+    static Force* ambiguous() {
+        static Force f(nullptr);
+        return &f;
+    }
 
     void forcedAt(MkArg* val, Force* force) {
         if (!count(val))
@@ -54,7 +57,7 @@ struct ForcedAt : public std::unordered_map<Value*, Force*> {
         if (!count(val))
             (*this)[val] = ambiguous();
     }
-    bool merge(ForcedAt& other) {
+    bool merge(ForcedBy& other) {
         bool changed = false;
         for (auto& e : *this) {
             auto v = e.first;
@@ -88,11 +91,11 @@ static Value* getValue(Force* f) {
     return cur;
 }
 
-class ForceDominanceAnalysis : public StaticAnalysis<ForcedAt> {
+class ForceDominanceAnalysis : public StaticAnalysis<ForcedBy> {
   public:
     ForceDominanceAnalysis(Closure* cls) : StaticAnalysis(cls) {}
 
-    void apply(ForcedAt& d, Instruction* i) const override {
+    void apply(ForcedBy& d, Instruction* i) const override {
         auto f = Force::Cast(i);
         if (f) {
             MkArg* arg = MkArg::Cast(getValue(f));
@@ -114,12 +117,12 @@ class ForceDominanceAnalysisResult {
         ForceDominanceAnalysis analysis(cls);
         analysis();
         analysis.foreach<PositioningStyle::AfterInstruction>(
-            [&](const ForcedAt& p, Instruction* i) {
+            [&](const ForcedBy& p, Instruction* i) {
                 auto f = Force::Cast(i);
                 if (f) {
                     if (p.find(getValue(f)) != p.end()) {
                         auto o = p.at(getValue(f));
-                        if (o != ForcedAt::ambiguous()) {
+                        if (o != ForcedBy::ambiguous()) {
                             if (f != o)
                                 domBy[f] = o;
                             else
@@ -132,7 +135,7 @@ class ForceDominanceAnalysisResult {
     }
 
     bool isSafeToInline(MkArg* a) {
-        return exit.count(a) && exit.at(a) != ForcedAt::ambiguous();
+        return exit.count(a) && exit.at(a) != ForcedBy::ambiguous();
     }
     bool isDominating(Force* f) { return dom.find(f) != dom.end(); }
     void mapToDominator(Force* f, std::function<void(Force* f)> action) {
@@ -141,7 +144,7 @@ class ForceDominanceAnalysisResult {
     }
 
   private:
-    ForcedAt exit;
+    ForcedBy exit;
     std::unordered_map<Force*, Force*> domBy;
     std::unordered_set<Force*> dom;
 };
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -166,7 +166,7 @@ void ForceDominance::apply(Closure* cls) {
                 auto mkarg = MkArg::Cast(getValue(f));
                 if (mkarg) {
                     if (analysis.isDominating(f)) {
-                        Value* strict = mkarg->arg<0>().val();
+                        Value* strict = mkarg->eagerArg();
                         if (strict != Missing::instance()) {
                             f->replaceUsesWith(strict);
                             next = bb->remove(ip);
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -45,9 +45,14 @@ using namespace rir::pir;
 struct ForcedAt : public std::unordered_map<Value*, Force*> {
     static Force* ambiguous() { return (Force*)0x22; }
 
-    void evalAt(Value* val, Force* force) {
+    void forcedAt(MkArg* val, Force* force) {
         if (!count(val))
             (*this)[val] = force;
+    }
+    void usedAt(MkArg* val, Force* force) {
+        // Used before force, this means we need to keep the promise as is
+        if (!count(val))
+            (*this)[val] = ambiguous();
     }
     bool merge(ForcedAt& other) {
         bool changed = false;
@@ -89,8 +94,17 @@ class ForceDominanceAnalysis : public StaticAnalysis<ForcedAt> {
 
     void apply(ForcedAt& d, Instruction* i) const override {
         auto f = Force::Cast(i);
-        if (f)
-            d.evalAt(getValue(f), f);
+        if (f) {
+            MkArg* arg = MkArg::Cast(getValue(f));
+            if (arg)
+                d.forcedAt(arg, f);
+        } else if (!CastType::Cast(i)) {
+            i->eachArg([&](Value* v) {
+                MkArg* arg = MkArg::Cast(v);
+                if (arg)
+                    d.usedAt(arg, f);
+            });
+        }
     }
 };
 
@@ -140,6 +154,7 @@ void ForceDominance::apply(Closure* cls) {
     ForceDominanceAnalysisResult analysis(cls);
 
     std::unordered_map<Force*, Value*> inlinedPromise;
+    std::unordered_map<MkArg*, MkArg*> forcedMkArg;
 
     // 1. Inline dominating promises
     Visitor::run(cls->entry, [&](BB* bb) {
@@ -180,10 +195,15 @@ void ForceDominance::apply(Closure* cls) {
                             f = Force::Cast(*split->begin());
                             assert(f);
                             f->replaceUsesWith(promRes);
-                            next = split->remove(split->begin());
-                            bb = split;
+                            split->remove(split->begin());
+
+                            MkArg* fixedMkArg =
+                                new MkArg(mkarg->prom, promRes, mkarg->env());
+                            next = split->insert(split->begin(), fixedMkArg);
+                            forcedMkArg[mkarg] = fixedMkArg;
 
                             inlinedPromise[f] = promRes;
+                            bb = split;
                         }
                     }
                 }
@@ -212,6 +232,10 @@ void ForceDominance::apply(Closure* cls) {
             ip = next;
         }
     });
+
+    // 3. replace remaining uses of the mkarg itself
+    for (auto m : forcedMkArg)
+        m.first->replaceUsesIn(m.second, m.second->bb());
 }
 }
 }
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -154,7 +154,7 @@ void ForceDominance::apply(Closure* cls) {
     ForceDominanceAnalysisResult analysis(cls);
 
     std::unordered_map<Force*, Value*> inlinedPromise;
-    std::unordered_map<MkArg*, MkArg*> forcedMkArg;
+    std::unordered_map<Instruction*, MkArg*> forcedMkArg;
 
     // 1. Inline dominating promises
     Visitor::run(cls->entry, [&](BB* bb) {
@@ -217,6 +217,7 @@ void ForceDominance::apply(Closure* cls) {
         auto ip = bb->begin();
         while (ip != bb->end()) {
             auto f = Force::Cast(*ip);
+            auto cast = CastType::Cast(*ip);
             auto next = ip + 1;
             if (f) {
                 // If this force instruction is dominated by another force we
@@ -228,6 +229,13 @@ void ForceDominance::apply(Closure* cls) {
                         f->replaceUsesWith(r);
                     next = bb->remove(ip);
                 });
+            }
+            if (cast) {
+                // Collect aliases of promises for step 3 bellow
+                auto in = Instruction::Cast(cast->arg<0>().val());
+                if (in && forcedMkArg.count(in)) {
+                    forcedMkArg[cast] = forcedMkArg.at(in);
+                }
             }
             ip = next;
         }
--- a/rir/src/compiler/opt/inline.cpp
+++ b/rir/src/compiler/opt/inline.cpp
@@ -41,6 +41,16 @@ class TheInliner {
                         continue;
                     assert(cls->fun->closureEnv() == Env::notClosed());
                     staticEnv = cls->env();
+                } else if (EagerCall::Cast(*it)) {
+                    EagerCall* call = EagerCall::Cast(*it);
+                    auto cls = MkFunCls::Cast(call->cls());
+                    if (!cls)
+                        continue;
+                    inlinee = cls->fun;
+                    if (inlinee->argNames.size() != call->nCallArgs())
+                        continue;
+                    assert(cls->fun->closureEnv() == Env::notClosed());
+                    staticEnv = cls->env();
                 } else if (StaticCall::Cast(*it)) {
                     StaticCall* call = StaticCall::Cast(*it);
                     inlinee = call->cls();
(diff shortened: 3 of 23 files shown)


o-:
There where too many bugs, when we tried to ad-hoc compile to `call_`. That was not a good idea. so I added call instructions to rir, that better match what we need for pir. See the last commit and its description.

this is now ready to review

rir/src/ir/insns.h
```diff
@@ -106,11 +111,24 @@ DEF_INSTR(movloc_, 2, 0, 0, 1)
...
+DEF_INSTR(call_stack_lazy_, 2, -1, 1, 0)
+
+/**
+ * call_stack_:: Like call_, but expects eager arguments on the stack
  *               on top of the callee.
  */
 DEF_INSTR(call_stack_, 2, -1, 1, 0)
 
+/**
+ * static_call_stack_lazy_:: Like call_stack_, but the callee is known
```
JanJecmen: It should mention that the args are promises, as above.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -32,7 +32,7 @@ void Rir2PirCompiler::compileClosure(SEXP closure, MaybeCls success,
     assert(isValidClosureSEXP(closure));
     DispatchTable* tbl = DispatchTable::unpack(BODY(closure));
 
-    if (!tbl->slot(1)) {
+    if (tbl->available(1)) {
```
JanJecmen, line 35: Oh, a bug... Nice catch!

rir/src/compiler/pir/instruction.h
```diff
@@ -725,155 +758,150 @@ SAFE_UNOP(Length);
...
-        for (size_t i = 0; i < nCallArgs(); ++i) {
-            it(arg(i + callArgOffset).val());
-        }
-    }
-
     void printArgs(std::ostream&) override;
 };
 
+// Call instruction for eager calls. Closure is
+// specified as `cls_`, args passed as values.
```
JanJecmen: comment is out of sync: no `cls_` field here

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -18,6 +18,60 @@ class TheCleanup {
...
+                        });
+                        if (allEager) {
+                            auto eagerCall = new StaticEagerCall(
+                                call->env(), call->cls(), args, call->srcIdx,
+                                call->origin());
+                            call->replaceUsesWith(eagerCall);
+                            bb->replace(ip, eagerCall);
+                        }
+                    }
+                }
```
JanJecmen, line 69: Too bad that lambdas cannot be templated yet didn't find a nice way of unifying this :(

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -45,10 +45,15 @@ using namespace rir::pir;
 struct ForcedAt : public std::unordered_map<Value*, Force*> {
     static Force* ambiguous() { return (Force*)0x22; }
```
JanJecmen: is this just something like `{ static Force dummy(nullptr); return &dummy; }`?

rir/src/compiler/pir/instruction.h
```diff
@@ -369,7 +391,7 @@ class VarLenInstruction
         assert(env);
         static_assert(ENV > EnvAccess::None,
```
JanJecmen: `Description.HasEnv`

rir/src/compiler/pir/instruction.h
```diff
@@ -329,37 +342,46 @@ template <Tag ITAG, class Base, Effect EFFECT, EnvAccess ENV>
...
+            return;
+        }
+        args_.push_back(args_.back());
+        args_[nargs() - 2] = InstrArg(a, t);
     }
 
+    void pushArg(Value* a) { pushArg(a, a->type); }
+
     VarLenInstruction(PirType return_type) : Super(return_type) {
         static_assert(ENV == EnvAccess::None,
```
JanJecmen: `!Description.HasEnv`

rir/src/compiler/pir/instruction.h
```diff
@@ -725,155 +758,150 @@ SAFE_UNOP(Length);
...
-    constexpr static size_t callArgOffset = 2;
+    constexpr static size_t clsIdx = 0;
 
     Value* cls() { return arg(clsIdx).val(); }
-    size_t nCallArgs() override { return nargs() - callArgOffset; }
 
     Call(Value * e, Value * fun, const std::vector<Value*>& args, unsigned src)
-        : VarLenInstruction(PirType::valOrLazy(), e) {
+        : CallInstructionImplementation(PirType::valOrLazy(), e) {
         this->pushArg(fun, RType::closure);
```
JanJecmen: Here and also all occurrences below of `this->pushArg`: no need for `this->`, right?

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -18,6 +18,60 @@ class TheCleanup {
         std::unordered_set<size_t> used_p;
         std::unordered_map<BB*, std::unordered_set<Phi*>> used_bb;
 
+        Visitor::run(function->entry, [&](BB* bb) {
+            auto ip = bb->begin();
+            while (ip != bb->end()) {
+                Instruction* i = *ip;
+                auto next = ip + 1;
+
+                // Convert lazy calls to eager calls, if all args are eager.
```
charig, line 27: Is this a cleanup or an optimization?

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -18,6 +18,60 @@ class TheCleanup {
...
+                        if (allEager) {
+                            auto eagerCall = new EagerCall(
+                                call->env(), call->cls(), args, call->srcIdx);
+                            call->replaceUsesWith(eagerCall);
+                            bb->replace(ip, eagerCall);
+                        }
+                    }
+                }
+
+                {
```
charig, line 49: factorize these two?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -89,8 +94,17 @@ class ForceDominanceAnalysis : public StaticAnalysis<ForcedAt> {
 
     void apply(ForcedAt& d, Instruction* i) const override {
         auto f = Force::Cast(i);
-        if (f)
-            d.evalAt(getValue(f), f);
+        if (f) {
+            MkArg* arg = MkArg::Cast(getValue(f));
+            if (arg)
```
charig, line 102: Is this if needed?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -45,10 +45,15 @@ using namespace rir::pir;
 struct ForcedAt : public std::unordered_map<Value*, Force*> {
```
charig: Forced/Used by instead of at?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -89,8 +94,17 @@ class ForceDominanceAnalysis : public StaticAnalysis<ForcedAt> {
...
+                d.forcedAt(arg, f);
+        } else if (!CastType::Cast(i)) {
+            i->eachArg([&](Value* v) {
+                MkArg* arg = MkArg::Cast(v);
+                if (arg)
+                    d.usedAt(arg, f);
+            });
+        }
     }
 };
```
charig, line 113: It is not included here, but isn't the merge to conservative? It moves to ambiguos without checking if the forcedAt is the same. Couldn't that happen on a merge coming from a loop?

rir/src/compiler/pir/instruction.h
```diff
@@ -94,13 +94,14 @@ class Instruction : public Value {
...
-    bool mightIO() const;
-    bool changesEnv() const;
-    bool leaksEnv() const;
-    bool hasEnv() const;
-    bool accessesEnv() const;
+    virtual bool mightIO() const = 0;
+    virtual bool changesEnv() const = 0;
+    virtual bool leaksEnv() const = 0;
+    virtual bool hasEnv() const = 0;
+    virtual bool accessesEnv() const = 0;
```
JanJecmen, line 101: Why abandon the static table with descriptions?
(I like this more, but doesn't it introduce a vtable lookup? I think these cannot be devirtualized)

rir/src/compiler/opt/inline.cpp
```diff
@@ -86,6 +96,7 @@ class TheInliner {
                                 a = cast;
                             }
                             ld->replaceUsesWith(a);
+                            next = bb->remove(ip);
```
charig, line 95: This was missing without being caught by any tests?????

rir/src/compiler/pir/instruction.h
```diff
@@ -725,155 +758,150 @@ SAFE_UNOP(Length);
     VarLenInstruction<Tag::type, type, io, env>
 
 // Common interface to all call instructions
-struct CallInstructionI {
+class CallInstructionI {
```
charig: I would remove the last I CallInstruction is ok or Callable(Instruction)

rir/src/compiler/pir/instruction.h
```diff
@@ -725,155 +758,150 @@ SAFE_UNOP(Length);
...
-            }
-        });
-        return res;
+    virtual void eachCallArgRev(Instruction::ArgumentValueIterator it) = 0;
+    static CallInstructionI* CastCall(Value* v);
+};
+
+template <Tag ITAG, class Base, Effect EFFECT, EnvAccess ENV,
+          bool HAS_TARGET_ARG>
+class CallInstructionImplementation
```
charig, line 771: CallInstruction?

rir/src/compiler/opt/inline.cpp
```diff
@@ -41,6 +41,16 @@ class TheInliner {
                         continue;
                     assert(cls->fun->closureEnv() == Env::notClosed());
                     staticEnv = cls->env();
+                } else if (EagerCall::Cast(*it)) {
```
JanJecmen: How about `if (auto call = EagerCall::Cast(*it))`? But this style is all over the place, so maybe not worth

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -307,8 +307,7 @@ class SSAAllocator {
                 // A, C to be in a stack slot, discard B (it will become
                 // a local variable later) and resize the stack to [xxx]
                 stack.resize(newStackSize);
-                StackArgVisitor::run(
-                    i, [&](Value* arg) { allocation[arg] = stackSlot; });
+                i->eachArgRev([&](Value* arg) { allocation[arg] = stackSlot; });
```
charig, line 310: We should change Rev for Reverse. But the naming is confusing here because eachArg returns the argument while eachArgRev returns the *value of the argument* in reverse mode

charig (commented):
The PR is quite big. It also does some refactorings. I went a couple of hours mostly through the promise inliner that is the part that I mostly know. I leave some comments. It would be great that @JanJecmen do a pass too.

rir/src/interpreter/interp.cpp
```diff
@@ -23,34 +23,37 @@ extern Rboolean R_Visible;
 using namespace rir;
 
 struct CallContext {
-    CallContext(Code* c, unsigned callsiteId, SEXP callee,
-                R_bcstack_t* eagerArgs, SEXP callerEnv, Context* ctx)
-        : eagerArgs(eagerArgs), caller(c->function()), callerEnv(callerEnv),
+    CallContext(Code* c, unsigned callsiteId, SEXP callee, bool hasEagerArgs,
```
JanJecmen, line 26: I find this a bit confusing. In RIR we have instructions that can be `*_lazy_*` and also `*_stack_*`, whereas in PIR we have `*Eager*`.
This context class uses eager and stack args, and eager callee.

Is the following accurate?
- `call_` is not supposed to be generated by PIR
- all args are passed on the stack after PIR
- `Call` ~ `call_stack_lazy_`
- `EagerCall` ~ `call_stack_`
- `StaticCall` ~ `static_call_stack_lazy_`
- `StaticEagerCall` ~ `static_call_stack_`
- in `CallContext`, `hasEagerArgs == true` implies `stackArgs != nullptr`

Summary: in RIR, the `lazy` instructions mean, that the args on the stack are promised, in PIR the `Eager` means the args are values, and `hasEagerCallee` says whether the callee expects values or promises?

Sorry, just trying to make sense of it.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -720,133 +645,141 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+                                   Value* what) {
+                    if (what == Missing::instance()) {
+                        // if missing flows into instructions with more than one
+                        // arg we will need stack shuffling here
+                        assert(MkArg::Cast(instr) &&
+                               "only mkarg supports missing");
+                        cs << BC::push(R_UnboundValue);
+                    } else {
+                        if (!alloc.hasSlot(what)) {
+                            std::cerr << "Don't know how to load the env ";
```
JanJecmen: `s/env/argument/`

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -720,133 +645,141 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+                        if (!alloc.onStack(what))
+                            cs << BC::ldloc(alloc[what]);
+                    }
+                };
+
+                auto loadArg = [&](BB::Instrs::iterator it, Instruction* instr,
+                                   Value* what) {
+                    if (what == Missing::instance()) {
+                        // if missing flows into instructions with more than one
+                        // arg we will need stack shuffling here
```
JanJecmen, line 683: Do you have an example?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -720,133 +645,141 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+                        }
+                        if (!alloc.onStack(what))
+                            cs << BC::ldloc(alloc[what]);
+                    }
+                };
+
+                // Step one: load and set env
+                if (!Phi::Cast(instr)) {
+                    if (instr->hasEnv() && !explicitEnvValue(instr)) {
+                            // If the env is passed on the stack, it needs
```
JanJecmen: doubly indented

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -861,164 +794,91 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
             }
             case Tag::LdFunctionEnv: {
                 // TODO: what should happen? For now get the current env (should
                 // be the promise environment that the evaluator was called
                 // with) and store it into local and leave it set as current
-                if (!alloc.dead(instr)) {
-                    cs << BC::getEnv();
-                    store(it, instr);
-                }
+                cs << BC::getEnv();
```
JanJecmen, line 865: There is now `caller_env_`, can that be used here?

JanJecmen (approved):
Great job cleaning up the PIR to RIR compiler!

rir/src/compiler/opt/cleanup.cpp
o-: it's probably more of an optimization :)

rir/src/compiler/opt/force_dominance.cpp
o-: is what needed?

rir/src/compiler/opt/inline.cpp
o-: well, this was later removed by dead code removal. but this is of course a bit brittle to rely on

rir/src/interpreter/interp.cpp
o-: That is all correct, except, `call_` can be generated by pir. My idea was that we have pass that is scheduled late, that converts `Call`, where none of the values are eager, into a (new) `LazyCall` which translates to `call_`.

I agree the naming is a bit confusing. I would like to keep `Call` in pir, because the idea is, that `Call` can be anything, we don't know yet. If all are eager we convert to eager, if all are lazy, we convert to lazy. But probably `call_stack_lazy` is not a good name, how about `call_stack_promised`?

I think we cannot fully harmonize the names, because fundamentally pir and rir have a different style of describing a call. How about this:

* `Call` -> `call_stack_promised`   (mixed lazy and eager, but all promised)
* `EagerCall` -> `call_stack_eager`
* `LazyCall` -> `call_promised`

o-: (though renaming `call_` might be a tedious task :) )

rir/src/compiler/opt/force_dominance.cpp
charig: `if (arg)` on line 99

o-: but I don't understand, why wouldn't it be necessary? We want to figure out where known promises are forced. and we only know if we see the mkarg.

rir/src/compiler/opt/cleanup.cpp
o-: if it's ok with you, I would fix this code + create a new pass for it in a later PR?

rir/src/compiler/opt/force_dominance.cpp
charig: Well it is probably just my misunderstanding, but isn't it always true that value of a force instruction is a mkarg instruction? That is the only thing I meant. The if(arg) is redundant because it will always be true with my assumption

rir/src/compiler/opt/cleanup.cpp
charig: Sure!

rir/src/compiler/opt/force_dominance.cpp
o-: it's not. see the last condition `f != other.at(v)` on the merge on line 62.

If you refer to the `usedAt`, this checks if the promise is used before being forced. In this case we cannot sink it.

rir/src/compiler/translations/pir_2_rir.cpp
o-: they are overloaded. You can also get the value in forward mode. but we just don't need the argument in reversemode as of yet, so I did not bother adding it.

rir/src/compiler/pir/instruction.h
o-: the problem was that I could not use the predicates in static assert. so, for example the `static_assert(Description.HasEnv)` was not possible, because I don't think we can create the table as a constexpr.

rir/src/compiler/opt/inline.cpp
o-: jeez I didn't know that this is possible, love it! we should switch to this pattern everywhere

rir/src/interpreter/interp.cpp
o-: I am leaving `call_` as is for now

rir/src/compiler/translations/pir_2_rir.cpp
o-: well, let's assume you have `foo(a,missing,c)`, now you have a and c on the stack and you need to push missing in the middle.

rir/src/compiler/translations/pir_2_rir.cpp
o-: no, that's a new env. the idea is that MkEnv creates an environment. So this just reloads the one, that was popped at the setEnv on 968. The idea is:

1. create env
2. set new created env as default env
3. set all variables
4. load env again, since this instruction should produce a value.

this is a new env, has nothing to do with parent_env

rir/src/compiler/opt/force_dominance.cpp
o-: ah, no. you can `force` anything that you are not sure is a value. for example `LdArg` returns a value which may or may not be evaluated. The type is `value^`, the `^` indicates it might be lazy, the `force` removes the `^`, so to speak.

rir/src/ir/insns.h
```diff
@@ -235,6 +240,8 @@ DEF_INSTR(ge_, 0, 2, 1, 0)
 DEF_INSTR(eq_, 0, 2, 1, 0)
 DEF_INSTR(ne_, 0, 2, 1, 0)
 
+DEF_INSTR(identical_, 0, 2, 1, 0)
```
mhyee, line 255: Isn't this a pure instruction?

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -18,6 +18,60 @@ class TheCleanup {
...
 
+        Visitor::run(function->entry, [&](BB* bb) {
+            auto ip = bb->begin();
+            while (ip != bb->end()) {
+                Instruction* i = *ip;
+                auto next = ip + 1;
+
+                // Convert lazy calls to eager calls, if all args are eager.
+                {
+                    auto call = Call::Cast(i);
```
mhyee, line 29: Since this kind of style is used elsewhere, we could change
```
{
    auto call = exp;
    if (call) {
        ...
    }
}
```
to just
```
if (auto call = exp) {
    ...
}
```

rir/src/compiler/opt/cleanup.cpp
mhyee: Yeah, generic lambdas are in C++14 (e.g. `[](autox) { ... }`), but templated lambdas (e.g. `[]<typename T>(T x) { ... }`) were only just voted into C++20. Proposal [here](https://wg21.link/p0428).

The proposal suggests that there may be a workaround using `decltype`, but that seems really ugly to me.

rir/src/interpreter/interp.cpp
mhyee: Urgh, I still don't understand any of this (or I've forgotten everything because it's been so long).

How many call instructions do we have for PIR? How many do we have for RIR? How are the call instructions different, and how do the PIR calls relate to RIR calls? How are arguments passed to callees? Are lazy and eager opposites? What does it mean for something to be eager _and_ promised? How do stack args relate to eager args (because a lot of changes in this diff are eager -> stack).

mhyee (commented):
I'm late, but here are some comments and questions. Mainly, I'll echo @charig in that large PRs are very difficult and time-consuming to review. It's more work for the author, but breaking it down into smaller commits makes it a lot easier for reviewers. (Sure, someone will always have to do the extra work, but it's going to be 1 author or N reviewers.)

rir/src/interpreter/interp.cpp
JanJecmen: Sure, thanks for shedding some light on this. The `promised` naming looks better to me

charig:
About @mhyee comment about the calling instructions...We mentioned this several times during the meetings but it still also happens to me that I get confused with these combinations. Maybe it is time to document this part of the design so we can resort on the documentation when get lost?

PR #155, opened 2018-07-10
remove dispatch_ and dispatch_stack_

o-:
this is our escape fire.

pp with @guido and @JanJecmen


o-:
there is some remaining issue that shows up in the print of `R_ENABLE_JIT=2 tools/R -e 'dend <- as.dendrogram(hclust(dist(USArrests), "ave"));dend2 <- cut(dend, h=70);str(dend2$upper)'`

PR #165, opened 2018-07-20
Make run script more flexible

charig:


JanJecmen:
As a side note, the broken mac build can be fixed by adding `TZ=Europe/Prague` or whatever to the environment.. I think
Don't know how to hack it into the build automatically, because mac needs sudo to tell you the time zone :/

PR #169, opened 2018-07-26
add a flag to enable the pir optimizer on second function invocation

o-:


rir/src/interpreter/interp.cpp
```diff
@@ -609,9 +609,16 @@ SEXP rirCall(const CallContext& call, Context* ctx) {
...
-
     Function* fun = table->at(slot);
+
     fun->registerInvocation();
+    if (slot == 0 && fun->invocationCount == 2) {
+        ctx->optimizer(call.callee());
+        slot = dispatch(call, table);
+        needsEnv = slot == 0;
+        fun = table->at(slot);
+        fun->registerInvocation();
```
charig: why another registerInvocation? Isn't it possible that the dispatch() returns again the rir function if pir can not compile?

rir/src/compiler/opt/inline.cpp
```diff
@@ -19,11 +19,12 @@ class TheInliner {
     TheInliner(Closure* function) : function(function) {}
 
     void operator()() {
```
charig, line 21: Is fuel the proper name here? I personally do not know that terminology. Inlining_threshold or something like that?
btw, it may be more tidy to squash this into another commit.

rir/src/api.cpp
```diff
@@ -155,13 +147,26 @@ REXPORT SEXP pir_tests() {
...
+    if (pir && std::string(pir).compare("on") == 0) {
+        initializeRuntime(rir_compile, pirOpt);
+    } else if (pir && std::string(pir).compare("force") == 0) {
+        initializeRuntime(
+            [](SEXP f, SEXP env) { return pirOpt(rir_compile(f, env)); },
+            [](SEXP f) { return f; });
+    } else if (pir && std::string(pir).compare("force_dryrun") == 0) {
+        initializeRuntime(
+            [](SEXP f, SEXP env) {
+                return pir_compile(rir_compile(f, env), R_FalseValue,
```
charig: Why the pir_compiler here and not in the optimizer callback?

charig (approved):

rir/src/compiler/opt/inline.cpp
o-: I am not sure how standard it is, but I did not invent it myself.. the idea is that the optimizer has some budget to optimize and it can apply optimizations until it runs out of fuel.

PR #172, opened 2018-08-02
Fine grained debugging information

charig:
This PR intends to give developers some more fine-grained control on the debugging information printed when compiling using pir. There is still an **important need** to print the name of the function being compiled.
I even tried with walking through the environment programatically but I think the binding between name and the SEXP is done after the compilation...

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, verbose = 0, dryRun = FALSE) {
-    .Call("pir_compile_", what, verbose, dryRun)
+pir.compile <- function(what, verbose = 0L, dryRun = FALSE) {
+    .Call("pir_compile", what, verbose, dryRun)
 }
 
 pir.tests <- function() {
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, verbose = FALSE, dryRun = FALSE) {
-    .Call("pir_compile", what, verbose, dryRun)
+pir.compile <- function(what, verbose = 0, dryRun = FALSE) {
+    .Call("pir_compile_", what, verbose, dryRun)
 }
 
 pir.tests <- function() {
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -95,7 +95,7 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-REXPORT SEXP pir_compile(SEXP what, unsigned int verbose, SEXP dryRun_) {
+REXPORT SEXP pir_compile(SEXP what, uint32_t verbose, SEXP dryRun_) {
     bool dryRun = false;
     if (dryRun_ && TYPEOF(dryRun_) == LGLSXP && Rf_length(dryRun_) > 0 &&
         LOGICAL(dryRun_)[0])
@@ -140,7 +140,7 @@ REXPORT SEXP pir_tests() {
 
 // startup ---------------------------------------------------------------------
 
-uint pir_verbose = (getenv("PIR_VERBOSE")) ? 
+uint32_t pir_verbose = (getenv("PIR_VERBOSE")) ? 
      std::stoul(getenv("PIR_VERBOSE"), nullptr, 0) : 0;
 
 SEXP pirOpt(SEXP fun) { return pir_compile(fun, pir_verbose, R_FalseValue); }
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -95,7 +95,7 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
+REXPORT SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
     bool dryRun = false;
     if (dryRun_ && TYPEOF(dryRun_) == LGLSXP && Rf_length(dryRun_) > 0 &&
         LOGICAL(dryRun_)[0])
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -95,7 +95,7 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-REXPORT SEXP pir_compile(SEXP what, uint32_t verbose, SEXP dryRun_) {
+SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
     bool dryRun = false;
     if (dryRun_ && TYPEOF(dryRun_) == LGLSXP && Rf_length(dryRun_) > 0 &&
         LOGICAL(dryRun_)[0])
@@ -133,6 +133,10 @@ REXPORT SEXP pir_compile(SEXP what, uint32_t verbose, SEXP dryRun_) {
     return what;
 }
 
+REXPORT SEXP pir_compile(SEXP what, SEXP verbose, SEXP dryRun_) {
+    return pir_compile_(what, INTEGER(INTEGER(verbose))[0], dryRun_);
+}
+
 REXPORT SEXP pir_tests() {
     PirTests::run();
     return R_NilValue;
@@ -143,7 +147,7 @@ REXPORT SEXP pir_tests() {
 uint32_t pir_verbose = (getenv("PIR_VERBOSE")) ? 
      std::stoul(getenv("PIR_VERBOSE"), nullptr, 0) : 0;
 
-SEXP pirOpt(SEXP fun) { return pir_compile(fun, pir_verbose, R_FalseValue); }
+SEXP pirOpt(SEXP fun) { return pir_compile_(fun, pir_verbose, R_FalseValue); }
 
 bool startup() {
     auto pir = getenv("PIR_ENABLE");
@@ -156,8 +160,8 @@ bool startup() {
     } else if (pir && std::string(pir).compare("force_dryrun") == 0) {
         initializeRuntime(
             [](SEXP f, SEXP env) {
-                return pir_compile(rir_compile(f, env), pir_verbose,
-                                   R_TrueValue);
+                return pir_compile_(rir_compile(f, env), pir_verbose,
+                                    R_TrueValue);
             },
             [](SEXP f) { return f; });
     } else {
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -143,16 +143,6 @@ REXPORT SEXP pir_tests() {
 uint pir_verbose = (getenv("PIR_VERBOSE")) ? 
      std::stoul(getenv("PIR_VERBOSE"), nullptr, 0) : 0;
 
-/********** VERBOSE BIT MODES ***********
-1: RIR Original version
-2: PIR Raw
-3: PIR Optimization phases
-4: PIR Inlining phases
-5: PIR Convert to CSSA
-6: PIR After liveness analysis phase
-7: RIR After stack allocation phase 
-8: RIR After passing through PIR*/
-
 SEXP pirOpt(SEXP fun) { return pir_compile(fun, pir_verbose, R_FalseValue); }
 
 bool startup() {
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -134,7 +134,7 @@ REXPORT SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
 }
 
 REXPORT SEXP pir_compile(SEXP what, SEXP verbose, SEXP dryRun_) {
-    return pir_compile_(what, INTEGER(INTEGER(verbose))[0], dryRun_);
+    return pir_compile_(what, (uint32_t)INTEGER(INTEGER(verbose))[0], dryRun_);
 }
 
 REXPORT SEXP pir_tests() {
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -140,7 +140,9 @@ REXPORT SEXP pir_tests() {
 
 // startup ---------------------------------------------------------------------
 
-uint pir_verbose = std::stoul(getenv("PIR_VERBOSE"), nullptr, 0);
+uint pir_verbose = (getenv("PIR_VERBOSE")) ? 
+     std::stoul(getenv("PIR_VERBOSE"), nullptr, 0) : 0;
+
 /********** VERBOSE BIT MODES ***********
 1: RIR Original version
 2: PIR Raw
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -95,7 +95,7 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-REXPORT SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
+SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
     bool dryRun = false;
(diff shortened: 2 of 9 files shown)


rir/src/compiler/translations/rir_compiler.h
```diff
@@ -26,10 +26,14 @@ class RirCompiler {
     }
 
     bool isVerbose() { return verbose; }
-    void setVerbose(bool v) { verbose = v; }
+    bool shouldPrintOriginalVersion() { return verbose & 0X1; }
+    bool shouldPrintCompiledVersion() { return verbose & 0X10; }
+    bool shouldPrintOptimizations() { return verbose & 0X100; }
+    bool shouldPrintInliningVersions() { return verbose & 0X1000; }
```
JanJecmen: all the bit masks are hex numbers, ie. `x & 0x10` will look at the sixteenth bit. I guess you meant to look at the second, which would be `x & 0x2`

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -93,13 +93,13 @@ class SSAAllocator {
 
     SSAAllocator(Code* code, bool verbose)
         : cfg(code), dom(code), code(code), bbsSize(code->nextBBId) {
-        computeLiveness(verbose);
+        computeLiveness(verbose & 0X10000);
```
mhyee: Nitpick: I prefer to write my hexadecimal literals with a lowercase `x`. Not sure what the consensus here is.

More generally: it would be nice to use names (e.g. `const unsigned int flagName = ...`, or an enum) rather than magic constants.

rir/src/compiler/translations/rir_compiler.h
mhyee: (Fifth bit, not sixteenth.)

Alternatively, C++14 has binary literals, e.g. `0b10`. Or you could use bitshifts, e.g. `1 << 1`.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -64,28 +63,47 @@ void Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
         srcFunction, formals.names, closureEnv, [&](Closure* pirFunction) {
             Builder builder(pirFunction, closureEnv);
             Rir2Pir rir2pir(*this, srcFunction);
+            if (isVerbose()) {
+                std::cout << "\n\n*********************************************"
```
mhyee: Nitpick: I know this is because `clang-format` want it this way, but I actually find this to be worse than just going past the 80 column max. Is there a way to tell `clang-format` to ignore some lines?

rir/src/compiler/translations/rir_compiler.h
charig: :). 
Fixed! 
Thanks.

rir/src/compiler/translations/rir_compiler.h
```diff
@@ -26,10 +26,14 @@ class RirCompiler {
...
+    void setVerbose(uint v) { verbose = v; }
 
   private:
-    bool verbose = false;
+    uint verbose = false;
```
mhyee: `uint` is not a standard type. But `unsigned int` and `uint32_t` are.

Also, if it's an int, it should be initialized to `0` and not `false`.

rir/src/compiler/translations/pir_2_rir.h
```diff
@@ -15,10 +15,10 @@ class Pir2RirCompiler {
 
     void compile(Closure* cls, SEXP origin);
 
-    bool shouldPrintCSSA() { return verbose & 0X10; }
-    bool shouldPrintAllocations() { return verbose & 0X20; }
-    bool shouldPrintLiveness() { return verbose & 0X40; }
-    bool shouldPrintRIRAfterPIR() { return verbose & 0X80; }
+    bool shouldPrintCSSA() { return verbose & 0b10000; }
```
mhyee: Sorry, but now that I see it, I think the binary literals are less readable. :(

rir/src/compiler/translations/pir_2_rir.cpp
mhyee: Missed replacing this magic constant.

rir/src/api.cpp
```diff
@@ -95,12 +95,8 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-REXPORT SEXP pir_compile(SEXP what, SEXP verbose, SEXP dryRun_) {
-    bool debug = false;
+REXPORT SEXP pir_compile(SEXP what, unsigned int verbose, SEXP dryRun_) {
```
mhyee: `unsigned int` or `uint32_t`?

rir/src/api.cpp
```diff
@@ -147,7 +140,10 @@ REXPORT SEXP pir_tests() {
 
 // startup ---------------------------------------------------------------------
 
-SEXP pirOpt(SEXP fun) { return pir_compile(fun, R_FalseValue, R_FalseValue); }
+uint pir_verbose = (getenv("PIR_VERBOSE")) ?
```
mhyee: `uint32_t` or `unsigned int`.

rir/src/api.h
```diff
@@ -8,6 +8,6 @@
 extern int R_ENABLE_JIT;
 
 REXPORT SEXP rir_eval(SEXP, SEXP);
-REXPORT SEXP pir_compile(SEXP, SEXP, SEXP);
+REXPORT SEXP pir_compile(SEXP, unsigned int, SEXP);
```
mhyee: Or `uint32_t`? (Not sure how much it matters, but let's be consistent.)

rir/src/compiler/translations/rir_compiler.h
```diff
@@ -26,10 +27,14 @@ class RirCompiler {
...
-    void setVerbose(bool v) { verbose = v; }
+    bool shouldPrintOriginalVersion() { return verbose & PRINT_ORIGINAL_MASK; }
+    bool shouldPrintCompiledVersion() { return verbose & PRINT_RAW_PIR_MASK; }
+    bool shouldPrintOptimizations() { return verbose & PRINT_OPT_PHASES_MASK; }
+    bool shouldPrintInliningVersions() { return verbose & PRINT_INLINIG_MASK; }
+    void setVerbose(uint v) { verbose = v; }
 
   private:
-    bool verbose = false;
+    uint32_t verbose = 0;
```
mhyee: `uint32_t` or `unsigned int`?

rir/src/compiler/translations/pir_2_rir.h
```diff
@@ -10,11 +11,17 @@ namespace pir {
 
 class Pir2RirCompiler {
   public:
-    bool verbose = false;
+    uint verbose = 0;
```
mhyee: `uint32_t` or `unsigned int`

rir/src/compiler/translations/rir_compiler.h
```diff
@@ -26,10 +27,14 @@ class RirCompiler {
     }
 
     bool isVerbose() { return verbose; }
-    void setVerbose(bool v) { verbose = v; }
+    bool shouldPrintOriginalVersion() { return verbose & PRINT_ORIGINAL_MASK; }
+    bool shouldPrintCompiledVersion() { return verbose & PRINT_RAW_PIR_MASK; }
+    bool shouldPrintOptimizations() { return verbose & PRINT_OPT_PHASES_MASK; }
+    bool shouldPrintInliningVersions() { return verbose & PRINT_INLINIG_MASK; }
+    void setVerbose(uint v) { verbose = v; }
```
mhyee: `uint32_t` or `unsigned int`

rir/src/compiler/pir/pir.h
```diff
@@ -1,6 +1,18 @@
 #ifndef PIR_PIR_H
 #define PIR_PIR_H
 
+// clang-format off
+//********** VERBOSE BIT MASK ***********
+#define PRINT_ORIGINAL_MASK   0x1  // bit 1: RIR Original version
```
mhyee: Sorry to keep nitpicking this, but I prefer to avoid C-style macros when possible. I'd suggest turning these to consts, e.g.

`const uint8_t PrintOriginalMask = 0x1`

I suppose we could use `uint8_t` instead of `uint32_t`.

(Too bad we can't really use enum classes unless we override all the bitwise operators. Probably not worth the trouble.)

o-:
@charig maybe you want to look at https://github.com/reactorlabs/rir/blob/master/rir/src/utils/EnumSet.h ?

The way you use it:

```
enum class DebugOptions{
 FIRST = optA,
 optA,
 optB,
 optC,
 LAST = optC
};
typedef EnumSet<DebugOptions> DebugOptionsSet;
DebugOptionsSet options = DebugOptionsSet() | DebugOption::optA | DebugOption::optB;
if (options.includes(DebugOptions::optA)) {
  ...
}
```

This produces essentially the same bitset code, but gives you much safer and nicer abstractions.

charig:
@JanJecmen Thanks for the pointer. I could use that but it seems a llittle like an overkill here. I would still need to have the constants to map from the user input to the elements of the set. And the entry point is only one.

o-:
why would it be overkill? it should not be more complicated, nor more verbose.


btw, `pir_compile` is called from R. See `rir/R/rir.R`. So you can only pass SEXP's as arguments. So we probably want to wrap it in an R integer?

you could do the following to easily create the debug options from R:

in `rir/R/rir.R`
```
pirDebugOpts(optionA=FALSE, optionB=FALSE, optionC=FALSE)
  .Call(pirDebugOpts, optionA, optionB, optionC)
```

then in `api.cpp`
```
REXPORT SEXP pirDebugOpts(SEXP optionA, SEXP optionB, SEXP optionC) {
  DebugOptionSet options;
  if (optionA != R_FalseValue)
    options = options | DebugOption::optionA;
  ...
  SEXP res = Rf_AllocVector(INTSXP,1);
  INTEGER(res)[0] = options.set_.to_ulong();    // <- should expose this as new api to EnumSet
  return res;
}
```

Now from R we can simply do:
```
pir.compile(function, pirDebugOpts(optionC=TRUE))
```

charig:
The original idea is that the user could simply add a parameter from the command line as an environment variable. I do not get how to convert that number into an enum set without duplicating the code.

o-:
yeah, i see. but then creating the magic number by manually constructing the bitset is also not super user friendly. Instead of an env variable, you could also have an api from R. such as `setPirDebug(pirDebugOpts(optionA=TRUE))`. This could be as easy as:

```
setPirDebug <- function(opts) .Call(setPirDebug, opts)
```
and
```
REXPORT SEXP setPirDebug(SEXP opts) {
  .. check opts is integer ...
  pirDebugOptions = INTEGER(opts)[0];
  return R_NilValue;
}
```

anyway, my main point was that pir_compile cannot take an integer, but only SEXP. So if you want to move forward with your version, you need to duplicate `pir_compile`, such that we have one version we can call from R.

which can be as simple as:

```
REXPORT SEXP pir_compile(SEXP what, SEXP verbose) {
  ... check verbose is integer...
  return pirCompile(what, INTEGER(verbose)[0]);
}

SEXP pirCompile(SEXP what, unsigned int verbose) {
  ...
}
```

but to be honest I would prefer an api with meaningful names instead of manually constructing bitsets.

rir/R/rir.R
```diff
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, verbose = FALSE, dryRun = FALSE) {
-    .Call("pir_compile", what, verbose, dryRun)
+pir.compile <- function(what, verbose = 0, dryRun = FALSE) {
+    .Call("pir_compile_", what, verbose, dryRun)
```
o-: this should be `pir_compile` not, `pir_compile_`

rir/src/api.cpp
```diff
@@ -133,6 +133,10 @@ REXPORT SEXP pir_compile(SEXP what, uint32_t verbose, SEXP dryRun_) {
     return what;
 }
 
+REXPORT SEXP pir_compile(SEXP what, SEXP verbose, SEXP dryRun_) {
+    return pir_compile_(what, (uint32_t)INTEGER(INTEGER(verbose))[0], dryRun_);
```
o-: Why twice `INTEGER`?

Also, this will segv, when called with anything but an INTSXP, so it needs checks, that gracefully handle non integer values (like throw an R level error...). Something like
`if (TYPEOF(verbose) != INTSXP || Rf_length(verbose) < 1) error...`

rir/src/api.cpp
```diff
@@ -95,7 +95,7 @@ REXPORT SEXP rir_body(SEXP cls) {
     return f->container();
 }
 
-REXPORT SEXP pir_compile(SEXP what, uint32_t verbose, SEXP dryRun_) {
+REXPORT SEXP pir_compile_(SEXP what, uint32_t verbose, SEXP dryRun_) {
```
o-: no REXPORT. This function should not be called from R

charig:
So I left a working version with the original approach. IMO it is ok for start using it and see if it is useful. If then it becomes worth to convert it into a more tidy R-level debugging option we can for sure do the refactoring. Anyway, if everybody prefers to go for the last option now I have no problem to do that.

rir/src/api.cpp
o-: REXPORT means we want to call this function from R code, using `.Call`. In this case R assumes that it has a signature, which takes only SEXP arguments and returns a SEXP.

A function with uint32_t as argument cannot be REXPORT

mhyee:
Some quick feedback, since I used this for 10 min while working on the debugging breakpoints:

I wanted to see the original RIR bytecode, the original PIR bytecode, the final PIR bytecode (after optimizations), and then the final RIR bytecode. In the end, I set bits 1, 2, and 8 (`PRINT_ORIGINAL_MASK | PRINT_RAW_PIR_MASK | PRINT_FINAL_RIR_MASK`). I guess I was missing bit 6 (`PRINT_LIVENESS_MASK`) if I wanted the final optimized PIR, but that wasn't clear from the comments.

I had to pull out my programmer's calculator to figure out that this was 131 in decimal. (I also tried 255 to get everything, but that was far too messy.)

I found it best to set it through the RIR API, even though fumbling with the bitmasks manually was annoying. Originally, I tried setting it on the command line, but then it started printing the bytecode for all the library functions and other things I didn't care about.

For my specific situation, I found it useful (since I didn't want to see all the intermediate PIR bytecode between phases). But other people may want to play around with it before we invest too much time refactoring it.

charig:
Thanks @mhyee for the feedback! Great that, at least, you found it useful.

Just a couple of small clarifications:

1. I had the same problem with the input in decimal format. That is why I used `std::stoul()` to parse the input because then you can write the input in hex if you prefer what makes life much easier: https://github.com/reactorlabs/rir/blob/master/rir/src/api.cpp#L150.

2. I also experienced the problem that sometimes lot of functions get printed while I want only one. I was thinking on adding one more parameter specifying a kind of function filter. The problem is that we still did not resolve the binding of functions to names when compiling. We must discuss about that.

o-:
sorry, my fingers are faster than my brain.
https://github.com/reactorlabs/rir/commit/36f65aecf433fc86f70571c4c64541eecbf508ca
so I quickly tried to implement the R api I tried to pitch yesterday, that allows you to create the bitsets.

as to 2., that would be cool! it's not always possible though, but we could do a simple heuristic:
1. if you do `[rir|pir].compile(f)`, then the binding is probably `f`. you can get that with substitute. for example: `rir.compile <- function(what) { name <- if (is.name(substitute(what))) substitute(what) else "dunno"); ...`
2. if we jit, then you can look at the ast of the last call, it will be something like `exp(a1,a2,a3)` so `CAR(ast)` is the `exp` left of the `(`, if that one is a symbol, then that's probably the name of the function...

mhyee:
How much source information do we have? It might be enough to filter based on the source filename, or to print the filename and line number instead of the address.

PR #173, opened 2018-08-02
fix promises keeping ast in pir

JanJecmen:
Note: this inserts the source into the source pool when going from PIR to RIR, we should maybe rewrite the `Pool` to work with sources as well? But this PR is just to fix the bug.

### DIFF
--- a/rir/src/compiler/opt/inline.cpp
+++ b/rir/src/compiler/opt/inline.cpp
@@ -112,7 +112,7 @@ class TheInliner {
                                 mk->prom = function->promises[newPromId[id]];
                             } else {
                                 Promise* clone =
-                                    function->createProm(mk->prom->ast);
+                                    function->createProm(mk->prom->srcPoolIdx);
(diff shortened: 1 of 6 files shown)


charig (approved):
So, the ast field is actually an index to the constant pool where the ast should be stored, isn't it? IMO I'd prefer a name more according to that behavior for the variable. astPoolIndex or something like that?

JanJecmen:
Ok sure. (One detail is we have two pools: a constant pool and a source pool, this is in the source one)

mhyee:
```
    // TODO: asts are NIL for now, how to deal with that? after inlining
    // functions and asts don't correspond anymore
```

I think this will still be an issue. We obviously punted on it which led to this bug, and this PR fixes it, but there's still the problem of inlined functions and ASTs. And as we've learned with this bug, some R code really does need to be able to access the AST...

charig (approved):

PR #174, opened 2018-08-02
Add an evaluation breakpoint for R code

mhyee:
In R code, insert a call to `debug.break()`. If a debugger is attached,
when this function is evaluated, it will trigger a breakpoint.
Otherwise, R will simply crash.

---

Note that you'll end up deep in some call stack, so it will take some effort to step out of the call and back to the point of interest (and hopefully you don't step past it either). But this should get you much closer to the point of failure.

A related thing would be to have a debug instruction that triggers a breakpoint during (1) RIR compilation, (2) RIR to PIR compilation, and/or (3) PIR to RIR compilation. But it seems like this would be a lot more work, and would require changes to GNU R.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -43,3 +43,9 @@ rir.eval <- function(what, env = globalenv()) {
 rir.body <- function(f) {
     .Call("rir_body", f);
 }
+
+# breakpoint during evaluation
+# insert a call to `debug.break()` in R code and get a breakpoint when the function is evaluated
+debug.break <- function() {
+    .Call("debug_break")
+}
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -45,7 +45,7 @@ rir.body <- function(f) {
 }
 
 # breakpoint during evaluation
-# insert a call to `debug.break()` in R code and get a breakpoint when the function is evaluated
-debug.break <- function() {
+# insert a call to `.debug.break()` in R code and get a breakpoint when the function is evaluated
+.debug.break <- function() {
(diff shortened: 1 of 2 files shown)


JanJecmen:
Hello `no-segfault.R`, my old friend ♪

There is a test generated by GNU-R for all APIs that will call this

charig:
IMO, you can remove the int 3. It makes things easier. When we want to debug, we just set a breakpoint on `debug_break()` at the `return R_NilValue` line.

mhyee:
Thanks for the pointer. I managed to get around it by renaming the function to make it hidden.

(I prefer to keep the `int 3` to make it explicit. It also seems kind of pointless to have a function that just returns `R_NilValue`.)

o-:
why don't we lower this to `int3_` in rir? that would put the breakpoint right into the bc stream...

in `rir/src/ir/Compiler.cpp` add a `compileSpecialCall` case that lowers `rir::.debug.break` to `int3_`. 


Probably would also need some support in pir (ie. add a native breakpoint instruction in `pir/instruction_list.h` and `pir/instructions.h`, then add the rir2pir translation in `rir_2_pir/stack_machine.cpp` and the back translation in `pir_2_rir.cpp`), so we can optimize a function with breakpoint.

mhyee:
Great suggestion, and thanks for the pointers!

Originally, this was just a quick fix. Breaking in the bytecode stream is a lot more satisfying. And a lot of this work could be reused for a break-during-compilation instruction.

I'll play around with this a bit more and open a new PR.

PR #176, opened 2018-08-06
Add an evaluation breakpoint for R code (attempt 2)

mhyee:
Supersedes #174. (Closing that PR was a mistake.)

Add a call to `.debug.break()` in your R code, and we compile it to the `int3_` instruction, so you get a nice breakpoint while evaluating R code. New in this version of the PR, we can compile the instruction to PIR and back.

One thing that bugs me is the `if (strcmp(CHAR(PRINTNAME(fun)), ".debug.break") == 0)` test in `Compiler.cpp`. Probably the "cleanest" way to do this is add a new symbol to R, so we don't have to compare strings? Thoughts on this?

Also, the reason I closed the other PR was because I wanted to do more, i.e. implement compilation breakpoints. But after experimenting for a bit, I'm not sure it's really worth it. You end up deep in some call stack, it's even harder to get back to a known location, and it's probably easier to put a breakpoint at `compileClosure` or whatever. And if you really wanted to break when `.debug.break()` is compiled, there's now obvious places to add the breakpoints.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -46,6 +46,5 @@ rir.body <- function(f) {
 
 # breakpoint during evaluation
 # insert a call to `.debug.break()` in R code and get a breakpoint when the function is evaluated
-.debug.break <- function() {
-    .Call("debug_break")
-}
+# this is an empty function because it gets directly lowered to the "int3_" bytecode
+.debug.break <- function() { }
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -45,6 +45,8 @@ rir.body <- function(f) {
 }
 
 # breakpoint during evaluation
-# insert a call to `.debug.break()` in R code and get a breakpoint when the function is evaluated
-# this is an empty function because it gets directly lowered to the "int3_" bytecode
-.debug.break <- function() { }
+# insert a call to `.debug.break()` in R code and get a breakpoint when it is evaluated
+# note: the actual body of this function is replaced by an "int3_" bytecode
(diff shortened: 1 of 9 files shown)


o-:
looks good, thanks.

> One thing that bugs me is the if (strcmp(CHAR(PRINTNAME(fun)), ".debug.break") == 0)

Yes, new symbols can be created using `Rf_install("asdf")`. We have our own list of symbols we create at `rir/src/R/Symbols.[h|cpp]`. just add it there and you should be able to say `if(fun == symbol::DebugBreak)`.

rir/R/rir.R
```diff
@@ -43,3 +43,8 @@ rir.eval <- function(what, env = globalenv()) {
 rir.body <- function(f) {
     .Call("rir_body", f);
 }
+
+# breakpoint during evaluation
+# insert a call to `.debug.break()` in R code and get a breakpoint when the function is evaluated
+# this is an empty function because it gets directly lowered to the "int3_" bytecode
+.debug.break <- function() { }
```
o-: maybe put an `error("missed breakpoint")` here, so it's less surprising, if one forgets to compile the code?

charig: Is it to cumbersome to create a SEXP for an empty function? I'd remove the function from here and just compile a programatically created empty function with the proper symbol when the compiler initializes.

mhyee: I prefer @o-'s suggestion. It's simpler and less surprising (tracking down a programmatically-generated function is not very fun).

PR #177, opened 2018-08-07
Cleanup in preparation for typefeedback

o-:


o-:
A Bunch of cleanups, in preparation for collecting typefeedback.

I wanted to simplify the Code struct. For that I removed the CallSite struct part. This means that I put the main parts of the CallSite struct into the BC stream, and the non-essential parts I threw away for now.

I think this makes a lot more sense, but this also means, that i had to add 2 BC and 3 BCs became variable length. On the plus side, the BC objects from `ir/BC_inc.h` now really capture the whole BC, with all arguments, and it is no longer necessary to pass around CallSite pointers with it.

Also, I removed the CodeEditor, since that one was quite intertwined with the CallSite structs.

The idea is to later add a new (and better designed) part in the Code object, where we can store additional metadata for any BC. But the idea is, that the metadata is not necessary to interpret the behavior of a BC!

rir/src/runtime/Code.h
```diff
@@ -202,7 +202,6 @@ struct CallSite {
 
     uint32_t hasNames : 1;
     uint32_t hasTarget : 1;
-    uint32_t hasImmediateArgs : 1;
```
JanJecmen: `free` should be 30 now, but not sure if it even matters

rir/src/interpreter/interp.cpp
```diff
@@ -1427,9 +1425,12 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
...
             Immediate id = readImmediate();
             advanceImmediate();
-            advanceImmediate(); // nargs, TODO: remove
-            CallContext call(c, id, ostack_top(ctx), getenv(), ctx);
+            size_t n = readImmediate();
+            advanceImmediate();
+            CallContext call(c, id, ostack_top(ctx), (Immediate*)pc, getenv(),
+                             ctx);
             res = doCall(call, ctx);
+            advanceImmediateN(n);
```
JanJecmen: Why is this after the call?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -923,29 +923,28 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
 
             case Tag::Call: {
                 auto call = Call::Cast(instr);
-                cs.insertCall(Opcode::call_, call->nCallArgs(), {},
-                              Pool::get(call->srcIdx));
+                cs.insertCall(call->nCallArgs(), Pool::get(call->srcIdx));
```
JanJecmen: So, a bit unrelated, but any reason why the src indices for calls are in the constant pool and not the source pool?
(and also why the source pool does not manage duplicities?)

rir/src/interpreter/interp.cpp
```diff
@@ -73,6 +90,11 @@ struct CallContext {
         return ostack_at_cell(stackArgs + i);
     }
 
+    SEXP name(unsigned i, Context* ctx) const {
```
JanJecmen, line 91: Another unrelated question: why do we pass around the Context pointers everywhere, when the context is only ever the globalContext?

rir/src/R/Printing.h
```diff
@@ -0,0 +1,17 @@
+#ifndef RIR_R_PRINT
+#define RIR_R_PRINT
+
+#include "R/r.h"
+#include "utils/capture_out.h"
+
+namespace rir {
+
+static std::string dumpSexp(SEXP src, size_t length = 50) {
```
JanJecmen, line 9: Maybe it would be useful to also include the type of the sexp in the dump? But not sure..

JanJecmen (approved):
I like the simplifications!

rir/src/runtime/Code.h
o-: uhm, that was not the final version :) its gone now

rir/src/compiler/translations/pir_2_rir.cpp
o-: ehm, no, not really.

rir/src/interpreter/interp.cpp
o-: yeah, there was once the idea to have it ready for multple contexts, i was not involved.

rir/src/R/Printing.h
o-: I mostly use it for the disassembly, and there i mainly care about it not being cluttered. for other cases maybe...

PR #180, opened 2018-08-07
pir debugging flags: use enum set and more user friendly api

o-:
wrote this to try it out, might as well put it here. I have no opinions on whether we should merge it...

### DIFF
--- a/rir/src/compiler/debugging.h
+++ b/rir/src/compiler/debugging.h
@@ -1,0 +1,44 @@
+#ifndef PIR_DEBUGGING
+#define PIR_DEBUGGING
+
+#include "utils/EnumSet.h"
+
+namespace rir {
+namespace pir {
+
+#define LIST_OF_PIR_PRINT_DEBUGGING_FLAGS(V)                                   \
+    V(PrintOriginal)                                                           \
+    V(PrintRawPir)                                                             \
+    V(PrintOptimizationPasses)                                                 \
+    V(PrintInlining)                                                           \
+    V(PrintLiveness)                                                           \
+    V(PrintStackAllocation)                                                    \
+    V(PrintFinalPir)
+
+#define LIST_OF_PIR_DEBUGGING_FLAGS(V)                                         \
+    V(ShowWarnings)                                                            \
+    V(DryRun)                                                                  \
+    V(PreserveVersions)                                                        \
+    V(DebugAllocator)                                                          \
+    LIST_OF_PIR_PRINT_DEBUGGING_FLAGS(V)
+
+enum class DebugFlag {
+#define V(n) n,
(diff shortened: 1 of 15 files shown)


rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -1106,7 +1108,7 @@ void Pir2RirCompiler::compile(Closure* cls, SEXP origin) {
     Pir2Rir pir2rir(*this, cls);
     auto fun = pir2rir.finalize();
 
-    if (shouldPrintRIRAfterPIR()) {
+    if (debug.includes(DebugFlag::PrintFinalPir)) {
```
mhyee: This should be named `PrintFinalRir`.

rir/src/compiler/debugging.h
```diff
@@ -0,0 +1,44 @@
+#ifndef PIR_DEBUGGING
+#define PIR_DEBUGGING
+
+#include "utils/EnumSet.h"
+
+namespace rir {
+namespace pir {
+
+#define LIST_OF_PIR_PRINT_DEBUGGING_FLAGS(V)                                   \
```
mhyee: Add some documentation here? My main concern is that someone will update the list here but then forget to update `rir.R`.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -621,12 +622,13 @@ class Pir2Rir {
 size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
     toCSSA(code);
 
-    if (compiler.shouldPrintCSSA())
+    if (compiler.debug.includes(DebugFlag::PrintLiveness))
```
mhyee: Maybe it's useful to alias `PrintFinalPir` or something? It's not immediately obvious that `PrintLiveness` gives you the final PIR bytecode (is this even correct?) before it's compiled back to RIR, and I think that's something useful to see.

Sure, there's some redundancy, but changing the condition to `compiler.debug.includes(DebugFlag::PrintLiveness | DebugFlag::PrintFinalPir)` is a small price for convenience.

mhyee (approved):
I think it's worth merging and then seeing how we like it. As I mentioned [in the other PR](https://github.com/reactorlabs/rir/pull/172#issuecomment-410849167), I found it useful though the interface was a bit unwieldy. This should improve things.

It may be worth figuring out what good verbosity/debug levels are. Before, it was easy to just type `pir.compile(f, verbose=TRUE)` but now you have to type `pir.compile(f, debugFlags=pir.debugFlags(PrintFinalPir=TRUE,DebugAllocator=TRUE))`. Would be nice if you could just write something like `pir.compile(f, debugFlags=pir.verbose1)` or something. But let's wait until we have some idea of what we want.

PR #181, opened 2018-08-08
Refactor source maps and function writer +  Check BC have sources attached

o-:
1. Rewrite the function writer, such that it can strip nops. In this commit I also simplified the source maps. They are now just an ordered list of pcOffset to srcidx.
2. In the BC verifier check that BC that need sources actually have them.


JanJecmen (approved):
One thing, you mention that you found a bunch of `decode` uses but none are changed in this pr?

rir/src/runtime/Code.cpp
```diff
@@ -4,17 +4,16 @@
 #include "utils/Pool.h"
 
 namespace rir {
-Code::Code(SEXP ast, unsigned cs, unsigned sourceSize, unsigned offset,
+Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
            bool isDefaultArg, size_t localsCnt) {
```
mhyee: Just noticing this now, but why not use the member initialization list?

rir/src/runtime/Code.h
```diff
@@ -94,13 +92,17 @@ struct Code {
...
+     *   srcList       cp_idx (ast)      srcLength * sizeof(SrcEntry)
      *
      */
 
+    // The source list contains pcOffset to src index
+    struct SrclistEntry {
+        unsigned pcOffset;
+        unsigned srcIdx;
+    };
+    static_assert(sizeof(SrclistEntry) == 2 * sizeof(unsigned), "");
```
mhyee: If/when we move to C++17 we can get rid of the `, ""` 😃

rir/src/ir/CodeStream.h
```diff
@@ -21,34 +21,32 @@ class CodeStream {
...
 
     SEXP ast;
 
     unsigned nextLabel = 0;
-    std::map<unsigned, BC::Label> patchpoints;
-    std::vector<unsigned> label2pos;
 
-    std::vector<unsigned> sources;
+    // Ordered map on purpose, see FunctionWriter
+    std::map<unsigned, BC::Label> patchpoints;
```
mhyee: Add some comments to explain what `unsigned` refers to?

For the first two, I can figure out that it's a map of positions to labels (or positions to list of labels), but it's easier if it's right there in a comment. For the third, a map of `unsigned` to `unsigned` isn't very helpful.

rir/src/ir/CodeStream.h
```diff
@@ -21,34 +21,32 @@ class CodeStream {
...
+    std::map<unsigned, std::list<BC::Label>> labels;
```
mhyee: I'd prefer `std::vector`. Conventional wisdom is that it's a good enough default and `std::list` is rarely the better choice. If we really think there's a difference and `std::list` is better, we should profile. Until then, I don't think it's worth changing the "default."

rir/src/runtime/Code.h
JanJecmen: Hehe didn't want to mention this, but it also crossed my mind with regard to `[[nodiscard]]`... Also `constexpr if` could be nice for the instructions

rir/src/runtime/Code.cpp
o-: ah yeah, thanks. this used to be C code

o-:
@JanJecmen 
> One thing, you mention that you found a bunch of decode uses but none are changed in this pr?

https://github.com/reactorlabs/rir/pull/181/commits/20b945ced16748335730c6be3bb4f4f75b6c2d80

(advance decodes using BC objects)

JanJecmen:
Oh damn sorry, missed that

rir/src/runtime/Code.h
```diff
@@ -89,13 +92,18 @@ struct Code {
...
      *
      */
 
+    // The source list contains pcOffset to src index
+    struct SrclistEntry {
+        unsigned pcOffset;
+        unsigned srcIdx;
+    };
+    static_assert(sizeof(SrclistEntry) == 2 * sizeof(unsigned),
+                  "Needs to correspond to the size of the record_call_ BC");
```
mhyee, line 105: Wrong assert message.

rir/src/runtime/Code.h
```diff
@@ -89,13 +92,18 @@ struct Code {
      *   ---------------------------------------------------------------------
      *   code stream   BC                pad4(codeSize)
      *
-     *   skiplist      (instr offset,    2 * skiplistLength * sizeof(unsigned)
-     *                  src index)
-     *
-     *   srcList       cp_idx (ast)      srcLength * sizeof(unsigned)
+     *   srcList       cp_idx (ast)      srcLength * sizeof(SrcEntry)
```
mhyee, line 95: Should be `sizeof(SrclistEntry)`.

rir/src/runtime/Code.h
o-: actually this is not even needed anymore in this version, where we use an actual array, to index the sourcelist entries

PR #182, opened 2018-08-08
PIR sources tracking

JanJecmen:


o-:
hey, thanks for tracking this down. lgtm.

I have one concern. I would like to have this a bit more abstract. For example we could have a template parameter `NeedsSources` in InstructionImplementation. Then you can add in the constructor a `static_assert(NeedsSources == NeedsSources::Required ? srcIdx != 0 : true)` and `static_assert(NeedsSources == NeedsSources::Unnecessary ? srcIdx == 0 : true)`

but please merge this pr as it is now. In my pr #181 i came up with a list for which BC need sources, which ones don't and which ones we don't care. So maybe we can build something on top of those two prs.

rir/src/compiler/pir/instruction.h
```diff
@@ -747,9 +749,9 @@ BINOP(Eq, RType::logical);
 #define BINOP_NOENV(Name, Type)                                                \
     class FLI(Name, 2, Effect::None, EnvAccess::None) {                        \
       public:                                                                  \
-        Name(Value* lhs, Value* rhs, unsigned srcIdx)                          \
```
o-: this should not have sources. this only works on integers!

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
```diff
@@ -475,6 +488,15 @@ bool StackMachine::tryRunCurrentBC(const Rir2Pir& rir2pir, Builder& insert) {
     case Opcode::ldddvar_:
         return false;
     }
+
+    // TODO: change to assert
+    // assert(srcIdx == 0 && "source index is getting lost in translation");
+    if (srcIdx)
```
o-, line 494: where does this happen currently?

JanJecmen:
Ok great, I agree that we need something a bit more thorough, this leaves everything to the programmer to remember and get right

rir/src/compiler/translations/rir_2_pir/stack_machine.cpp
JanJecmen: Sorry, missed the comments
right now, this happens with a variety of bytecodes, eg. push_ or guard_env_ or anything random, really.. that is caused by the broken lookup in the skiplist which you fixed yesterday, so after merging it all it should never happen

o-: ah, i see

PR #183, opened 2018-08-09
Reintroduce recording of function targets

o-:
based on top of #181

### DIFF
--- a/rir/src/ir/Compiler.cpp
+++ b/rir/src/ir/Compiler.cpp
@@ -1077,10 +1077,11 @@ void compileCall(Context& ctx, SEXP ast, SEXP fun, SEXP args) {
     }
     assert(callArgs.size() < BC::MAX_NUM_ARGS);
 
-    if (hasNames)
+    if (hasNames) {
         cs << BC::callImplicit(callArgs, names, ast);
-    else
-        cs << BC::callImplicit(callArgs, ast);
+    } else {
+        cs << BC::recordCall() << BC::callImplicit(callArgs, ast);
+    }
 }
 
 // Lookup
--- a/rir/src/ir/RuntimeFeedback.h
+++ b/rir/src/ir/RuntimeFeedback.h
@@ -1,0 +1,43 @@
+#ifndef RIR_RUNTIME_FEEDBACK
+#define RIR_RUNTIME_FEEDBACK
+
+#include "R/r.h"
+#include <cstdint>
+
+namespace rir {
+
+#pragma pack(push)
+#pragma pack(1)
+
+struct CallFeedback {
+    static constexpr unsigned CounterBits = 30;
+    static constexpr unsigned CounterOverflow = (1 << CounterBits) - 1;
+    static constexpr unsigned TargetBits = 2;
+    static constexpr unsigned MaxTargets = (1 << TargetBits) - 1;
+
+    uint32_t taken : CounterBits;
+    uint32_t numTargets : TargetBits;
+    SEXP targets[MaxTargets];
+
+    void record(SEXP callee) {
+        if (taken < CounterOverflow)
+            taken++;
+        if (numTargets < MaxTargets) {
+            int i = 0;
+            for (; i < numTargets; ++i)
+                if (targets[i] == callee)
+                    break;
+            if (i == numTargets)
+                targets[numTargets++] = callee;
+        }
+    }
+};
+
+static_assert(sizeof(CallFeedback) == 7 * sizeof(uint32_t),
+              "Size needs to fit inside a record_ bc immediate args");
+
+#pragma pack(pop)
+
+} // namespace rir
+
+#endif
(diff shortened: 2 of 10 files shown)


rir/src/ir/Compiler.cpp
```diff
@@ -1077,10 +1077,11 @@ void compileCall(Context& ctx, SEXP ast, SEXP fun, SEXP args) {
     }
     assert(callArgs.size() < BC::MAX_NUM_ARGS);
 
-    if (hasNames)
+    if (hasNames) {
```
mhyee, line 1080: Do we have an agreed convention for braces? I've seen braced and braceless if-statements, but never commented on them. (I prefer always having braces.)

Unfortunately, it looks like clang-format can't enforce this.

rir/src/ir/RuntimeFeedback.h
```diff
@@ -0,0 +1,43 @@
...
+namespace rir {
+
+#pragma pack(push)
+#pragma pack(1)
+
+struct CallFeedback {
+    static constexpr unsigned CounterBits = 30;
+    static constexpr unsigned CounterOverflow = (1 << CounterBits) - 1;
+    static constexpr unsigned TargetBits = 2;
+    static constexpr unsigned MaxTargets = (1 << TargetBits) - 1;
```
mhyee, line 16: I think we're off-by-one here... with 2 bits, we can represent 4 targets. But then the condition below (line 25) will need to be more complicated...

Actually, the problem is on line 31, where we do not want to increment `numTargets` if its value is 4 (assuming we have 2 target bits).

(Or maybe I'm completely wrong. I hate bit twiddling and off-by-one errors.)

rir/src/ir/RuntimeFeedback.h
```diff
@@ -0,0 +1,43 @@
...
+            for (; i < numTargets; ++i)
+                if (targets[i] == callee)
+                    break;
+            if (i == numTargets)
+                targets[numTargets++] = callee;
+        }
+    }
+};
+
+static_assert(sizeof(CallFeedback) == 7 * sizeof(uint32_t),
```
mhyee, line 36: Add a comment for where `7` comes from, in case it needs to be modified?

rir/src/ir/insns.h
```diff
@@ -438,4 +438,9 @@ DEF_INSTR(ret_, 0, 1, 0, 1)
  */
 DEF_INSTR(int3_, 0, 0, 0, 1)
 
+/*
+ * recording bytecodes are used to collect information
```
mhyee, line 442: Add some documentation about how this instruction connects to `CallFeedback`? Specifically, it's a bit weird seeing that this single instruction needs 7 immediate args.

rir/src/ir/RuntimeFeedback.h
mhyee: Yeah, I think there's some conflation in how `numTargets` is being used as an index into the array, and so `MaxTargets` is actually `MaxTargetIndex`.

mhyee (commented):
I don't really remember how the old target recording worked, but from my first impressions, this approach looks pretty good.

I guess my question is how frequently will `record_call_` be emitted? Seven immediates seems like quite a bit. And if start recording types, that's also more stuff we're shoving into the bytecode stream.

Then again, this is probably simpler than keeping stuff separate from the bytecode stream. (I guess that's what we did before?)

rir/src/ir/RuntimeFeedback.h
o-: I am not sure if I understand. I think it's actually correct. In this concrete case we have 2 target bits, so MaxTargets is 3.  If we find a target, then its stored at targets[0] and numTargets is increased to 1. If we find a second target, we store at 1 and increase to 2. Finally if we find a third target, we store at 2 and increase to MaxTargets. From now on we will no longer store more targets. What we lost, is the information if we had exactly 3 targets or more. We could steal one bit from the counter, if we think it is important to know that..

rir/src/ir/RuntimeFeedback.h
o-: i documented it in the other PR no

o-:
hey, I agree with the general sentiment, that its kind of expensive to store this in the BC stream... on the other hand it is so much more convenient. We used to have it dangling at the end of the Code objects and it was a mess. another nice thing is: if we don't do it this way (with dedicated instructions) then every instruction which potentially wants to record feedback, needs a bit telling it, if it should record or not (because presumably we only want to record when it makes sense...)

rir/src/ir/RuntimeFeedback.h
mhyee: So I think the code works exactly as you described. When I first looked at this code, my thought was "Well, with 2 bits to record the target, we should be able to store 4 targets, but right now you're only storing up to 3."

After thinking some more, I think it's fine the way it is (but maybe with some explanation so I don't run into this and get confused again). Storing one more target probably isn't super helpful, and the logic would become far more complicated. (Being able to index "one past the last element" is so useful...)

mhyee:
> if we don't do it this way (with dedicated instructions) then every instruction which potentially wants to record feedback, needs a bit telling it, if it should record or not (because presumably we only want to record when it makes sense...)

Yeah, having the dedicated instructions seems like a better approach to me.

PR #184, opened 2018-08-10
Add env leaks to force_

JanJecmen:


rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -99,7 +99,9 @@ class TheScopeResolution {
                         // This load can be resolved to a unique value
                         aval.ifSingleValue([&](Value* val) {
                             if (ldfun) {
-                                replaceLdFun(val);
+                                // The force can leak the environment where the ldfun
+                                // we are replacing would be executed
+                                replaceLdFun(val, ldfun->env());
```
JanJecmen: Still not sure `ldfun->env()` is the correct one... We want the one where the `ldfun` instruction is executed, this gives us the one where it does the lookup. Do we know they are always the same?

JanJecmen:
Also, this is still / now broken. It seems like it has to do with the same, but not sure yet.
```
showCols2 <- function(bg = "grey", cex = 0.75, rot = 30) {
    m <- ceiling(sqrt(n <- length(cl <- colors())))
    length(cl) <- m*m; cm <- matrix(cl, m)
    ##
    require("grid")
    grid.newpage(); vp <- viewport(w = .92, h = .92)
    grid.rect(gp=gpar(fill=bg))
    grid.text(cm, x = col(cm)/m, y = rev(row(cm))/m, rot = rot,
              vp=vp, gp=gpar(cex = cex, col = cm))
}
showCols2()
```

o-:
@JanJecmen thanks. I extracted two minimal  error cases. This is the one you saw:

```
validGP <- function(gpars) {
  numnotnull <- function(gparname) {
    match(gparname, names(gpars))
    gpars[[gparname]]
  }
  numnotnull('a')
  numnotnull('c')
}

(function() {
 validGP(list(a=1))
})()
```

which complains about `gparname`

and I also stumbled on
```
validGP <- function(gpars) {
  check.length <- function(gparname) {
      NULL
  }
  numnotnull <- function(gparname) {
    if (match(gparname, names(gpars))) {
        check.length(gparname)
    }   
  }
  numnotnull('a')
}

(function() {
   validGP(list(a=1))
})()
```

which doesn't find the function `check.length`.

With those it should be easy enough to figure out what's going on (didn't do that yet).

JanJecmen:
Cool, will have a look at that. I found the matching but this minimal example seems to work:
```
foo <- function(x) {
    bar <- function(y) {
        is.na(match(y, names(x)))
    }
    bar('aa')
}
foo(list(aa = 7))
```

As for the pir tests, I just commented the ones that failed so that I could share the changes. I think mostly they fail because they `CHECK(Query::noEnv())`

o-:
lgtm! thanks

PR #188, opened 2018-08-14
Add a Logger in charge of managing the logging of the different (inne…

charig:
…r) functions in different streams

### DIFF
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -41,7 +41,7 @@ void StreamLogger::compilationInit() const {
     }
 }
 
-void StreamLogger::compilationEarlyPir(Closure& closure) const {
+void StreamLogger::compilationEarlyPir(const Closure& closure) const {
     if (options.includes(DebugFlag::PrintEarlyPir)) {
         innerHeader(" Compiled to PIR Version ");
         closure.print(getLog());
@@ -70,14 +70,14 @@ void StreamLogger::rirFromPir(rir::Function* function) const {
     }
 }
 
-void StreamLogger::phiInsertion(Code* code) const {
+void StreamLogger::phiInsertion(const Code* code) const {
     if (options.includes(DebugFlag::PrintCSSA)) {
         innerHeader(" After Phi Copies Inserted ");
         code->print(getLog());
     }
 }
 
-void StreamLogger::afterCSSA(Code* code) const {
+void StreamLogger::afterCSSA(const Code* code) const {
     if (options.includes(DebugFlag::PrintCSSA)) {
         innerHeader(" PIR After Converting to CSSA ");
         code->print(getLog());
@@ -89,14 +89,15 @@ void StreamLogger::afterCSSA(Code* code) const {
             allocator.print(getLog());
 }*/
 
-void StreamLogger::finalPIR(Code* code) const {
+void StreamLogger::finalPIR(const Code* code) const {
     if (options.includes(DebugFlag::PrintFinalPir)) {
         innerHeader(" Final PIR Version ");
         code->print(getLog());
     }
 }
 
-void StreamLogger::finish(rir::Function* function, std::ostream& stream) const {
+void StreamLogger::finish(const rir::Function* function,
+                          std::ostream& stream) const {
     if (options.intersects(PrintDebugPasses)) {
         header("Finished compiling:", function, stream, false);
     }
@@ -120,7 +121,7 @@ void StreamLogger::failCompilingPir() const {
                  << &currentFunction.top() << "\n";
 }
 
-void StreamLogger::header(std::string header, rir::Function* function,
+void StreamLogger::header(std::string header, const rir::Function* function,
                           std::ostream& stream, bool opening) const {
     std::stringstream ss;
     ss << " " << header << " " << function << " ";
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -8,9 +8,18 @@ namespace pir {
 void StreamLogger::startLogging(rir::Function* function) {
     if (options.intersects(PrintDebugPasses)) {
         if (!streams.count(function)) {
-            streams.emplace(function, new std::stringstream());
+            std::ostream* o;
+            if (options.includes(DebugFlag::PrintIntoFiles)) {
+                // o = new std::ofstream();
+                //((std::ofstream*)o)->open(".pir/pirCompilation.data");
+            } else if (options.includes(DebugFlag::PrintIntoStdout)) {
+                o = &std::cout;
+            } else {
+                o = new std::stringstream();
+            }
+            streams.emplace(function, o);
+            currentFunction.push(function);
         }
-        currentFunction.push(function);
     }
 }
 
@@ -87,8 +96,7 @@ void StreamLogger::finalPIR(Code* code) const {
     }
 }
 
-void StreamLogger::finish(rir::Function* function,
-                          std::stringstream& stream) const {
+void StreamLogger::finish(rir::Function* function, std::ostream& stream) const {
     if (options.intersects(PrintDebugPasses)) {
         header("Finished compiling:", function, stream, false);
     }
@@ -113,7 +121,7 @@ void StreamLogger::failCompilingPir() const {
 }
 
 void StreamLogger::header(std::string header, rir::Function* function,
-                          std::stringstream& stream, bool opening) const {
+                          std::ostream& stream, bool opening) const {
     std::stringstream ss;
     ss << " " << header << " " << function << " ";
     stream << std::setfill('*') << std::left;
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -64,10 +64,10 @@ void StreamLogger::pirOptimizations(const Closure& closure,
     }
 }
 
-void StreamLogger::rirFromPir(rir::Function* function) const {
+void StreamLogger::rirFromPir(const rir::Function* function) const {
     if (options.includes(DebugFlag::PrintFinalRir)) {
         innerHeader(" Final RIR Version ");
-        for (auto code : *function) {
+        for (const auto code : *function) {
             code->print(getLog());
         }
         getLog() << "\n";
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -1,5 +1,8 @@
 #include "stream_logger.h"
-#include "../../utils/Pool.h"
+#include "../pir/pir_impl.h"
+#include "runtime/Function.h"
+#include "utils/Pool.h"
+
 #include <iomanip>
 
 namespace rir {
@@ -10,8 +13,9 @@ void StreamLogger::startLogging(rir::Function* function) {
         if (!streams.count(function)) {
             std::ostream* o;
             if (options.includes(DebugFlag::PrintIntoFiles)) {
-                // o = new std::ofstream();
-                //((std::ofstream*)o)->open(".pir/pirCompilation.data");
+                o = new std::ofstream();
+                // TODO: shouldn't the function identifier be part of the name?
+                ((std::ofstream*)o)->open(".pir/pirCompilation.data");
             } else if (options.includes(DebugFlag::PrintIntoStdout)) {
                 o = &std::cout;
             } else {
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -8,8 +8,7 @@ namespace pir {
 void StreamLogger::startLogging(rir::Function* function) {
     if (options.intersects(PrintDebugPasses)) {
         if (!streams.count(function)) {
-            streams.insert(std::pair<rir::Function*, std::stringstream*>(
-                function, new std::stringstream()));
+            streams.emplace(function, new std::stringstream());
         }
         currentFunction.push(function);
     }
@@ -20,63 +19,58 @@ void StreamLogger::endLogging() {
         currentFunction.pop();
 }
 
-void StreamLogger::compilationInit() {
+void StreamLogger::compilationInit() const {
     if (options.intersects(PrintDebugPasses)) {
-        getLog() << "\n***********************************************"
-                 << "***************\n";
-        getLog() << "*********** Start compiling: " << std::setw(20)
-                 << std::left << currentFunction.top() << " ************\n\n";
+        this->header("Start compiling:", currentFunction.top(), getLog(), true);
         if (options.includes(DebugFlag::PrintEarlyRir)) {
-            getLog() << "=============== Original version:" << std::setw(16)
-                     << std::left << " ==========\n\n";
-            auto it = currentFunction.top()->begin();
-            while (it != currentFunction.top()->end()) {
-                (*it)->print(getLog());
-                ++it;
+            innerHeader(" Original version ");
+            for (auto code : *(currentFunction.top())) {
+                code->print(getLog());
             }
             getLog() << "\n";
         }
     }
 }
 
-void StreamLogger::compilationEarlyPir(Closure& closure) {
+void StreamLogger::compilationEarlyPir(Closure& closure) const {
     if (options.includes(DebugFlag::PrintEarlyPir)) {
-        getLog() << "\n========= Compiled to PIR Version" << std::setw(16)
-                 << std::left << " ==========\n";
+        innerHeader(" Compiled to PIR Version ");
         closure.print(getLog());
     }
 }
 
-void StreamLogger::pirOptimizations(Closure& closure,
+void StreamLogger::pirOptimizations(const Closure& closure,
                                     const std::string& category,
-                                    const std::string& pass, size_t passnr) {
-    getLog() << "============== " << category << ": " << pass
-             << " == " << passnr << std::setw(10) << std::left
-             << " ======================\n";
-    closure.print(getLog());
-}
-
-void StreamLogger::phiInsertion(Closure& closure) {
-    getLog() << "========= After phi copies inserted" << std::setw(14)
-             << std::left << " ==========\n";
-    closure.print(getLog());
-}
-
-void StreamLogger::rirFromPir(rir::Function* function) {
-    if (options.includes(DebugFlag::PrintFinalRir)) {
-        getLog() << "\n============= Final RIR Version" << std::setw(16)
-                 << std::left << " ==========\n";
-        auto it = function->begin();
-        while (it != function->end()) {
-            (*it)->print(getLog());
-            ++it;
-        }
+                                    const std::string& pass,
+                                    size_t passnr) const {
+    if (options.includes(DebugFlag::PrintOptimizationPasses)) {
+        std::stringstream ss;
+        ss << category << ": " << pass << " == " << passnr;
+        innerHeader(ss.str());
+        closure.print(getLog());
     }
 }
 
-void StreamLogger::afterCSSA(Code* code) {
+void StreamLogger::rirFromPir(rir::Function* function) const {
+    if (options.includes(DebugFlag::PrintFinalRir)) {
+        innerHeader(" Final RIR Version ");
+        for (auto code : *function) {
+            code->print(getLog());
+        }
+        getLog() << "\n";
+    }
+}
+
+void StreamLogger::phiInsertion(Code* code) const {
     if (options.includes(DebugFlag::PrintCSSA)) {
-        getLog() << " ========== PIR After Converting to CSSA\n";
+        innerHeader(" After Phi Copies Inserted ");
+        code->print(getLog());
+    }
+}
+
+void StreamLogger::afterCSSA(Code* code) const {
+    if (options.includes(DebugFlag::PrintCSSA)) {
+        innerHeader(" PIR After Converting to CSSA ");
         code->print(getLog());
     }
 }
@@ -86,26 +80,24 @@ void StreamLogger::afterCSSA(Code* code) {
             allocator.print(getLog());
 }*/
 
-void StreamLogger::finalPIR(Code* code) {
+void StreamLogger::finalPIR(Code* code) const {
     if (options.includes(DebugFlag::PrintFinalPir)) {
-        getLog() << "\n ========== Final PIR Version\n";
+        innerHeader(" Final PIR Version ");
         code->print(getLog());
     }
 }
 
-void StreamLogger::finish(rir::Function* function, std::stringstream& stream) {
+void StreamLogger::finish(rir::Function* function,
+                          std::stringstream& stream) const {
     if (options.intersects(PrintDebugPasses)) {
-        stream << "\n*********** Finished compiling: " << std::setw(17)
-               << std::left << function << " ************\n";
-        stream << "*************************************************"
-               << "*************\n";
+        header("Finished compiling:", function, stream, false);
     }
 }
 
-void StreamLogger::warningBC(std::string warning, rir::BC bc) {
+void StreamLogger::warningBC(std::string warning, rir::BC bc) const {
     if (options.includes(DebugFlag::ShowWarnings)) {
         getLog() << "Warning" << warning << ": ";
-        if (warning.compare("Guard ignored") == 0) {
+        if (warning.compare(WARNING_GUARD_STRING) == 0) {
             CHAR(PRINTNAME(rir::Pool::get(bc.immediate.guard_fun_args.name)));
         } else {
             bc.print(getLog());
@@ -114,11 +106,35 @@ void StreamLogger::warningBC(std::string warning, rir::BC bc) {
     }
 }
 
-void StreamLogger::failCompilingPir() {
+void StreamLogger::failCompilingPir() const {
     if (options.includes(DebugFlag::ShowWarnings))
         getLog() << " Failed verification after p2r compile "
                  << &currentFunction.top() << "\n";
 }
 
+void StreamLogger::header(std::string header, rir::Function* function,
+                          std::stringstream& stream, bool opening) const {
+    std::stringstream ss;
+    ss << " " << header << " " << function << " ";
+    stream << std::setfill('*') << std::left;
+    if (opening)
+        stream << "\n"
+               << std::setw(70) << ""
+               << "\n";
+    else
+        stream << "\n";
+    stream << std::left << std::setw(21) << "" << std::setw(49) << ss.str()
+           << "\n";
+    if (!opening)
+        stream << std::setw(70) << ""
+               << "\n";
+    stream << std::setfill(' ') << "\n";
+}
+
+void StreamLogger::innerHeader(std::string header) const {
+    getLog() << "\n"
+             << std::setfill('=') << std::setw(15) << "" << std::left
+             << std::setw(39) << header << std::setfill(' ') << "\n";
+}
 } // namespace pir
 } // namespace rir
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -1,4 +1,5 @@
 #include "stream_logger.h"
+#include "../../utils/Pool.h"
 #include <iomanip>
 
 namespace rir {
@@ -101,10 +102,14 @@ void StreamLogger::finish(rir::Function* function, std::stringstream& stream) {
     }
 }
 
-void StreamLogger::warningBC(const char* warning, rir::BC bc) {
+void StreamLogger::warningBC(std::string warning, rir::BC bc) {
     if (options.includes(DebugFlag::ShowWarnings)) {
         getLog() << "Warning" << warning << ": ";
-        bc.print(getLog());
+        if (warning.compare("Guard ignored") == 0) {
+            CHAR(PRINTNAME(rir::Pool::get(bc.immediate.guard_fun_args.name)));
+        } else {
+            bc.print(getLog());
+        }
         getLog() << "\n";
     }
 }
--- a/rir/src/compiler/debugging/stream_logger.cpp
+++ b/rir/src/compiler/debugging/stream_logger.cpp
@@ -1,0 +1,119 @@
+#include "stream_logger.h"
+#include <iomanip>
+
+namespace rir {
+namespace pir {
+
+void StreamLogger::startLogging(rir::Function* function) {
+    if (options.intersects(PrintDebugPasses)) {
+        if (!streams.count(function)) {
+            streams.insert(std::pair<rir::Function*, std::stringstream*>(
+                function, new std::stringstream()));
+        }
+        currentFunction.push(function);
+    }
+}
+
+void StreamLogger::endLogging() {
+    if (options.intersects(PrintDebugPasses))
+        currentFunction.pop();
+}
+
+void StreamLogger::compilationInit() {
+    if (options.intersects(PrintDebugPasses)) {
+        getLog() << "\n***********************************************"
+                 << "***************\n";
+        getLog() << "*********** Start compiling: " << std::setw(20)
+                 << std::left << currentFunction.top() << " ************\n\n";
+        if (options.includes(DebugFlag::PrintEarlyRir)) {
+            getLog() << "=============== Original version:" << std::setw(16)
+                     << std::left << " ==========\n\n";
+            auto it = currentFunction.top()->begin();
+            while (it != currentFunction.top()->end()) {
+                (*it)->print(getLog());
+                ++it;
+            }
+            getLog() << "\n";
+        }
+    }
+}
+
+void StreamLogger::compilationEarlyPir(Closure& closure) {
+    if (options.includes(DebugFlag::PrintEarlyPir)) {
+        getLog() << "\n========= Compiled to PIR Version" << std::setw(16)
+                 << std::left << " ==========\n";
+        closure.print(getLog());
+    }
+}
+
+void StreamLogger::pirOptimizations(Closure& closure,
+                                    const std::string& category,
+                                    const std::string& pass, size_t passnr) {
+    getLog() << "============== " << category << ": " << pass
+             << " == " << passnr << std::setw(10) << std::left
+             << " ======================\n";
+    closure.print(getLog());
+}
+
+void StreamLogger::phiInsertion(Closure& closure) {
+    getLog() << "========= After phi copies inserted" << std::setw(14)
+             << std::left << " ==========\n";
+    closure.print(getLog());
+}
+
+void StreamLogger::rirFromPir(rir::Function* function) {
+    if (options.includes(DebugFlag::PrintFinalRir)) {
+        getLog() << "\n============= Final RIR Version" << std::setw(16)
+                 << std::left << " ==========\n";
+        auto it = function->begin();
+        while (it != function->end()) {
+            (*it)->print(getLog());
+            ++it;
+        }
+    }
+}
+
+void StreamLogger::afterCSSA(Code* code) {
+    if (options.includes(DebugFlag::PrintCSSA)) {
+        getLog() << " ========== PIR After Converting to CSSA\n";
+        code->print(getLog());
+    }
+}
+
+/*void StreamLogger::afterLiveness(SSAAllocator& allocator) {
+    if (options.includes(DebugFlag::PrintLivenessIntervals))
+            allocator.print(getLog());
+}*/
+
+void StreamLogger::finalPIR(Code* code) {
+    if (options.includes(DebugFlag::PrintFinalPir)) {
+        getLog() << "\n ========== Final PIR Version\n";
+        code->print(getLog());
+    }
+}
+
+void StreamLogger::finish(rir::Function* function, std::stringstream& stream) {
+    if (options.intersects(PrintDebugPasses)) {
+        stream << "\n*********** Finished compiling: " << std::setw(17)
+               << std::left << function << " ************\n";
+        stream << "*************************************************"
+               << "*************\n";
+    }
+}
+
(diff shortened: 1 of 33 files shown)


rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
         BC bc = BC::decode(pc);
 
-        Rprintf(" %5d ", ((uintptr_t)pc - (uintptr_t)code()));
+        out << " %5d " << ((uintptr_t)pc - (uintptr_t)code());
 
         unsigned s = getSrcIdxAt(pc, true);
         if (s != 0)
-            Rprintf("   ; %s\n       ",
-                    dumpSexp(src_pool_at(globalContext(), s)).c_str());
+            out << "   ; " << dumpSexp(src_pool_at(globalContext(), s)).c_str()
```
JanJecmen: `.c_str()` is not needed, ostreams can handle std::strings

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
 
-void Code::disassemble() {
+void Code::disassemble(std::ostream& out) {
     Opcode* pc = code();
 
     while (pc < endCode()) {
         BC bc = BC::decode(pc);
 
-        Rprintf(" %5d ", ((uintptr_t)pc - (uintptr_t)code()));
+        out << " %5d " << ((uintptr_t)pc - (uintptr_t)code());
```
JanJecmen: Here you left the format specifier so it will just get sent to the stream as is

JanJecmen: in c++ you need to `#inlcude <iomanip>` and then use `setw` and such for the formatting

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
 
-void Code::print() {
-    Rprintf("Code object (%p offset %x (hex))\n", this, header);
-    Rprintf("  Source:      %u (index to src pool)\n", src);
-    Rprintf("  Magic:       %x (hex)\n", magic);
-    Rprintf("  Stack (o):   %u\n", stackLength);
-    Rprintf("  Code size:   %u [B]\n", codeSize);
-    Rprintf("  Default arg? %s\n", isDefaultArgument ? "yes" : "no");
+void Code::print(std::ostream& out) {
+    out << "Code object (" << this << "offset" << std::hex << header << "(hex))"
```
JanJecmen: spaces: `out << "Code object (" << this << " offset " << std::hex << header << " (hex))"`

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
+        << "\n";
+    out << "Source: " << src << " index to src pool\n";
+    out << "Magic: " << std::hex << magic << "(hex)\n";
+    out << "Stack (o): " << stackLength << "\n";
+    out << "Code size: " << codeSize << "[B]\n";
```
JanJecmen: here indentation is missing

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
+
+    out << "Default arg? ";
+    isDefaultArgument ? out << "yes\n" : out << "no\n";
```
JanJecmen: you can do the ternary `?:` inline: `out << "Default arg? " << (isDefaultArgument ? "yes" : "no") << '\n';`

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +11,57 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
     if (magic != CODE_MAGIC)
-        Rf_error("Wrong magic number -- corrupted IR bytecode");
+        out << "Wrong magic number -- corrupted IR bytecode";
```
JanJecmen: I think this still needs to abort the printing, because `Rf_error` doesn't return, so the disassemble below will segfault

rir/src/ir/BC.cpp
```diff
@@ -179,76 +179,74 @@ void BC::write(CodeStream& cs) const {
...
     for (auto arg : immediateCallArguments) {
         if (arg == MISSING_ARG_IDX)
-            Rprintf(" _");
+            out << " _";
         else if (arg == DOTS_ARG_IDX)
-            Rprintf(" ...");
+            out << " ...";
         else
-            Rprintf(" %x", arg);
+            out << std::hex << arg;
```
JanJecmen: The `hex` changes the flag in the stream, so it will affect all future output. Restore to `dec` after the `arg` (I assume decimal is what is set before)

rir/src/ir/BC.cpp
```diff
@@ -260,37 +258,37 @@ void BC::print() const {
     case Opcode::stvar_:
     case Opcode::stvar_super_:
     case Opcode::missing_:
-        Rprintf(" %s", CHAR(PRINTNAME((immediateConst()))));
+        out << " " << CHAR(PRINTNAME((immediateConst())));
```
JanJecmen: too many parentheses :) `CHAR(PRINTNAME(immediateConst()))`

rir/src/ir/BC.cpp
```diff
@@ -260,37 +258,37 @@ void BC::print() const {
...
+        out << " " << immediate.i;
         break;
     case Opcode::ldarg_:
-        Rprintf(" %u", immediate.arg_idx);
+        out << " " << immediate.arg_idx;
         break;
     case Opcode::ldloc_:
     case Opcode::stloc_:
-        Rprintf(" @%i", immediate.loc);
+        out << " " << immediate.loc;
```
JanJecmen: `out << " @" << immediate.loc;`

rir/src/ir/BC.cpp
```diff
@@ -260,37 +258,37 @@ void BC::print() const {
...
         break;
     case Opcode::movloc_:
-        Rprintf(" @%i -> @%i", immediate.loc_cpy.source,
-                immediate.loc_cpy.target);
+        out << " " << immediate.loc_cpy.source << " -> "
```
JanJecmen: `out << " @" << immediate.loc_cpy.source << " -> @" << immediate.loc_cpy.target;`

rir/src/ir/BC.cpp
```diff
@@ -390,19 +388,19 @@ void BC::print() const {
         break;
     case Opcode::promise_:
     case Opcode::push_code_:
-        Rprintf(" %x", immediate.fun);
+        out << " " << std::hex << immediate.fun;
```
JanJecmen: Again, restore decimal

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -14,23 +14,20 @@
...
-#include <iostream>
-
 #include "interpreter/runtime.h"
 
 namespace rir {
 namespace pir {
 
 Rir2PirCompiler::Rir2PirCompiler(Module* module, const DebugOptions& debug)
-    : RirCompiler(module, debug) {
+    : RirCompiler(module, debug), log(StreamLogger(debug)) {
```
JanJecmen: I believe you meant to just pass `debug` to the logger constructor, but what happens here is you create a temporary `StreamLogger` object and then pass that to the constructor for `log`, so a copy / move constructor is invoked.
You probably want just `log(debug)`

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -42,71 +39,57 @@ void Rir2PirCompiler::compileClosure(SEXP closure, MaybeCls success,
...
+            log.failCompilingPir();
             failed = true;
             return false;
         });
 
     if (failed)
         fail();
-    else
+    else {
+        module->get(srcFunction)->print(std::cout);
```
JanJecmen: Shouldn't this also be in the logger? And only printed conditionally?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -1008,10 +1003,7 @@ void Pir2Rir::toCSSA(Code* code) {
         }
     });
 
-    DEBUGCODE(PHI_REMOVE_DEBUG, {
-        std::cout << "--- phi copies inserted ---\n";
-        code->print(std::cout);
-    });
+    DEBUGCODE(PHI_REMOVE_DEBUG, { log.phiInsertion(code); });
```
JanJecmen: Maybe the DEBUGCODE we can now get rid of..

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace rir {
+namespace pir {
+
+class StreamLogger {
+  public:
+    StreamLogger(DebugOptions options) : options(options) {}
+    ~StreamLogger() {
+        std::cout << "Cantidad de streams: " << streams.size() << "\n\n";
+        for (std::map<rir::Function*, std::stringstream*>::iterator it =
```
JanJecmen: this is what `auto` is for ;)
`for (auto it = streams.begin(); it != streams.end(); ++it) ...`
Or better yet, `for (auto p : streams) ...`

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,64 @@
...
+//#include <fstream>
+
+namespace rir {
+namespace pir {
+
+class StreamLogger {
+  public:
+    StreamLogger(DebugOptions options) : options(options) {}
+    ~StreamLogger() {
+        std::cout << "Cantidad de streams: " << streams.size() << "\n\n";
```
JanJecmen: Slowly switching to Spanish, eh? :)

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,64 @@
...
+        for (std::map<rir::Function*, std::stringstream*>::iterator it =
+                 streams.begin();
+             it != streams.end(); ++it) {
+            this->finish(it->first, *streams.at(it->first));
```
JanJecmen: `this->` is not needed here

JanJecmen (commented):
One thought, do we want this to be present in release mode? maybe we should do conditional compilation only for debug mode?

rir/src/interpreter/deoptimizer.cpp
```diff
@@ -61,7 +61,7 @@ static DeoptInfo* DeoptInfo_get(uint32_t idx) {
     return (DeoptInfo*)((char*)deoptTable + off);
 }
 
-void Deoptimizer_print(uint32_t idx) {
+void Deoptimizer_print(uint32_t idx, std::ostream& out) {
```
mhyee, line 64: Is the `ostream` used anywhere?

rir/src/interpreter/runtime.cpp
```diff
@@ -56,8 +58,10 @@ void printFunction(Function* f) {
         Rf_error("Wrong magic number -- not rir bytecode");
 
     // print respective code objects
+    std::stringstream output;
```
mhyee, line 61: So we do have `capture_out.h`, but I guess it's more of a hack for cases when we don't have any other way of controlling the output stream...

rir/src/ir/BC_inc.h
```diff
@@ -206,10 +206,10 @@ class BC {
     void write(CodeStream& cs) const;
 
     // Print it to stdout
```
mhyee: Comment needs updating.

rir/src/runtime/Code.cpp
mhyee: Alignment is also missing.

rir/src/runtime/Code.cpp
mhyee: Actually, this could get a bit messy. If `out` is a string stream and we just abort, then we lose everything. If it's standard out, then we don't want to print everything twice...

I guess the simplest fix is to throw away everything before this point, and then write to standard error and abort. But it'll be kind of annoying to debug.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+void StreamLogger::finish(rir::Function* function, std::stringstream& stream) {
+    if (options.intersects(PrintDebugPasses)) {
+        stream << "\n*********** Finished compiling: " << std::setw(17)
+               << std::left << function << " ************\n";
+        stream << "*************************************************"
+               << "*************\n";
+    }
+}
+
+void StreamLogger::warningBC(std::string warning, rir::BC bc) {
```
mhyee: Minor, but I think I prefer flipping the order of arguments?

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+    if (options.includes(DebugFlag::ShowWarnings)) {
+        getLog() << "Warning" << warning << ": ";
+        if (warning.compare("Guard ignored") == 0) {
```
mhyee: I don't like how we have to compare strings here; it's very brittle. A slight improvement might be to just pass a flag as an argument? At least that way it's more explicit.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -620,17 +620,12 @@ class Pir2Rir {
...
 
     SSAAllocator alloc(code,
                        compiler.debug.includes(DebugFlag::DebugAllocator));
 
-    if (compiler.debug.includes(DebugFlag::PrintLivenessIntervals))
-        alloc.print();
-
-    if (compiler.debug.includes(DebugFlag::PrintFinalPir))
-        code->print(std::cout);
+    // compiler.getLog().afterLiveness(&alloc);
```
mhyee: Why is this commented out?

rir/src/compiler/translations/pir_2_rir.cpp
mhyee: Yeah, if all this logging will be conditionally compiled, we should probably merge it with DEBUGCODE.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
```diff
@@ -11,17 +13,22 @@ class Rir2PirCompiler : public RirCompiler {
...
-    void compileClosure(SEXP, MaybeCls success, Maybe fail) override;
+    void compileClosure(SEXP, MaybeCls success, Maybe fail,
+                        bool isIndependent) override;
     void compileFunction(rir::Function*, FormalArgs const&, MaybeCls success,
                          Maybe fail);
     void optimizeModule();
     void printAfterPass(const std::string&, const std::string&, Closure*,
                         size_t);
 
+    StreamLogger& getLog() { return log; }
```
mhyee: I would return a const reference here, or just get rid of the getter entirely and make `log` public.

rir/src/compiler/translations/pir_2_rir.h
```diff
@@ -1,24 +1,29 @@
...
+    Pir2RirCompiler(const DebugOptions& debug, StreamLogger& log)
+        : debug(debug), log(log) {}
 
     const DebugOptions debug;
 
     void compile(Closure* cls, SEXP origin);
+    StreamLogger& getLog() { return log; }
 
   private:
+    StreamLogger& log;
```
mhyee, line 26: This looks weird to me, but I see that the `StreamLogger` is shared between `Pir2RirCompiler` and `Rir2PirCompiler`. Or more specifically, `Rir2PirCompiler` owns the `StreamLogger` and `Pir2RirCompiler` has a reference to it. This seems messy to me: it's not symmetric (easy to mix up which one gets the reference and which one gets the value) and intuitively, it's not obvious which one should "own" the logger.

Would it make more sense for something else to own the logger, and then both `Rir2PirCompiler` and `Pir2RirCompiler` have their own reference to the logger?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
+#ifndef PIR_LOGGER_H
+#define PIR_LOGGER_H
+
+#include "../../runtime/Function.h"
+#include "../pir/pir_impl.h"
+#include "debugging.h"
+//#include "../translations/pir_2_rir.h"
```
mhyee: Delete this?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <stack>
+//#include <fstream>
```
mhyee: Delete this?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
+#ifndef PIR_LOGGER_H
+#define PIR_LOGGER_H
+
+#include "../../runtime/Function.h"
+#include "../pir/pir_impl.h"
+#include "debugging.h"
+//#include "../translations/pir_2_rir.h"
+
+#include <iostream>
+#include <map>
```
mhyee: Do we require the map to be ordered? If not, use `unordered_map`.

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+//#include <fstream>
+
+namespace rir {
+namespace pir {
+
+class StreamLogger {
+  public:
+    StreamLogger(DebugOptions options) : options(options) {}
+    ~StreamLogger() {
+        for (std::map<rir::Function*, std::stringstream*>::iterator it =
```
mhyee: I think the new commits hid @JanJecmen's comments, but this can be rewritten as `for (auto p : streams)`, and a few lines below, `this` is not needed.

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+                 streams.begin();
+             it != streams.end(); ++it) {
+            this->finish(it->first, *streams.at(it->first));
+            /*if (options.includes(DebugFlag::PrintIntoFiles)) {
```
mhyee: Uncomment or delete.

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+
+    void startLogging(rir::Function* function);
+    void endLogging();
+    void compilationInit();
+    void compilationEarlyPir(Closure&);
+    void pirOptimizations(Closure&, const std::string&, const std::string&,
+                          size_t);
+    void phiInsertion(Closure& closure);
+    void afterCSSA(Code*);
+    // void afterLiveness(SSAAllocator&);
```
mhyee: Uncomment or delete.

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+            std::cout << it->second->str();
+            //}
+            delete it->second;
+        }
+    }
+
+    void startLogging(rir::Function* function);
+    void endLogging();
+    void compilationInit();
+    void compilationEarlyPir(Closure&);
```
mhyee: A lot of these logging helpers can be marked as `const`, right?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+#include <stack>
+//#include <fstream>
+
+namespace rir {
+namespace pir {
+
+class StreamLogger {
+  public:
+    StreamLogger(DebugOptions options) : options(options) {}
+    ~StreamLogger() {
```
mhyee, line 24: [Rule of Three](https://en.cppreference.com/w/cpp/language/rule_of_three). I don't think we want to be able to copy a `StreamLogger`, so make sure the copy constructor and copy assignment operator are deleted.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "../../utils/Pool.h"
+#include <iomanip>
+
+namespace rir {
+namespace pir {
+
+void StreamLogger::startLogging(rir::Function* function) {
+    if (options.intersects(PrintDebugPasses)) {
+        if (!streams.count(function)) {
+            streams.insert(std::pair<rir::Function*, std::stringstream*>(
```
mhyee: `streams.emplace(function, new std::stringstream())` is the idiomatic way of inserting into a map. It's shorter, clearer, and constructs the pair directly in the map.

A slightly more verbose way is to use initializer lists: `streams.insert({function, new std::stringstream()})`, but I think it's messier. (Apparently, there used to an issue with move-only semantics.)

Calling the `std::pair` constructor or `std::make_pair` is more verbose, and can actually lead to unnecessary copies, because the map actually requires the type `std::pair<const rir::Function*, std::stringstream*>`.

See:
https://stackoverflow.com/a/19942856
https://stackoverflow.com/a/17174245

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+    if (options.intersects(PrintDebugPasses)) {
+        getLog() << "\n***********************************************"
+                 << "***************\n";
+        getLog() << "*********** Start compiling: " << std::setw(20)
+                 << std::left << currentFunction.top() << " ************\n\n";
+        if (options.includes(DebugFlag::PrintEarlyRir)) {
+            getLog() << "=============== Original version:" << std::setw(16)
+                     << std::left << " ==========\n\n";
+            auto it = currentFunction.top()->begin();
+            while (it != currentFunction.top()->end()) {
```
mhyee: I'd probably rewrite this as a for loop.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+                (*it)->print(getLog());
+                ++it;
+            }
+            getLog() << "\n";
+        }
+    }
+}
+
+void StreamLogger::compilationEarlyPir(Closure& closure) {
```
mhyee: Can we make `closure` const?

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+             << std::left << " ==========\n";
+    closure.print(getLog());
+}
+
+void StreamLogger::rirFromPir(rir::Function* function) {
+    if (options.includes(DebugFlag::PrintFinalRir)) {
+        getLog() << "\n============= Final RIR Version" << std::setw(16)
+                 << std::left << " ==========\n";
+        auto it = function->begin();
+        while (it != function->end()) {
```
mhyee: For loop?

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void StreamLogger::afterCSSA(Code* code) {
+    if (options.includes(DebugFlag::PrintCSSA)) {
+        getLog() << " ========== PIR After Converting to CSSA\n";
+        code->print(getLog());
+    }
+}
+
+/*void StreamLogger::afterLiveness(SSAAllocator& allocator) {
```
mhyee, line 91: Delete?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
charig: Making it a constant is a mess because then the compiler requires to label as constant all the public functions.
I do not get why it hurts to access the member through a getter. Is it a c++ convention not to do that?

rir/src/compiler/translations/pir_2_rir.h
charig: Fully agree. There may be one, but unfortunately I do not see a solution right now. In the past, I tried to push the idea of having a compiler and then compilation phases.  Currently RIR2PIR and PIR2RIR are like isolated phases connected through some lamdas that trigger when the first phase succeed. If you find some common owner great!

rir/src/compiler/debugging/stream_logger.h
charig: I think we may want in the (near) future to print the (inner) functions in some specified order...

rir/src/compiler/debugging/stream_logger.h
charig: Oh, I would like to keep this option but there is a very strange compilation error when including `fstream: /Library/Developer/CommandLineTools/usr/include/c++/v1/fstream:872:34: error: no member named 'Rf_error' in 'std::__1::codecvt_base'
        if (__r == codecvt_base::error)
                   ~~~~~~~~~~~~~~^`.

Can anybody check it?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
o-: I agree that the consts can sometimes be a bit like cancer, you need to put them everywhere. So I think it is fine to have the StreamLogger be mutuable by any class that can touch it, that seems fine. But I think the issue here is, why is `getLog()` public? it seems to me that the logger should only be used internally and there is no need for a *client* of the rir_2_pir_compiler to access the logger.

rir/src/compiler/translations/pir_2_rir.h
```diff
@@ -1,24 +1,29 @@
...
 class Pir2RirCompiler {
   public:
-    Pir2RirCompiler(const DebugOptions& debug) : debug(debug) {}
+    Pir2RirCompiler(const DebugOptions& debug, StreamLogger& log)
+        : debug(debug), log(log) {}
 
     const DebugOptions debug;
 
     void compile(Closure* cls, SEXP origin);
+    StreamLogger& getLog() { return log; }
```
o-, line 23: should be private (or protected)

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
o-: as to the question, is a getter not idiomatic: At least for my personal taste a getter, that has no logic and is only used internally is very strange.

Usually my preference is:

1. private member, no getter
2. external client needs to access: private member, public getter

Const or non-const is an orthogonal issue. It's kind of hard to grasp at the beginning.

`const StreamLogger& getLog()` means that it will return you a const reference. That means only methods marked const can be executed on the reference that you get from `getLog`. Usually if I have a member that i can initialize in the constructor and never need to modify during its lifetime, I make it const.

If I have to modify it internally, but no external client should ever modify it, then I try to return a const reference/ptr. But as you say, in particular if you have const and non-const refs it might be annoying.

The advantage of a const member is, that the list above changes to:

1. private member, no getter
2. only read access to member by client : public const member, no getter
3. write access to member by client : private member, public non-const getter

If the member is const, then it is fine to make it public, since the client should not break your invariants, since the internal object you give it const. Thus you can avoid having a getter.

rir/src/compiler/translations/pir_2_rir.h
o-: I think you could just have it such that both do not own the StreamLog. So we have to create a log from the outside and pass it to both of them. That is simpler imho

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+        for (std::map<rir::Function*, std::stringstream*>::iterator it =
+                 streams.begin();
+             it != streams.end(); ++it) {
+            this->finish(it->first, *streams.at(it->first));
```
o-: can you manage to do this, when the function is done compiling and not only at the end? Because imagine you are in the middle of the compiler, and you want to see the inner function results as they are being compiled...

rir/src/compiler/debugging/stream_logger.h
charig: So the problem here is that it is kind of a mess to pass the SSAAllocator to the logger because that type is defined inside pir_2_rir but without a namespace. One alternative is to change the definition. But I do not think we will need to use this debug information yet. 
I left the comments so that it is explicit that it got commented now and we can uncommented it when someone finds the necessity to debug liveness.

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -0,0 +1,63 @@
...
+    void finalPIR(Code*);
+    void rirFromPir(rir::Function* function);
+    void warningBC(std::string, rir::BC bc);
+    void failCompilingPir();
+    void finish(rir::Function*, std::stringstream&);
+
+    std::stringstream& getLog() { return *streams.at(currentFunction.top()); };
+
+  private:
+    std::map<rir::Function*, std::stringstream*> streams;
```
o-: Hey, please use [std::istream](http://www.cplusplus.com/reference/istream/istream/) here and not std::stringstream. istream is the superclass of ifstream (which directly streams to a file), stringstream (which streams to a string) and iostream (for example std::cout).

If we want to log to stdout, then we put std::cout there, if we want to log to stdout, but without interleaving, then you do what you do now (use stringstream and dump at the end). if you want to log to a file, then you put a ifstream there.

JanJecmen: `s/istream/ostream`

rir/src/runtime/Code.cpp
charig: I rollback and put the streaming before. Then in case when debugging we can analyze what is the best alternative I guess.

rir/src/compiler/debugging/stream_logger.h
charig: Yes, maybe. The problem is where? Because if for instance inner functions do not run pir. So we would need a special flag for inner functions. Do you think is worth adding?

rir/src/compiler/debugging/stream_logger.h
o-: oh, that is a R macro! There is some shitty R macro that does
`#define error Rf_error`
you have to be careful not to include r.h in any header files. So transitively (probably through Function.h) you include it in stream_logger.h. only include that in the cpp file and move everything that needs it there.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
mhyee: > At least for my personal taste a getter, that has no logic and is only used internally is very strange.

I basically agree, that kind of getter seems a bit pointless. If it's used internally, then there's no need for a public interface. If there's no logic and it just returns a member, then why not just access the member directly?

If the getter is used externally, it makes sense to hide the implementation details. Even if there's no logic, maybe one day we want to change the implementation and add some logic without breaking clients. But in this case, for something like the logger, changes to the implementation will probably break the client anyway, so a getter wouldn't really hide anything.

rir/src/compiler/debugging/stream_logger.h
mhyee: If we use `map`, then it will be ordered by `Function*`. Consistent, but probably not a meaningful order.

rir/src/compiler/debugging/stream_logger.h
mhyee: In general, I don't like commented out code, because it's not clear what it's for. And it often diverges from the rest of the code, since it's not executed or tested.

As a temporary measure, could you leave a comment?

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -8,9 +8,18 @@ namespace pir {
 void StreamLogger::startLogging(rir::Function* function) {
     if (options.intersects(PrintDebugPasses)) {
         if (!streams.count(function)) {
-            streams.emplace(function, new std::stringstream());
+            std::ostream* o;
+            if (options.includes(DebugFlag::PrintIntoFiles)) {
+                // o = new std::ofstream();
```
mhyee: WIP?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -20,14 +21,15 @@ class StreamLogger {
...
-                // outFile << stream.str();
-                // outFile.close();
-            } else
-                std::cout << fun2stream.second->str();
-            delete fun2stream.second;
+            if (options.includes(DebugFlag::PrintIntoStdout))
+                continue;
+            else {
+                if (options.includes(DebugFlag::PrintIntoFiles)) {
+                    //((std::ofstream*)fun2stream.second)->close();
```
mhyee: WIP?

rir/src/interpreter/deoptimizer.cpp
```diff
@@ -61,10 +61,10 @@ static DeoptInfo* DeoptInfo_get(uint32_t idx) {
...
     assert(idx != NO_DEOPT_INFO);
     DeoptInfo* i = DeoptInfo_get(idx);
-    Rprintf("[d#%d: %p]", idx, i->oldPc);
+    out << "d#" << idx << ": " << static_cast<void*>(i->oldPc);
```
mhyee: Output string lost the `[` and `]` brackets.

rir/src/runtime/Code.cpp
```diff
@@ -11,54 +13,58 @@ Code::Code(SEXP ast, unsigned cs, unsigned sourceLength, unsigned offset,
...
-void Code::print() {
-    Rprintf("Code object (%p offset %x (hex))\n", this, header);
-    Rprintf("  Source:      %u (index to src pool)\n", src);
-    Rprintf("  Magic:       %x (hex)\n", magic);
-    Rprintf("  Stack (o):   %u\n", stackLength);
-    Rprintf("  Code size:   %u [B]\n", codeSize);
-    Rprintf("  Default arg? %s\n", isDefaultArgument ? "yes" : "no");
-    if (magic != CODE_MAGIC)
+void Code::print(std::ostream& out) {
+    out << "Code object (" << this << "offset" << std::hex << header << "(hex))"
```
mhyee: Still missing spaces:

`out << "Code object (" << this << " offset " << std::hex << header << " (hex))"`

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,148 @@
...
+            innerHeader(" Original version ");
+            for (auto code : *(currentFunction.top())) {
+                code->print(getLog());
+            }
+            getLog() << "\n";
+        }
+    }
+}
+
+void StreamLogger::compilationEarlyPir(Closure& closure) const {
```
mhyee: `const Closure& closure`

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,148 @@
...
+                                    size_t passnr) const {
+    if (options.includes(DebugFlag::PrintOptimizationPasses)) {
+        std::stringstream ss;
+        ss << category << ": " << pass << " == " << passnr;
+        innerHeader(ss.str());
+        closure.print(getLog());
+    }
+}
+
+void StreamLogger::rirFromPir(rir::Function* function) const {
```
mhyee: Can we do `const rir::Function* function`? We shouldn't need to modify what `function` points to.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (options.includes(DebugFlag::PrintFinalRir)) {
+        innerHeader(" Final RIR Version ");
+        for (auto code : *function) {
+            code->print(getLog());
+        }
+        getLog() << "\n";
+    }
+}
+
+void StreamLogger::phiInsertion(Code* code) const {
```
mhyee: `const Code* code`? And other places too.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+void StreamLogger::failCompilingPir() const {
+    if (options.includes(DebugFlag::ShowWarnings))
+        getLog() << " Failed verification after p2r compile "
+                 << &currentFunction.top() << "\n";
+}
+
+void StreamLogger::header(std::string header, rir::Function* function,
```
mhyee: 👍

mhyee (commented):
Thanks for the fixes! Still a few more comments.

rir/src/compiler/debugging/stream_logger.cpp
charig: I can't do the loop if it is a const:
```
for (auto code : *function) {
      code->print(getLog());
}
```

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
charig: So, in this case the member is used by clients. The `StreamLogger` is owned by `Rir2PirCompiler`and used elsewhere by the compilation phases. To avoid passing it as parameter I added the getter so that any client already having a reference to the compiler can access the log. The situation is quite similar in the `Pir2RirCompiler`. The class actually doing stuff and needing the logger is `Pir2Rir`which has a reference to the compiler. So, at least under this model I do not see an alternative. 

Then, the getter can not be *const* because the clients use startLogging which uses emplace that need a *non-const* member. And finally, I could remove the reference and make the member just public ok. But I still do not see how the getter hurts. And from why I understand from Oli's quote, he is not againts it: 
> 3. write access to member by client : private member, public non-const getter

mhyee: I don't think it hurts, but I wanted to clarify why I didn't really like it in the first place. Maybe the problem is that `Rir2PirCompiler` owns the `StreamLogger` (and `Pir2RirCompiler` has a reference)?

Anyway, I won't push any more on this, it's already been enough of a detour!

PR #189, opened 2018-08-14
Bugssss

o-:
:bug: :skull: :hear_no_evil:


mhyee (approved):
Thanks! I also confirmed that it fixes the snippet I found:

```R
    aaa <- function(x) bbb(x)
    bbb <- function (x) {
        ccc <- function(x, ind) {
            if (ind) {
                for (i in 1:5) ccc(x, FALSE)
            }
            x
        }
        ccc(x, TRUE)
    }
    aaa(42)
```

PR #190, opened 2018-08-15
enable pir by default

o-:
* by default we optimize functions with pir
* on travis we also test compiling every function through pir upfront


o-:
:tada:

PR #192, opened 2018-08-16
Fix inliner

o-:


rir/src/compiler/pir/module.h
```diff
@@ -62,7 +64,7 @@ class Module {
 
   private:
     Closure* declare(rir::Function*, const std::vector<SEXP>& a, Env* env);
-    std::unordered_map<rir::Function*, VersionedClosure> functions;
+    std::map<FunctionAndEnv, VersionedClosure> functions;
```
mhyee, line 67: Oh... I forgot that you can't (easily) use `std::pair` as a key for `unordered_map`, because it doesn't satisfy `std::hash`...

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -663,10 +665,19 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
-                        cs << BC::push(Env::Cast(what)->rho);
-                    } else if (what == Env::notClosed()) {
+                    if (what == Env::notClosed()) {
                         cs << BC::parentEnv();
+                    } else if (what == Env::nil()) {
+                        cs << BC::push(R_NilValue);
+                    } else if (Env::isStaticEnv(what)) {
+                        auto env = Env::Cast(what);
+                        // Here we could also load env->rho, but if the user
+                        // where to change the environment on the closure our
```
mhyee: where -> were

rir/src/compiler/pir/module.h
o-: yeah, I usually make a mistake when i implement hash<>, so first I wanna see it running with map :) if we care enough to change it, we still can..

mhyee: Oh look, there's a C++ library proposal for `hash_combine`: http://wg21.link/p0814r2

Because the proposals in 2012 and 2014 died...

I'm fine with leaving it like this for now. Do we have any guess as to how big the map would be?

rir/src/compiler/opt/inline.cpp
```diff
@@ -33,16 +35,24 @@ class TheInliner {
...
                 } else if (auto call = StaticCall::Cast(*it)) {
                     inlinee = call->cls();
+                    // if we don't know the closure of the inlinee, we can't
+                    // inline.
+                    if (inlinee->closureEnv() == Env::notClosed() &&
+                        inlinee != function)
+                        continue;
+                    // TODO: honestly I have no clue how namespaces work. For
+                    // now if we hit any env with attribs, we do not inline.
+                    if (ATTRIB(inlinee->closureEnv()->rho))
```
JanJecmen: Don't they put `R_NilValue` to the slot if there are no attributes? This tests for nullptr

PR #198, opened 2018-08-21
Code objects are now embedded in SEXPs

### LINKED ISSUE #197: Make Code objects actual objects
**@o- wrote**:
> I have also found the root cause of the inliner bug [#192]
>
> it's really just that our code objects can outlive the closure objects
> so, I think one refactoring, that would really help us, is to make code objects actual objects...
> when we have that, then we can have: code objects that point back to the closure or promise and we finally do not have to keep all the code we ever generate alive
>
> Basically, currently our rir Functions (in runtime/function.h), we just put them into a raw byte array
> at the end of the Function struct there are all the Code structs (from runtime/code.h) inline
> so we get one massive blob per function, with all the promises
> and we have in-object pointers to the code objects (pointing inside the Function struct)
> and we prevent R from sev, by just faking a SEXP header at the beginning of the Code objects
> the problem with that is, (1) the R GC always just ignores the Code objects (thus we cannot have the code objects keep closure/promise alive)
> and (2) if a promise outlives a closure, then the whole function with all code objects needs to be kept alive
> so what I would like to move to is one real SEXP per code object
> Then the function just pointing to all the code objects (like the dispatch table points to all functions)
> there will be some issues to resolve. for example in several places we iterate over the code object by just advancing a pointer by the size of the code object...
> and the function writer (in utils/functionwirter) has to be rewritten
> though in general the one EXTERNALSXP per code object approach should be much easier

mhyee:
Before, Code objects were serialized into a binary array at the end of
the Function struct. This meant they were opaque to the garbage
collector and couldn't be used to keep closures alive. There was also
some tricky business with doing raw pointer arithmetic and having the
Code object header pretend to be a SEXP.

Now, Code objects are actual SEXPs -- or more specifically, they are
embedded inside EXTERNALSXPs, similar to Function objects and
DispatchTable objects.

This required some other changes:

  - The FunctionWriter no longer has a create() method; you should use
    the constructor directly. You will also need to explicitly call
    `finalize()` when you are done adding code. (Internally, code
    objects are added to a vector and then a Function is created at the
    end, avoiding the need to constantly resize/reallocate the
    Function.)

  - All the Code iterators had to be rewritten, but are now simpler
    because they use an index instead of constantly computing pointer
    offsets.

  - findDefaultArgument was completely refactored. Now it a method of
    Function.

This commit also includes some other refactors:

  - Function constructor uses member initializer list.
  - Consolidated isValid{Code,Function,DispatchTable}Object() into
    {Code,Function,DispatchTable}::check().
  - Removed isValidFunctionSEXP and other redundant or unused code.
  - Moved some code from Code.h to Code.cpp.

Fixes #197.


rir/src/interpreter/interp.cpp
```diff
@@ -2656,16 +2654,13 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
 }
 
 SEXP rirExpr(SEXP f) {
-    if (isValidCodeObject(f)) {
-        Code* c = (Code*)f;
+    if (auto c = isValidCodeObject(f)) {
```
o-: `Code::check`

rir/src/runtime/Function.h
```diff
@@ -47,20 +47,25 @@ typedef SEXP FunctionSEXP;
...
-        invocationCount = 0;
-        deopt = false;
-        markOpt = false;
-        codeLength = 0;
-        foffset = 0;
+        for (size_t i = 0; i < codeLength; ++i) {
+            Code* c = Code::unpack(codeVec[i]);
+            // TODO(mhyee): write barrier?
+            c->function_ = this->container();
+            data[i] = c->container();
```
o-: yes, this and the previous line needs a write barrier. `EXTERNALSXP_SET_ENTRY`

rir/src/runtime/Function.h
```diff
@@ -153,11 +158,12 @@ struct Function {
...
-    // We can get to this by searching, but this is faster and so worth the
-    // extra four bytes
-    unsigned foffset; ///< Offset to the code of the function (last code)
+  private:
+    FunctionSEXP origin_; /// Same Function with fewer optimizations,
+                          //   NULL if original
+    FunctionSEXP next_;
 
-    uint8_t data[]; // Code objects stored inline
+    CodeSEXP data[]; // Code objects stored inline
```
o-: give it a more telling name?

rir/src/utils/FunctionWriter.h
```diff
@@ -7,78 +7,66 @@
...
-        assert(function->size <= capacity);
-
-        Code* code = new (insert) Code(ast, codeSize, sources.size(), offset,
-                                       markDefaultArg, localsCnt);
-
-        assert(code->function() == function);
+        SEXP store = Rf_allocVector(EXTERNALSXP, totalSize);
+        void* payload = INTEGER(store);
+        // TODO(mhyee): nullptr might be problematic if GC tries to trace it too
+        // soon
```
o-: ok, gc skipps nulls

rir/src/interpreter/interp.cpp
o-: (more of those below, you can remove isValidCodeObj)

o- (commented):
can't wait :)

rir/src/interpreter/interp.cpp
mhyee: Looks like `isValid{Code,Function,DispatchTable}Object` checks that the object is an `EXTERNALSXP` first, and then calls `{Code,Function,DispatchTable}::check`.  I could move the `EXTERNALSXP` check into `check`.

o-: sgtm

mhyee:
Net change: 12 lines deleted. I was hoping for more. :)

I tried to split out the refactoring, but it was too integrated. So now it's one giant commit. Probably the main files to look at are `Code.{h,cpp}`, `Function.h`, and `FunctionWriter.h`.

Everything should pass the Travis tests, but when I ran a gctorture test, it failed. The same gctorture fails on master, but maybe with a different error.

Right now, I store all the Code objects in the constant pool. In a later PR (or maybe a later commit in this PR), I can delete some of the calls to `Pool::insert()` so we actually take advantage of Code objects being SEXPs.

mhyee:
Well, you can review this if you want, but it looks like I still need to debug. One of the tests failed, when I restarted in debug mode it passed, but when I re-ran it with environment variable `R_GCTORTURE=100` it failed. But then it worked fine locally...

mhyee:
I fixed the aforementioned bug (thanks to a pointer from @o-), but it looks like there's more. Seeing as #208 passed, the bug is in this PR. I'll try to keep working on it, but it's starting to get more difficult and annoying to reproduce.

mhyee:
This is now ready for review. I abandoned tracking down that bug I mentioned; it's extremely transient and may be a bug in R.

The last two commits ensure we don't keep Code objects and closures alive forever --- they can be collected. However, if a promise escapes, then it ends up keeping everything alive. Changing this will be tricky, so it's out of scope. See issue #215.

rir/src/runtime/Code.h
```diff
@@ -9,25 +10,22 @@
...
 struct FunctionSignature;
 
-// Code magic constant is intended to trick the GC into believing that it is
-// dealing with already marked SEXP.
-// Note: gcgen needs to be 1, otherwise the write barrier will trigger and
-//       named count is 3 (ie. NAMEDMAX) to make it stable
-//  It also has an unique bitpattern for gp (0xeeee) so that we can keep it
-//  apart from functions
-#define CODE_MAGIC (uint64_t)0x311eeee1a
+#define CODE_MAGIC 0x3111eeee
```
o-: can we change that, so it does *not* look like a sexp anymore?

rir/src/runtime/Code.h
```diff
@@ -98,11 +101,33 @@ struct Code {
         unsigned srcIdx;
     };
 
+    SEXP container() {
```
o-: would it make sense to have a super-class for Runtime objects (ie. for dispatchtable, code, function). Something like:

```
tamplate<typename BASE>
struct RirRuntimeObject<BASE> {
  rir::rir_header info;
  RirRuntimeObject(SEXP* gc_area_start, size_t gc_area_length) {
    memset(gc_area_start, ....)_
    initialize_rir_header ....
  }
  setElement(size_t pos, SEXP v) {
    EXTERNALSXP_SET_ENTRY(...)
  }
  SEXP container();
  BASE* unpack(SEXP);
  BASE* check(SEXP);
};
```

o- (approved):
lgtm, except for my two nitpicks

rir/src/runtime/Code.h
JanJecmen: Could then also get rid of the header and put the stuff here. Also, the last two functions are static

rir/src/interpreter/interp.cpp
```diff
@@ -479,21 +479,19 @@ SEXP closureArgumentAdaptor(const CallContext& call, SEXP arglist,
     f = FORMALS(op);
     a = actuals;
     // get the first Code that is a compiled default value of a formal arg
-    // (or end() if no such exist)
+    // (or nullptr if no such exist)
     Function* fun = DispatchTable::unpack(BODY(op))->first();
-    Code* c = findDefaultArgument(fun->first());
-    Code* e = fun->codeEnd();
+    Code* c = fun->findDefaultArg();
```
JanJecmen: be explicit and put the zero here at the callsite?

rir/src/interpreter/interp.cpp
```diff
@@ -2671,16 +2669,13 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
...
 
 SEXP rirExpr(SEXP f) {
-    if (isValidCodeObject(f)) {
-        Code* c = (Code*)f;
+    if (auto c = Code::check(f)) {
         return src_pool_at(globalContext(), c->src);
     }
-    Function* ff;
-    if ((ff = Function::check(f))) {
+    if (auto ff = Function::check(f)) {
```
JanJecmen: Why is the parameter even called `f`? Could be `SEXP s` and then `auto f = Function::check(s)` (feel free to ignore)

rir/src/interpreter/interp_data.h
```diff
@@ -21,10 +21,6 @@ typedef rir::Code Code;
 typedef rir::Function Function;
 typedef rir::DispatchTable DispatchTable;
 
-// all sizes in bytes,
-// length in element sizes
-
-
 // ============
 // Please do not change those without also changing how they are handled in the
 // CodeEditor
```
JanJecmen: Unrelated, but `CodeEditor` no longer exists

rir/src/interpreter/runtime.cpp
```diff
@@ -10,33 +10,20 @@ SEXP promExecName;
...
         return nullptr;
-    return f;
-}
-
-Code* isValidPromiseSEXP(SEXP promise) {
-    return isValidCodeObject(PRCODE(promise));
+    }
+    if (auto t = DispatchTable::check(BODY(closure))) {
+        Function* f = t->first();
+        return f->info.magic == FUNCTION_MAGIC ? f : nullptr;
```
JanJecmen: `return Function::check(t->first()) ? f : nullptr`
(and shouldn't it check all functions in the table?)

rir/src/interpreter/runtime.h
```diff
@@ -8,15 +8,11 @@
...
-C_OR_CPP Code* isValidPromiseSEXP(SEXP promise);
-C_OR_CPP Function* isValidClosureSEXP(SEXP closure);
-C_OR_CPP Function* isValidFunctionSEXP(SEXP wrapper);
-C_OR_CPP DispatchTable* isValidDispatchTableSEXP(SEXP wrapper);
-
 /** Checks if given closure should be executed using RIR.
 
   If the given closure is RIR function, returns its Function object, otherwise returns nullptr.
  */
+C_OR_CPP Function* isValidClosureSEXP(SEXP closure);
```
JanJecmen, line 15: oh the comment was misplaced :)

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -1083,6 +1083,7 @@ rir::Function* Pir2Rir::finalize() {
     ctx.pushBody(R_NilValue);
     size_t localsCnt = compileCode(ctx, cls);
     ctx.finalizeCode(localsCnt);
+    function.finalize();
```
JanJecmen, line 1143: Hide the `function` behind a getter that would check that the writer is finalized? Didn't come up with a better solution to forgetting to call finalize

rir/src/runtime/Code.h
JanJecmen: Are there any good hexspeak words left? ;]

rir/src/runtime/Function.h
```diff
@@ -47,20 +47,33 @@ typedef SEXP FunctionSEXP;
...
   public:
-    Function() {
-        info.gc_area_start = sizeof(rir_header); // just after the header
-        info.gc_area_length = 2;                 // origin, next
+    Function(size_t functionSize, const std::vector<SEXP>& codeVec)
+        : size(functionSize), signature(nullptr), invocationCount(0),
+          deopt(false), markOpt(false), codeLength(codeVec.size()),
+          origin_(nullptr), next_(nullptr) {
+        // GC area starts just before the end of Function
+        info.gc_area_start =
```
JanJecmen: as long as we have the rir_header (see Oli's comment), this can be also in the member initializer
`info { sizeof(Function) - CODEOBJ_OFFSET * sizeof(FunctionSEXP), CODEOBJ_OFFSET + codeLength, FUNCTION_MAGIC }`

rir/src/runtime/Function.h
```diff
@@ -47,20 +47,33 @@ typedef SEXP FunctionSEXP;
...
-        next_ = nullptr;
-        size = sizeof(Function);
-        signature = nullptr;
-        invocationCount = 0;
-        deopt = false;
-        markOpt = false;
-        codeLength = 0;
-        foffset = 0;
+        // initialize codeObjects array to nullptr first!
+        for (size_t i = 0; i < codeLength; ++i) {
```
JanJecmen: Aren't all these nulls always overwritten in the following for loop? And the gc shouldn't happen before we set them?

rir/src/runtime/Code.h
mhyee: Sure, I just punted on choosing the constant. But then I found [this](https://nedbatchelder.com/text/hexwords.html), so I think I'll make it `c0de0000`. And while I'm being ~clever~ obnoxious, maybe I'll change FUNCTION_MAGIC to `ca11ab1e` and DISPATCH_MAGIC to `d7ab1e00`.

rir/src/runtime/Code.h
mhyee: That's a good idea. The repetition was starting to annoy me.

I don't know how likely it is that we'll add another runtime object, but the (non-C++) [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) would apply.

rir/src/runtime/Function.h
```diff
@@ -47,20 +47,33 @@ typedef SEXP FunctionSEXP;
 #pragma pack(push)
 #pragma pack(1)
 struct Function {
+    friend class FunctionCodeIterator;
+    friend class ConstFunctionCodeIterator;
+
+    static constexpr size_t CODEOBJ_OFFSET = 2;
+
   public:
```
JanJecmen: this `public` is redundant

rir/src/runtime/Function.h
```diff
@@ -133,12 +143,20 @@ struct Function {
...
+        return nullptr;
+    }
 
-  private:
-    FunctionSEXP origin_; /// Same Function with fewer optimizations,
-                          //   NULL if original
-    FunctionSEXP next_;
+    rir::rir_header info; /// for exposing SEXPs to GC
 
   public:
```
JanJecmen: this `public` as well

rir/src/runtime/Function.h
```diff
@@ -153,11 +171,18 @@ struct Function {
...
+    // !!! Furthermore, you need to update                 !!!
+    // !!!     the CODEOBJ_OFFSET constant.                !!!
+
+    FunctionSEXP origin_; /// Same Function with fewer optimizations,
+                          //   NULL if original
+    FunctionSEXP next_;
 
-    uint8_t data[]; // Code objects stored inline
+    CodeSEXP codeObjects[]; /// Pointers to CodeSEXPs (Code objects embedded
+                            //   inside EXTERNALSXP)
```
JanJecmen, line 150: And there is no way in the api to update `codObjects` after the constructor, that is intentional? (I guess it makes sense)

rir/src/utils/FunctionWriter.h
```diff
@@ -5,80 +5,69 @@
...
+        assert(function == nullptr && "Trying to finalize a second time");
+
+        size_t dataSize = codeVec.size() * sizeof(SEXP);
+        assert(functionSize == sizeof(Function) + dataSize);
+        assert(functionSize % sizeof(int) == 0);
+
+        SEXP store = Rf_allocVector(EXTERNALSXP, functionSize);
         void* payload = INTEGER(store);
+        Function* fun = new (payload) Function(functionSize, codeVec);
+        preserve(fun->container());
```
JanJecmen: The `Function` constructor doesn't trigger gc, but wouldn't it be safer to `preserve(store)` immediately after allocating it?

rir/src/utils/FunctionWriter.h
```diff
@@ -5,80 +5,69 @@
...
-        Code* code = new (insert) Code(ast, codeSize, sources.size(), offset,
-                                       markDefaultArg, localsCnt);
-
-        assert(code->function() == function);
+        SEXP store = Rf_allocVector(EXTERNALSXP, totalSize);
+        void* payload = INTEGER(store);
+        Code* code =
+            new (payload) Code(nullptr, index, ast, codeSize, sources.size(),
+                               markDefaultArg, localsCnt);
+        preserve(store);
```
JanJecmen, line 74: The same, preserve right after alloc?

rir/src/runtime/Code.h
```diff
@@ -164,25 +150,26 @@ struct Code {
 
 #pragma pack(pop)
 
-class CodeHandleIterator {
-    Code* code;
+class FunctionCodeIterator {
+    Function const* const function;
```
JanJecmen, line 129: Nice! `const` everything :D

JanJecmen (approved):
:+1:

rir/src/interpreter/interp.cpp
mhyee: Probably an artifact of old code, but since I'm refactoring, I might as well change it.

rir/src/compiler/translations/pir_2_rir.cpp
mhyee: Excellent idea. This bit me a few times while working on this...

rir/src/runtime/Function.h
mhyee: Correct. (There wasn't a way before to update the Code objects.)

rir/src/runtime/Function.h
mhyee: Why can't GC happen before we set the array?

In any case, when I refactor the RIR objects to use a superclass (per Oli's comment), I'll probably be conservative and have a `memset` anyway.

rir/src/interpreter/runtime.cpp
mhyee: Looks like this is actually redundant. `t->first()` calls `Function::unpack` which checks the magic number in an assert.

So I'll just change this to `return t->first()`.

rir/src/utils/FunctionWriter.h
mhyee: Turns out that's wrong. We can only preserve after initializing the object.

mhyee:
I rebased with the changes responding to feedback. For your convenience, the diff is included below.

I also pushed a new commit with the refactor that adds a `RirRuntimeObject` type. It saves 36 lines of code, eliminates some duplication, but adds a bit more complexity.

<details>
<summary>Diff of the rebased changes</summary>

```diff
diff --git c/rir/src/compiler/translations/pir_2_rir.cpp w/rir/src/compiler/translations/pir_2_rir.cpp
index a879971..d199cd2 100644
--- c/rir/src/compiler/translations/pir_2_rir.cpp
+++ w/rir/src/compiler/translations/pir_2_rir.cpp
@@ -1086,11 +1086,11 @@ rir::Function* Pir2Rir::finalize() {
     function.finalize();
     LOGGING(compiler.getLogger().finalPIR(*cls));
 #ifdef ENABLE_SLOWASSERT
-    CodeVerifier::verifyFunctionLayout(function.function->container(),
+    CodeVerifier::verifyFunctionLayout(function.function()->container(),
                                        globalContext());
 #endif
-    LOGGING(compiler.getLogger().rirFromPir(function.function));
-    return function.function;
+    LOGGING(compiler.getLogger().rirFromPir(function.function()));
+    return function.function();
 }

 } // namespace
diff --git c/rir/src/interpreter/interp.cpp w/rir/src/interpreter/interp.cpp
index 0f4af67..3c74943 100644
--- c/rir/src/interpreter/interp.cpp
+++ w/rir/src/interpreter/interp.cpp
@@ -481,7 +481,7 @@ SEXP closureArgumentAdaptor(const CallContext& call, SEXP arglist,
     // get the first Code that is a compiled default value of a formal arg
     // (or nullptr if no such exist)
     Function* fun = DispatchTable::unpack(BODY(op))->first();
-    Code* c = fun->findDefaultArg();
+    Code* c = fun->findDefaultArg(0);
     while (f != R_NilValue) {
         if (CAR(f) != R_MissingArg) {
             if (CAR(a) == R_MissingArg) {
@@ -2668,19 +2668,19 @@ eval_done:
     return ostack_pop(ctx);
 }

-SEXP rirExpr(SEXP f) {
-    if (auto c = Code::check(f)) {
+SEXP rirExpr(SEXP s) {
+    if (auto c = Code::check(s)) {
         return src_pool_at(globalContext(), c->src);
     }
-    if (auto ff = Function::check(f)) {
-        return src_pool_at(globalContext(), ff->body()->src);
+    if (auto f = Function::check(s)) {
+        return src_pool_at(globalContext(), f->body()->src);
     }
-    if (auto t = DispatchTable::check(f)) {
+    if (auto t = DispatchTable::check(s)) {
         // Default is the source of the first function in the dispatch table
-        Function* ff = t->first();
-        return src_pool_at(globalContext(), ff->body()->src);
+        Function* f = t->first();
+        return src_pool_at(globalContext(), f->body()->src);
     }
-    return f;
+    return s;
 }

 SEXP rirEval_f(SEXP what, SEXP env) {
diff --git c/rir/src/interpreter/interp_data.h w/rir/src/interpreter/interp_data.h
index d31fcb9..26f14fd 100644
--- c/rir/src/interpreter/interp_data.h
+++ w/rir/src/interpreter/interp_data.h
@@ -21,9 +21,6 @@ typedef rir::Code Code;
 typedef rir::Function Function;
 typedef rir::DispatchTable DispatchTable;

-// ============
-// Please do not change those without also changing how they are handled in the
-// CodeEditor
 // Indicates an argument is missing
 #define MISSING_ARG_IDX ((unsigned)-1)
 // Indicates an argument does not correspond to a valid CodeObject
diff --git c/rir/src/interpreter/runtime.cpp w/rir/src/interpreter/runtime.cpp
index f49e62a..b534a0a 100644
--- c/rir/src/interpreter/runtime.cpp
+++ w/rir/src/interpreter/runtime.cpp
@@ -20,8 +20,7 @@ Function* isValidClosureSEXP(SEXP closure) {
         return nullptr;
     }
     if (auto t = DispatchTable::check(BODY(closure))) {
-        Function* f = t->first();
-        return f->info.magic == FUNCTION_MAGIC ? f : nullptr;
+        return t->first();
     }
     return nullptr;
 }
diff --git c/rir/src/ir/Compiler.cpp w/rir/src/ir/Compiler.cpp
index 693f0e7..bb6bdc4 100644
--- c/rir/src/ir/Compiler.cpp
+++ w/rir/src/ir/Compiler.cpp
@@ -924,14 +924,14 @@ SEXP Compiler::finalize() {
     ctx.pop();
     function.finalize();

-    function.function->signature = signature;
+    function.function()->signature = signature;

 #ifdef ENABLE_SLOWASSERT
-    CodeVerifier::verifyFunctionLayout(function.function->container(),
+    CodeVerifier::verifyFunctionLayout(function.function()->container(),
                                        globalContext());
 #endif

-    return function.function->container();
+    return function.function()->container();
 }

 }  // namespace rir
diff --git c/rir/src/runtime/Function.h w/rir/src/runtime/Function.h
index 304e3af..26eebf1 100644
--- c/rir/src/runtime/Function.h
+++ w/rir/src/runtime/Function.h
@@ -52,7 +52,6 @@ struct Function {

     static constexpr size_t CODEOBJ_OFFSET = 2;

-  public:
     Function(size_t functionSize, const std::vector<SEXP>& codeVec)
         : size(functionSize), signature(nullptr), invocationCount(0),
           deopt(false), markOpt(false), codeLength(codeVec.size()),
@@ -143,7 +139,7 @@ struct Function {
             invocationCount++;
     }

-    Code* findDefaultArg(size_t index = 0) const {
+    Code* findDefaultArg(size_t index) const {
         while (index < codeLength) {
             Code* c = Code::unpack(codeObjects[index++]);
             if (c->isDefaultArgument) {
@@ -158,7 +154,6 @@ struct Function {

     rir::rir_header info; /// for exposing SEXPs to GC

-  public:
     unsigned size; /// Size, in bytes, of the function and its data

     FunctionSignature* signature; /// pointer to this version's signature
diff --git c/rir/src/runtime/RirHeader.h w/rir/src/runtime/RirHeader.h
index 937be7d..7fbae9e 100644
--- c/rir/src/runtime/RirHeader.h
+++ w/rir/src/runtime/RirHeader.h
@@ -25,9 +25,6 @@ struct rir_header {
     // TODO:  Later maybe also add type of the object here and have just
     //        one EXTERNALSXP with a union of other types.
     // For now just make sure that this header is in all RIR
-
-    // The exception is Code, since it is not a valid SEXP in the
-    // first place.
 };

 }
diff --git c/rir/src/utils/FunctionWriter.h w/rir/src/utils/FunctionWriter.h
index 694ad19..6235c3f 100644
--- c/rir/src/utils/FunctionWriter.h
+++ w/rir/src/utils/FunctionWriter.h
@@ -16,24 +16,30 @@
 namespace rir {

 class FunctionWriter {
+  private:
+    Function* function_;
+
   public:
     typedef unsigned PcOffset;

     size_t functionSize; // size of the Function, including pointers to its code
                          // objects

-    Function* function;
-
     std::vector<SEXP> codeVec;

     Preserve preserve;

-    FunctionWriter() : functionSize(sizeof(Function)), function(nullptr) {}
+    FunctionWriter() : function_(nullptr), functionSize(sizeof(Function)) {}

     ~FunctionWriter() {}

+    Function* function() {
+        assert(function_ && "FunctionWriter has not been finalized");
+        return function_;
+    }
+
     void finalize() {
-        assert(function == nullptr && "Trying to finalize a second time");
+        assert(function_ == nullptr && "Trying to finalize a second time");

         size_t dataSize = codeVec.size() * sizeof(SEXP);
         assert(functionSize == sizeof(Function) + dataSize);
@@ -42,11 +48,11 @@ class FunctionWriter {
         SEXP store = Rf_allocVector(EXTERNALSXP, functionSize);
         void* payload = INTEGER(store);
         Function* fun = new (payload) Function(functionSize, codeVec);
-        preserve(fun->container());
+        preserve(store);

         assert(fun->info.magic == FUNCTION_MAGIC);

-        function = fun;
+        function_ = fun;
     }

     Code* writeCode(SEXP ast, void* bc, unsigned originalCodeSize,
@@ -54,7 +60,7 @@ class FunctionWriter {
                     const std::map<PcOffset, BC::Label>& patchpoints,
                     const std::map<PcOffset, std::vector<BC::Label>>& labels,
                     bool markDefaultArg, size_t localsCnt, size_t nops) {
-        assert(function == nullptr &&
+        assert(function_ == nullptr &&
                "Trying to add more code after finalizing");
         unsigned codeSize = originalCodeSize - nops;
         unsigned totalSize = Code::size(codeSize, sources.size());
```
</details>

rir/src/runtime/RirRuntimeObject.h
```diff
@@ -0,0 +1,78 @@
...
+        if (TYPEOF(s) != EXTERNALSXP) {
+            return nullptr;
+        }
+        BASE* b = (BASE*)INTEGER(s);
+        return b->info.magic == MAGIC ? b : nullptr;
+    }
+
+  protected:
+    RirRuntimeObject(uint32_t gc_area_start, uint32_t gc_area_length)
+        : info{gc_area_start, gc_area_length, MAGIC} {
```
o-, line 70: whoo, looks like ruby :)

mhyee:
Just a note, but I would have preferred not squashing this commit, since I took the trouble to separate out my commits...

o-:
> Just a note, but I would have preferred not squashing this commit, since I took the trouble to separate out my commits...

oh, i am sorry.

My assumption was that when the history should be preserved, the branch is rebased and force updated on review and otherwise I just squash.

will keep it in mind the next time and ask you.

o-:
ah crap, you actually did force update the branch. but I was looking at a stale cached  version of the history on github. I am sorry

mhyee:
No big deal, I think the majority of changes were in one commit anyway. Oh well.

rir/src/runtime/Function.h
```diff
@@ -46,63 +46,45 @@ typedef SEXP FunctionSEXP;
  */
 #pragma pack(push)
```
charig, line 47: Unrelated, but...are these pragmas still needed?

mhyee: Hmm, I'm not sure. I tried compiling with and without, and with my compiler, the pragma saves us 4 bytes in every `Function` object.

Here's the layout I observed:

```c++
struct Function {
  // 0 bytes from start
  // struct RirRuntimeObject {
  //   struct rir_header {
  //     uint32_t gc_area_start;
  //     uint32_t gc_area_end;
  //     uint32_t magic;
  //   };
  // }; // 12 bytes

  // 12 bytes from start
  unsigned size; // 4 bytes

  // 16 bytes from start
  FunctionSignature* signature; // 8 bytes

  // 4 bytes padding (can't remove with #pragma pack)

  // 28 bytes from start
  unsigned deopt : 1;
  unsigned markOpt : 1;
  unsigned spare : 30;  // 4 bytes

  // 32 bytes from start
  unsigned codeLength; // 4 bytes

  // 4 bytes padding (removed with #pragma pack)

  // 40 bytes from start (36 with #pragma pack)
  FunctionSEXP origin_; // 8 bytes

  // 48 bytes from start (44 with #pragma pack)
  FunctionSEXP next_; // 8 bytes

  CodeSEXP codeObjs[]; // 0 bytes (dynamically allocated)
}; // end is 56 bytes from start (52 bytes with #pragma pack)
```

PR #199, opened 2018-08-22
Add rir support for deoptimization

o-:
Add a deopt rir instruction. Currently this one does nothing more
than changing the pc and code pointer. All the smarts are in pir,
where we make sure that an env is created and all the values on
the stack are captured.

Currently we do not inline the deopt instruction. Thus we do not
need to synthesize call frames. I plan to support those features
in a subsequent patch.


JanJecmen:
Cool removing that guessing... I found myself disabling it multiple times because it introduced different call instructions than I expected when testing stuff...

You can remove the `SEXP findClosure(SEXP sym, SEXP rho)` function just above it too, it's only used for the deleted code

PR #201, opened 2018-08-23
Large Cleanup of Rir2Pir

o-:
In preparation using a dedicated Safepoint instruction and automatically constructing appropriate safepoints through the builder, I started a rather large refactoring of Rir2Pir.

I also stumbled upon the branches having meaningless names (next0 and next1) and fixed that.

There is some bycatch, this commit also splits Deopt and Safepoint into two instructions. But currently they directly follow each other, so nothig really changes here.

All in all, if you do not want to review the PR, I think it really is just a cleanup, makes the code strictly simpler and adds a lot of asserts, if you where to wrongly use the builder.

I had to get rid of the StackMachine. The Rir2Pir and the stack machine had so much duplicated state, it was always just a pain.


rir/src/compiler/pir/bb.h
```diff
@@ -82,6 +79,45 @@ class BB {
 
     void gc();
 
+    bool isExit() const { return !next0 && !next1; }
+
+    void replaceNextBranches(BB* from) {
```
charig: Override?

rir/src/compiler/pir/bb.h
```diff
@@ -82,6 +79,45 @@ class BB {
...
+        assert(!next0 && !next1);
+        this->next0 = trueBranch;
+        this->next1 = falseBranch;
+    }
+
+    BB* trueBranch() { return next0; }
+
+    BB* falseBranch() { return next1; }
+
+    void overrideNext(BB* bb) {
```
charig: Or replace here?

rir/src/compiler/pir/bb.h
```diff
@@ -82,6 +79,45 @@ class BB {
...
     void gc();
 
+    bool isExit() const { return !next0 && !next1; }
+
+    void replaceNextBranches(BB* from) {
+        next0 = from->next0;
+        next1 = from->next1;
+    }
+
+    void setNextBranches(BB* trueBranch, BB* falseBranch) {
```
charig: Naming is a little inconsistent. Why set here and not setNext()? On the other hand why nextBranches. Maybe only branches? or setBranches otherwise?

rir/src/compiler/pir/bb.h
```diff
@@ -82,6 +79,45 @@ class BB {
...
+    BB* trueBranch() { return next0; }
+
+    BB* falseBranch() { return next1; }
+
+    void overrideNext(BB* bb) {
+        assert(next0 && !next1);
+        next0 = bb;
+    }
+
+    void next(BB* bb) {
```
charig: Maybe initializeNext to clarify? And remove the replace or override?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -150,8 +150,8 @@ void LdConst::printArgs(std::ostream& out) {
 
 void Branch::printArgs(std::ostream& out) {
     FixedLenInstruction::printArgs(out);
-    out << " -> BB" << bb()->next1->id << " (if true) | BB" << bb()->next0->id
-        << " (if false)";
+    out << " -> BB" << bb()->trueBranch()->id << " (if true) | BB"
```
charig, line 153: Was previously printing the other way?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -63,77 +115,518 @@ void recoverCFG(rir::Code* srcCode, rir::Function* srcFunction,
...
+            mergepoints.insert(m.first);
+    return mergepoints;
 }
 
 } // namespace
 
 namespace rir {
 namespace pir {
 
+bool Rir2Pir::interpret(
```
charig: TranslateBC? SimulateBC?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -63,77 +115,518 @@ void recoverCFG(rir::Code* srcCode, rir::Function* srcFunction,
...
+    std::deque<State> worklist;
+    State cur;
+    cur.seen = true;
 
-    StackMachine state(srcFunction, srcCode);
+    std::unordered_map<Value*, CallFeedback> callFeedback;
 
-    auto popFromWorklist = [&]() {
+    Opcode* end = srcCode->endCode();
+    Opcode* finger = srcCode->code();
```
charig, line 577: finger?

charig (approved):

rir/src/compiler/pir/instruction.cpp
o-: yes, somebody decided that it was a good idea to have the branches the wrong way round, but then print them inverted :)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: if you scan through a range, then the current position is the position where you put your finger, to remember where it is :)

charig: Yes I got it. But from my pov it is still confusing whenever you see later in the code references to finger. Finger of what? I'd prefer currentOp or sth like that. Anyway, not sth fundamental to begin and endless discussion about naming :P

PR #202, opened 2018-08-24
[wip] Deoptimizer 3

o-:
this will be about inlining functions with deopts, but for that i need to chain safepoints from inlinee and caller.

### DIFF
--- a/rir/src/compiler/opt/cleanup_safepoint.cpp
+++ b/rir/src/compiler/opt/cleanup_safepoint.cpp
@@ -4,14 +4,10 @@
 #include "../util/cfg.h"
 #include "../util/visitor.h"
 
-#include <unordered_set>
-
 namespace rir {
 namespace pir {
 
 void CleanupSafepoint::apply(Closure* function) const {
-    std::unordered_map<Deopt*, Safepoint*> scheduled;
-
     auto apply = [](Code* code) {
         Visitor::run(code->entry, [&](BB* bb) {
             for (auto it = bb->begin(); it != bb->end(); ++it) {
--- a/rir/src/compiler/opt/cleanup_safepoint.cpp
+++ b/rir/src/compiler/opt/cleanup_safepoint.cpp
@@ -1,0 +1,33 @@
+#include "cleanup_safepoint.h"
+#include "../pir/pir_impl.h"
+#include "../transform/replace.h"
+#include "../util/cfg.h"
+#include "../util/visitor.h"
+
+#include <unordered_set>
+
+namespace rir {
+namespace pir {
+
+void CleanupSafepoint::apply(Closure* function) const {
+    std::unordered_map<Deopt*, Safepoint*> scheduled;
+
+    auto apply = [](Code* code) {
+        Visitor::run(code->entry, [&](BB* bb) {
+            for (auto it = bb->begin(); it != bb->end(); ++it) {
+                if (auto sp = Safepoint::Cast(*it)) {
+                    if (sp->unused())
+                        it = bb->remove(it);
+                }
+                if (it == bb->end())
+                    break;
+            }
+        });
+    };
+    apply(function);
+    for (auto& p : function->promises)
+        if (p)
+            apply(p);
+}
+} // namespace pir
+} // namespace rir
--- a/rir/src/compiler/opt/delay_env.h
+++ b/rir/src/compiler/opt/delay_env.h
@@ -18,7 +18,7 @@ class DelayEnv : public PirTranslator {
   public:
     DelayEnv() : PirTranslator("Delay Environment"){};
 
-    void apply(Closure* function) override;
+    void apply(Closure* function) const final override;
 };
 }
 }
--- a/rir/src/compiler/translations/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir.cpp
@@ -257,10 +257,18 @@ class SSAAllocator {
                         break;
                     bool argsInRightOrder = true;
                     phi->eachArg([&](BB* in, Value* v) {
-                        argsInRightOrder = argsInRightOrder && in->isJmp() &&
-                                           in->next() == bb &&
-                                           in->size() >= pos &&
-                                           *(in->end() - pos) == v;
+                        argsInRightOrder =
+                            argsInRightOrder &&
+                            /* 1. the phi input block must not be a branch block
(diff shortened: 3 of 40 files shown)


rir/src/compiler/opt/cleanup_safepoint.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "../util/cfg.h"
+#include "../util/visitor.h"
+
+#include <unordered_set>
+
+namespace rir {
+namespace pir {
+
+void CleanupSafepoint::apply(Closure* function) const {
+    std::unordered_map<Deopt*, Safepoint*> scheduled;
```
charig: What is this for?

rir/src/compiler/opt/cleanup_safepoint.cpp
```diff
@@ -0,0 +1,33 @@
+#include "cleanup_safepoint.h"
+#include "../pir/pir_impl.h"
+#include "../transform/replace.h"
+#include "../util/cfg.h"
+#include "../util/visitor.h"
+
+#include <unordered_set>
```
charig: ?

rir/src/compiler/opt/delay_env.h
```diff
@@ -18,7 +18,7 @@ class DelayEnv : public PirTranslator {
   public:
     DelayEnv() : PirTranslator("Delay Environment"){};
 
-    void apply(Closure* function) override;
+    void apply(Closure* function) const final;
```
charig: Isn't it possible to keep the override?

charig: Same for all the applies...

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -257,10 +257,10 @@ class SSAAllocator {
                         break;
                     bool argsInRightOrder = true;
                     phi->eachArg([&](BB* in, Value* v) {
-                        if (in->next0 != bb || in->size() < pos ||
-                            *(in->end() - pos) != v) {
-                            argsInRightOrder = false;
-                        }
+                        argsInRightOrder = argsInRightOrder && in->isJmp() &&
```
charig: This is quite criptic. Does a function argsInRightOrder makes sense in BBs?

rir/src/utils/configurations.cpp
```diff
@@ -23,54 +26,43 @@ void Configurations::parseINIFile() {
...
-        } else if (optimizationName == "delayInstructions") {
-            optimizations.insert(
-                new Optimization(new pir::DelayInstr(), order));
-        } else if (optimizationName == "elideEnvironments") {
-            optimizations.insert(new Optimization(new pir::ElideEnv(), order));
-        } else if (optimizationName == "delayEnvironments") {
-            optimizations.insert(new Optimization(new pir::DelayEnv(), order));
+    for (int j = 0; j < 2; ++j) {
+        for (int i = 0; i < 3; ++i) {
+            addDefaultOpt();
```
charig: Are we sure we want every optimization to run 6 times?

rir/src/utils/configurations.cpp
```diff
@@ -23,54 +26,43 @@ void Configurations::parseINIFile() {
...
+    PARSE_OPT("forceDominance", front, ForceDominance());
+    PARSE_OPT("scopeResolution", front, ScopeResolution());
+    PARSE_OPT("delayInstructions", front, DelayInstr());
+    PARSE_OPT("elideEnvironments", front, ElideEnv());
+    PARSE_OPT("delayEnvironments", front, DelayEnv());
+    PARSE_OPT("inline", front, Inline());
+    PARSE_OPT("cleanup", back, Cleanup());
+    PARSE_OPT("cleanupSafepoints", back, CleanupSafepoint());
+#undef PARSE_OPT
+    for (auto& optlist : opts)
```
charig, line 44: Does this iterator respect the order?

charig (approved):

charig (commented):
I've just seen tests travis is failing. It was ok for #201.

rir/src/utils/configurations.cpp
o-: yes, opts is a `map`. maps are ordered by keys. the keys are of type short, so they will be ordered by increasing value. Then the `optlist.second` is a `deque`, which is sortof a `vector` you can insert from both sides. It is ordered too.

o-: to schedule cleanups after normal passes (if they have the same order), i use `push_front` for normal opts and `push_back` for cleanups. This should have the same semantics as your code had.

for example `foo=2; bla=3; cleanup=3;` will give you the following opts map:
`2 -> [foo], 3 -> [bla, cleanup]`

rir/src/utils/configurations.cpp
o-: it's less than what we had before. before every opt was only once in the list, but optimizeModule iterated 2*6 times over the list.

o-: and since the cleanupsafepoint would have been another special case in optimizeModule, i decided to remove all scheduling from optimizeModule and solely use the list of optimizations for scheduling.

So previously the scheduling was a combination of the cofiguration and the logic in optimizeModule. Now it is only controlled by the configuration and optimizeModule just stupidly iterates over the list.

rir/src/compiler/opt/delay_env.h
o-: `final > override`

o-: I am wrong, you are right. (https://en.cppreference.com/w/cpp/language/final)

rir/src/interpreter/interp.cpp
```diff
@@ -2569,15 +2569,13 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
...
-                SET_NAMED(val, 2);
-            }
+            ENSURE_NAMEDMAX(val);
             NEXT();
         }
 
         INSTRUCTION(make_unique_) {
             SEXP val = ostack_top(ctx);
-            if (NAMED(val) == 2) {
+            if (NAMED(val) > 2) {
```
JanJecmen, line 2578: @o- I know it's already merged, but why is this `>` and not `>=`?

o-: uh, good catch. thanks

PR #204, opened 2018-08-28
downgrade to cxx14?

o-:
my poor jessie machine does not have cxx17. would you mind if we
hold back with 17?


mhyee:
I'm fine with this, but let's see what @JanJecmen says.

I think C++14 is a good compromise between old and stable (it's been in Clang for 5 years and GCC for 3 years) and updating to a more recent toolchain.

Oh, and you might also want to update `.ycm_extra_conf.py`.

PR #209, opened 2018-09-04
Inlining of Safepoints

o-:
this supersedes #206

### DIFF
--- a/rir/src/compiler/translations/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir.cpp
@@ -737,7 +737,7 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
 
                 // Step one: load and set env
                 if (!Phi::Cast(instr)) {
-                    if (instr->hasEnv() && !explicitEnvValue(instr)) {
+                    if (instr->accessesEnv() && !explicitEnvValue(instr)) {
                         // If the env is passed on the stack, it needs
                         // to be TOS here. To relax this condition some
                         // stack shuffling would be needed.
@@ -757,7 +757,7 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 // Step two: load the rest
                 if (!Phi::Cast(instr)) {
                     instr->eachArg([&](Value* what) {
-                        if (instr->hasEnv() && instr->env() == what) {
+                        if (instr->accessesEnv() && instr->env() == what) {
                             if (explicitEnvValue(instr))
                                 loadEnv(it, what);
                         } else {
@@ -986,16 +986,14 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
             case Tag::Deopt: {
                 assert(Safepoint::Cast(*(it - 1)) &&
                        "Deopt MUST be scheudled after Safepoint");
+                return;
             }
             case Tag::Safepoint: {
-                assert(Deopt::Cast(*(it + 1)) &&
-                       "Unused Safepoint must be removed");
                 auto sp = Safepoint::Cast(instr);
-                assert(sp->frames.size() == 1 &&
-                       "rir deopt cannot synthesize frames yet");
-                auto frame = sp->frames[0];
-                cs << BC::deopt(frame.pc, frame.code);
-                return;
+                assert(!sp->unused() && "Unused Safepoint must be removed");
+                assert(!sp->next() && "rir deopt cannot synthesize frames yet");
+                cs << BC::deopt(sp->pc, sp->code);
+                break;
             }
             // values, not instructions
             case Tag::Missing:
--- a/rir/src/compiler/translations/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir.cpp
@@ -9,6 +9,7 @@
 
 #include <algorithm>
 #include <iomanip>
+#include <sstream>
 
 // #define DEBUGGING
 #define ALLOC_DEBUG 1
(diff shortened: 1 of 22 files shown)


o-:
The general strategy for safepoints is the following:
1. At the places where we want to deopt and after every call instruction we place a safepoint. Safepoints need to be added in the first rir_2_pir step. A safepoint captures the environment and all arguments on the stack. Thus, with the information of the safepoint it is possible to switch back to unoptimized rir.
2. A deopt instruction refers to a safepoint. The deopt is unconditional. To make it conditional, place it behind a branch.
3. When we inline a safepoint, then the inlinee safepoint would only help to deoptimize the inlinee. But the inlinee needs to be able to return to the deoptimized caller. Therefore we use the safepoint that we put after the call instruction of the caller, and chain it to the safepoints of the inlinee.
4. All pir optimizations are more or loss oblivious to safepoints and deopts. They just need to know that deopt is a returning instruction. Data dependencies are encoded as normal pir arguments and values.
5. For pir_2_rir the disjoint safepoints and deopts are a pain to implement. Thus there is a safepoint scheduling phase. In that phase the deopt instruction is replaced by a `ScheduledDeopt`, which is a super-instruction that contains the transitive closure of deopt and all necessary safepoints. It takes all arguments of all the safepoints in the right order, such that they can be passed to the rir `deopt_` instruction.
6. The rir deopt instruction uses the `DeoptMetadata` frame description to deopt. The top frame is eagerly replaced and we continue in unoptimized code. The remaining frames are remembered. When the rir interpreter returns from a function, it checks whether there are frames to synthesize. In this case it will create one frame and continue to execute this one.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -694,8 +699,12 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                             std::cerr << " (" << tagToStr(what->tag) << ")\n";
                             assert(false);
                         }
-                        if (!alloc.onStack(what))
+                        if (!alloc.onStack(what)) {
                             cs << BC::ldloc(alloc[what]);
+                            std::stringstream ss;
+                            what->printRef(ss);
+                            cs.addSrc(Rf_install(ss.str().c_str()));
```
JanJecmen: What is this? We add source to the load that is just the symbol of the PIR value? Cannot this blow up if someone accesses this from R?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -714,14 +723,18 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
```
JanJecmen: maybe refactor into something like `cs.addSrc(pirValueSymbol(what))`

rir/src/interpreter/interp.cpp
```diff
@@ -2393,13 +2390,38 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
...
+                std::cout << "  - env\n";
+                Rf_PrintValue(ostack_at(ctx, pos++));
+                for( size_t j = 0; j < m->frames[i].stackSize; ++j) {
+                    std::cout << "  - stack " << j <<"\n";
+                    Rf_PrintValue(ostack_at(ctx, pos++));
+                }
+            }
+#endif
+
+            if (m->numFrames > 1) {
```
JanJecmen: this `if` is redundant

rir/src/interpreter/interp.cpp
```diff
@@ -2652,9 +2674,29 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env,
     }
 
 eval_done:
+    while (!synthesizeFrames.empty()) {
```
JanJecmen, line 2690: How do we get here from the deopt_ instruction? I thought this would be done immediately at the deopt point. Cannot there be some call in between the deopt and here that would inspect the missing frames?

JanJecmen (commented):
Cool!
A question, why is the safepoint *after* call instructions and not before? When we deoptimize, don't we want the state before the inlinee started executing?

rir/src/compiler/translations/pir_2_rir.cpp
o-: that's just the name of the pir value, it's very useful for debugging... If you want I can put it behind a debug flag

rir/src/interpreter/interp.cpp
o-: yes, but doing it eagerly is hard, so that was my first approach, and it does not seem to break immediately... but maybe I should try harder to create them eagerly

o-:
> A question, why is the safepoint after call instructions and not before? When we deoptimize, don't we want the state before the inlinee started executing?

Well, not really. Because when we are in the middle of the inlinee, we have already consumed the arguments to the function. so we are actually closer to the state after the function, except the result variable is still missing.

so you could either use: statepoint before call, and remove all arguments; or statepoint after call and remove result, that is not yet there

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: no need

rir/src/interpreter/interp.cpp
JanJecmen: I am fine with dealing with it after something breaks

o-:
@JanJecmen are you happy with the current state?

JanJecmen:
yes, let's go with it

PR #210, opened 2018-09-04
Phi placement fix

JanJecmen:


JanJecmen:
In one of the mac builds, it said it exited with 0, but then that no output was received for 10 mins and it failed... After restarting the build it passed. Do we know what can cause this?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -116,6 +116,9 @@ class TheScopeResolution {
                             !ldfun) {
                             auto hasAllInputs = [&](BB* load) -> bool {
                                 return aval.checkEachSource([&](ValOrig& src) {
+                                    // we cannot move the phi above its src
+                                    if (src.origin->bb() == load)
```
charig, line 120: Sorry, it is already merged, but shouldn't this check goes inside isPredecessor?

JanJecmen: I am not sure, in some sense they are predecessors (via a cycle) and this is a special case that we don't want to cross the definition..

PR #211, opened 2018-09-07
Re-introduce subassign bytecodes

o-:
this patch adds subassign1 and subassign2 again. Those bytecodes
support `[<-` and `[[<-`, in the case of one dimensional lists and
vectors.

This patch additionally enables object support for them.

### DIFF
--- a/rir/src/compiler/debugging/stream_logger.h
+++ b/rir/src/compiler/debugging/stream_logger.h
@@ -51,6 +51,7 @@ class StreamLogger {
     void innerHeader(rir::Function*, std::string);
 
     std::ostream& getLog(rir::Function* function) {
+        return std::cout;
         if (!streams.count(function))
             startLogging(function);
         return *streams.at(function);
--- a/rir/src/compiler/translations/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir.cpp
@@ -863,8 +863,8 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 break;
             }
             case Tag::Subassign2_1D: {
-                auto res = Subassign2_1D::Cast(instr);
-                cs << BC::subassign2(res->sym);
+                cs << BC::subassign2();
+                cs.addSrcIdx(instr->srcIdx);
                 break;
             }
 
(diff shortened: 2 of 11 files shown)


rir/src/compiler/debugging/stream_logger.h
```diff
@@ -51,6 +51,7 @@ class StreamLogger {
     void innerHeader(rir::Function*, std::string);
 
     std::ostream& getLog(rir::Function* function) {
+        return std::cout;
```
JanJecmen, line 54: Probably a leftover from debugging?

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -863,8 +863,8 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 break;
             }
             case Tag::Subassign2_1D: {
-                auto res = Subassign2_1D::Cast(instr);
-                cs << BC::subassign2(res->sym);
+                cs << BC::subassign2();
```
JanJecmen, line 866: where does the symbol parameter disappear?

rir/src/ir/Compiler.cpp
```diff
@@ -339,247 +339,83 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
...
+                        cs.addSrc(ast);
+
+                        // store the result as "target"
+                        cs << (superAssign ? BC::stvarSuper(target)
+                                           : BC::stvar(target));
+
+                        cs << BC::invisible();
+                        return true;
+                    }
+                }
```
JanJecmen, line 414: The rest is deleted intentionally?

JanJecmen (approved):
looks good

rir/src/compiler/translations/pir_2_rir.cpp
o-: It was used solely for a hack that I removed (see inter.cpp, where we would skip the assignment bytecode after the subset bytecode if the named count was 0...)

rir/src/ir/Compiler.cpp
o-: yeah, it's bitrotted anyway and we can loot into the git history if we ever want.

PR #212, opened 2018-09-07
Binops no env

charig:
- In case there is type feedback at compilation time from rir to pir, and this feedback indicates that previously there were no objects observed, the compiler generate speculatevely a binop without an environment and add guards to deopt accordingly.

- In this PR I also add a PrintAfterOptimizations flag to de logger. The problem was that if there is no compilation back from pir to rir, the only way to see the pir code after optimization was printing all the optimization phases which overloads the output.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, debugFlags = pir.debugFlags()) {
-    .Call("pir_compile", what, debugFlags)
+pir.compile <- function(what, debugFlags) {
+    .Call("pir_compile", what, if (missing(debugFlags)) NULL else debugFlags)
 }
 
 pir.tests <- function() {
@@ -32,7 +32,6 @@ pir.tests <- function() {
 pir.debugFlags <- function(ShowWarnings = FALSE,
                            DryRun = FALSE,
                            PreserveVersions = FALSE,
-                           DebugAllocator = FALSE,
                            PrintIntoFiles = FALSE,
                            PrintIntoStdout = FALSE,
                            PrintEarlyRir = FALSE,
@@ -40,21 +39,22 @@ pir.debugFlags <- function(ShowWarnings = FALSE,
                            PrintOptimizationPasses = FALSE,
                            PrintPirAfterOpt = FALSE,
                            PrintCSSA = FALSE,
-                           PrintLivenessIntervals = FALSE,
+                           PrintAllocator = FALSE,
                            PrintFinalPir = FALSE,
                            PrintFinalRir = FALSE ) {
     # !!!  This list of arguments *must* be exactly equal to the   !!!
     # !!!    LIST_OF_PIR_DEBUGGING_FLAGS in compiler/debugging.h   !!!
     .Call("pir_debugFlags", ShowWarnings, DryRun, PreserveVersions,
-          DebugAllocator, PrintIntoFiles, PrintIntoStdout, PrintEarlyRir, PrintEarlyPir, PrintOptimizationPasses,
-          PrintPirAfterOpt, PrintCSSA, PrintLivenessIntervals, PrintFinalPir, PrintFinalRir,
+          PrintIntoFiles, PrintIntoStdout, PrintEarlyRir, PrintEarlyPir,
+          PrintOptimizationPasses, PrintPirAfterOpt, PrintCSSA, PrintAllocator, PrintFinalPir,
+          PrintFinalRir,
           # wants a dummy parameter at the end for technical reasons
           NULL)
 }
 
 # sets the default debug options for pir compiler
 pir.setDebugFlags <- function(debugFlags = pir.debugFlags()) {
-    .Call("pir_setDebugFlags", debugFlags)
+    invisible(.Call("pir_setDebugFlags", debugFlags))
 }
 
 # compiles code of the given file and returns the list of compiled version.
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -38,15 +38,16 @@ pir.debugFlags <- function(ShowWarnings = FALSE,
                            PrintEarlyRir = FALSE,
                            PrintEarlyPir = FALSE,
                            PrintOptimizationPasses = FALSE,
+                           PrintPirAfterOpt = FALSE,
                            PrintCSSA = FALSE,
                            PrintLivenessIntervals = FALSE,
                            PrintFinalPir = FALSE,
-                           PrintFinalRir = FALSE) {
+                           PrintFinalRir = FALSE ) {
     # !!!  This list of arguments *must* be exactly equal to the   !!!
     # !!!    LIST_OF_PIR_DEBUGGING_FLAGS in compiler/debugging.h   !!!
     .Call("pir_debugFlags", ShowWarnings, DryRun, PreserveVersions,
           DebugAllocator, PrintIntoFiles, PrintIntoStdout, PrintEarlyRir, PrintEarlyPir, PrintOptimizationPasses,
-          PrintCSSA, PrintLivenessIntervals, PrintFinalPir, PrintFinalRir,
+          PrintPirAfterOpt, PrintCSSA, PrintLivenessIntervals, PrintFinalPir, PrintFinalRir,
           # wants a dummy parameter at the end for technical reasons
           NULL)
 }
--- a/rir/src/compiler/pir/env.cpp
+++ b/rir/src/compiler/pir/env.cpp
@@ -12,11 +12,13 @@ void Env::printRef(std::ostream& out) {
     if (this == notClosed()) {
         out << "?";
         return;
-    }
-    if (this == nil()) {
+    } else if (this == elided()) {
+        out << "elided";
+        return;
+    } else if (this == nil()) {
         out << "nil";
         return;
-    }
+    } 
     assert(rho);
     std::string val;
     {
@@ -28,7 +30,7 @@ void Env::printRef(std::ostream& out) {
 }
 
 bool Env::isStaticEnv(Value* v) {
-    return Env::Cast(v) && v != Env::notClosed() && v != Env::nil();
+    return Env::Cast(v) && v != Env::notClosed() && v != Env::nil() && v != Env::elided();
 }
 
 bool Env::isPirEnv(Value* v) {
--- a/rir/src/compiler/pir/env.cpp
+++ b/rir/src/compiler/pir/env.cpp
@@ -46,7 +46,7 @@ Value* Env::parentEnv(Value* e) {
     if (Cast(e))
         return Cast(e)->parent;
     if (MkEnv::Cast(e))
-        return MkEnv::Cast(e)->parent();
+        return MkEnv::Cast(e)->lexicalEnv();
     assert(false);
     return nullptr;
 }
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -4,6 +4,7 @@
 #include "R/r.h"
 #include "env.h"
 #include "instruction_list.h"
+#include "ir/Deoptimization.h"
 #include "pir.h"
 #include "singleton_values.h"
 #include "tag.h"
@@ -101,13 +102,10 @@ class Instruction : public Value {
     virtual bool hasEffect() const = 0;
     virtual bool changesEnv() const = 0;
     virtual bool leaksEnv() const = 0;
+    virtual bool mayAccessEnv() const = 0;
     virtual bool hasEnv() const = 0;
-    virtual bool accessesEnv() const = 0;
 
     virtual size_t nargs() const = 0;
-    virtual size_t envSlot() const = 0;
-    virtual Value* env() const = 0;
-    virtual void env(Value*) = 0;
 
     virtual Instruction* clone() const = 0;
 
@@ -182,6 +180,22 @@ class Instruction : public Value {
         }
         return nullptr;
     }
+
+    virtual Value* env() const {
+        assert(!mayAccessEnv() &&
+               "subclass must override env() if it uses env");
+        assert(false && "this instruction has no env");
+    }
+    virtual void env(Value* env) {
+        assert(!mayAccessEnv() &&
+               "subclass must override env() if it uses env");
+        assert(false && "this instruction has no env");
+    }
+    virtual size_t envSlot() const {
+        assert(!mayAccessEnv() &&
+               "subclass must override envSlot() if it uses env");
+        assert(false && "this instruction has no env");
+    }
 };
 
 template <Tag ITAG, class Base, Effect EFFECT, EnvAccess ENV, class ArgStore>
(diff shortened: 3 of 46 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -209,11 +210,13 @@ class InstructionImplementation : public Instruction {
         bool LeaksEnv;
         bool HasEnv;
         bool AccessesEnv;
+        bool mayElideEnv;
```
o-: I don't really like this new description field. It feels kind of redundant. Can we do this instead:

```
diff --git a/rir/src/compiler/pir/instruction.h b/rir/src/compiler/pir/instruction.h
index 2864b53..c8055f0 100644
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -4,6 +4,7 @@
 #include "R/r.h"
 #include "instruction_list.h"
 #include "pir.h"
+#include "env.h"
 #include "singleton_values.h"
 #include "tag.h"
 #include "value.h"
@@ -73,7 +74,6 @@ enum class EnvAccess : uint8_t {
     Read,
     Write,
     Leak,
-    NoneOrLeak,
 };
 
 // Effect that can be produced by an instruction.
@@ -208,20 +208,21 @@ class InstructionImplementation : public Instruction {
         bool HasEffect;
         bool ChangesEnv;
         bool LeaksEnv;
-        bool HasEnv;
+        bool MightHaveEnv;
         bool AccessesEnv;
-        bool mayElideEnv;
     };
 
     static constexpr InstrDescription Description = {
         EFFECT > Effect::None,    ENV >= EnvAccess::Write,
         ENV == EnvAccess::Leak,   ENV > EnvAccess::None,
-        ENV > EnvAccess::Capture, ENV >= EnvAccess::NoneOrLeak};
+        ENV > EnvAccess::Capture};
 
     bool hasEffect() const final { return Description.HasEffect; }
     bool changesEnv() const final { return Description.ChangesEnv; }
     bool leaksEnv() const final { return Description.LeaksEnv; }
-    bool hasEnv() const final { return Description.HasEnv; }
+    bool hasEnv() const final {
+        return Description.MightHaveEnv && env() != Env::elided();
+    }
     bool accessesEnv() const final { return Description.AccessesEnv; }
 
     static const Base* Cast(const Value* i) {
```

(not a complete patch)

o-: so basically, the static description field is `mightHaveEnv` and we use that one whenever we need to know the static size of the instruction.

The static description `hasEnv` goes away, because morally with this change, we cannot tell statically anymore.

then the dynamic `isEnv()` function will check if it's actually present.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -573,6 +585,11 @@ void Rir2Pir::translate(rir::Code* srcCode, Builder& insert,
 
     std::unordered_map<Value*, CallFeedback> callFeedback;
 
+    /* Captures the last type feedback observed. The assumption is that this
```
o-: I don't think that this assumption works in general. In particular typefeedback instructions can be inserted anywhere, not all binops might even have one, you might have the typefeedback hoisted. Basically they should be completely independent of the usage. I would prefer if we can collect them in a map like in the call feedback case.

rir/src/compiler/util/builder.cpp
```diff
@@ -57,12 +57,16 @@ Safepoint* Builder::registerSafepoint(rir::Code* srcCode, Opcode* pos,
...
-void Builder::deoptUnless(Value* condition, rir::Code* srcCode, Opcode* pos,
-                          const RirStack& stack) {
+void Builder::conditionalDeopt(Value* condition, rir::Code* srcCode,
+                               Opcode* pos, const RirStack& stack,
+                               bool unlessInsteadOfIf) {
     add(new Branch(condition));
     auto cont = createBB();
     auto fail = createBB();
-    setBranch(cont, fail);
+    if (unlessInsteadOfIf)
```
o-: what?

rir/src/ir/RuntimeFeedback.h
```diff
@@ -73,6 +75,17 @@ struct TypeFeedback {
                 seen[numTypes++] = type;
         }
     }
+
+    bool observedObject() {
+        bool answer = false;
+        for (const auto& record : seen) {
+            if (record.isObj()) {
+                answer = true;
```
o-: `return true`

o- (commented):
Hey, thanks for the patch.

rir/src/compiler/util/builder.cpp
```diff
@@ -57,12 +57,16 @@ Safepoint* Builder::registerSafepoint(rir::Code* srcCode, Opcode* pos,
     return sp;
 };
 
-void Builder::deoptUnless(Value* condition, rir::Code* srcCode, Opcode* pos,
-                          const RirStack& stack) {
+void Builder::conditionalDeopt(Value* condition, rir::Code* srcCode,
+                               Opcode* pos, const RirStack& stack,
+                               bool unlessInsteadOfIf) {
```
JanJecmen: This name really stretches my brain ;)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -415,7 +417,17 @@ bool Rir2Pir::compileBC(
     case Opcode::Op: {                                                         \
         auto rhs = pop();                                                      \
         auto lhs = pop();                                                      \
-        push(insert(new Name(lhs, rhs, env, srcIdx)));                         \
+        if (typeFeedback[0].numTypes > 0 && typeFeedback[1].numTypes > 0 &&    \
```
JanJecmen: Wouldn't this work even if we don't have the type feedback? `if (isobj(a) || isobj(b)) then env else elided`

rir/src/compiler/util/builder.cpp
charig: conditionalDeopt or unlessInsteadOfIf?

JanJecmen: `unlessInsteadOfIf`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
charig: Yes, sure it will work. Semantics are preserved. But then whenever binops are used with objects it will trigger a deopt. Deopt should be avoided. Unless we know beforehand that objects are very infrequent?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
charig: So, for this issue I think we have a problem with the current design then. The feedback is stored inside the record_binop instruction. But there is no way to connect a binop with its record_binop. Then when translating the record_binop to pir I may be able to store the feedback in a map on the opcode* or something like that. But then, when translating the binop, there will be no way to find the position in the map. So, according to this design I though there was always a record_binop just before each binop...

o-: Currently we have it for this list: https://github.com/reactorlabs/rir/blob/master/rir/src/ir/Compiler.cpp#L197

But I would rather not have the assumption, that a binop must have feedback. That's why I created two separate and independent instructions.

> But there is no way to connect a binop with its record_binop.

of course there is... The record binop records the two values which are top of stack. You map pir `Value*` to type feedback. Baiscally:

```
std::unordered_map<Value*, TypeFeedback&> typeFeedback;
...
case record_binop:
    typeFeedback[at(0)] = typeFeedback[0];
    typeFeedback[at(1)] = typeFeedback[1];
```

then at any point in the compiler,  if you want to know something, you just:

```
if (typefeedback(aValue).noObject())
  doo something
if (typefeeedback(aValue).scalar())
  do something else
```

charig: The record_binop is not popping anything from the stack. That is why it seems quite coupled with a following binop concrete operation.  It assume it would not be striaghtforward to hoist this operation much as it is designed right now. 

Anyway, I submitted the change. I agree it is much more robust and will probably keep working if we change how the feedback ops are managed.

rir/R/rir.R
```diff
@@ -37,15 +37,16 @@ pir.debugFlags <- function(ShowWarnings = FALSE,
                            PrintEarlyRir = FALSE,
                            PrintEarlyPir = FALSE,
                            PrintOptimizationPasses = FALSE,
+                           PrintPirAfterOpt = FALSE,
                            PrintCSSA = FALSE,
                            PrintAllocator = FALSE,
                            PrintFinalPir = FALSE,
-                           PrintFinalRir = FALSE) {
+                           PrintFinalRir = FALSE ) {
```
mhyee, line 44: Nitpick: added space.

rir/src/compiler/pir/env.cpp
```diff
@@ -12,11 +12,13 @@ void Env::printRef(std::ostream& out) {
...
-    }
-    if (this == nil()) {
+    } else if (this == elided()) {
+        out << "elided";
+        return;
+    } else if (this == nil()) {
         out << "nil";
         return;
-    }
+    }
```
mhyee, line 21: Nitpick: trailing whitespace

Does clang-format check for this?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -113,6 +113,10 @@ void Rir2PirCompiler::optimizeModule() {
     module->eachPirFunction([&](Module::VersionedClosure& v) {
         assert(Verify::apply(v.current()));
     });
+#else
+    module->eachPirFunction([&](Module::VersionedClosure& v) {
+        LOGGING(logger.pirOptimizationsFinished(*(v.current())));
```
mhyee, line 118: This is a bit redundant. `LOGGING` is only defined if `ENABLE_SLOWASSERT` is defined, see `debugging.h`:

```c
#ifdef ENABLE_SLOWASSERT
#define LOGGING(code) code
#else
#define LOGGING(code) /* nothing */
#endif
```

Unrelated, but what I don't understand is a few lines up, L112, where it's `#ifndef`. Shouldn't it be `#ifdef`?

rir/src/compiler/util/builder.h
```diff
@@ -42,8 +42,8 @@ class Builder {
 
     Safepoint* registerSafepoint(rir::Code* srcCode, Opcode* pos,
                                  const RirStack& stack);
-    void deoptUnless(Value* condition, rir::Code* srcCode, Opcode* pos,
-                     const RirStack& stack);
+    void conditionalDeopt(Value* condition, rir::Code* srcCode, Opcode* pos,
+                          const RirStack& stack, bool);
```
mhyee, line 46: I know it's not strictly necessary, but for consistency, please give a name to the `bool` parameter.

rir/src/compiler/pir_tests.cpp
```diff
@@ -45,10 +44,14 @@ compile(const std::string& context, const std::string& expr, pir::Module* m,
     SEXP env = p(Rf_allocSExp(ENVSXP));
     ENCLOS(env) = super;
     eval(expr, env);
+    return env;
+}
 
-    pir::Rir2PirCompiler cmp(m, pir::DebugOptions());
+std::unordered_map<std::string, pir::Closure*>
```
mhyee, line 50: Typedef this map? It's long enough and being used in enough places that it might be worth it.

rir/src/compiler/pir_tests.cpp
```diff
@@ -45,10 +44,14 @@ compile(const std::string& context, const std::string& expr, pir::Module* m,
...
+compileRir2Pir(SEXP env, pir::Module* m){
+    pir::Rir2PirCompiler cmp(m, pir::DebugOptions() | pir::DebugFlag::PrintFinalPir);
```
mhyee, line 52: Why is `PrintFinalPir` enabled? Was this just for debugging?

rir/src/compiler/pir_tests.cpp
```diff
@@ -316,6 +370,8 @@ static Test tests[] = {
          }),
     Test("context_load",
          []() { return canRemoveEnvironment("f <- function() 123"); }),
+    Test("binop_nonobjects",
+         []() { return canRemoveEnvironmentOnBinaryWithTypeFeedback("f <- function() 1 + 2"); }),
```
mhyee, line 374: Can you add a negative test?

rir/src/compiler/pir_tests.cpp
```diff
@@ -142,6 +152,51 @@ bool compileAndVerify(const std::string& input) {
     return compileAndVerify("", input);
 }
 
+void insertTypeFeedbackForBinops(rir::Function* srcFunction, std::array<SEXP, 2> typeFeedback){
+    auto it = srcFunction->begin();
```
mhyee, line 156: `for (auto c : srcFunction) { ... }`?

rir/src/compiler/pir_tests.cpp
```diff
@@ -142,6 +152,51 @@ bool compileAndVerify(const std::string& input) {
...
+    while (it != srcFunction->end()) {
+        Opcode* end = (*it)->endCode();
```
mhyee, line 158: Just noticing now, you can ignore this comment because it's out of scope, but I wonder if it's worth adding the boilerplate to `Code` to get an `Opcode` iterator...

rir/src/compiler/pir_tests.cpp
```diff
@@ -142,6 +152,51 @@ bool compileAndVerify(const std::string& input) {
...
+        Opcode* finger = (*it)->code();
+        while (finger != end) {
+            Opcode* prev = finger;
+            BC bc = BC::advance(&finger);
+            if (bc.bc == Opcode::record_binop_) {
+                TypeFeedback* feedback = (TypeFeedback*)++prev;
```
mhyee, line 164: `(TypeFeedback*)++prev` is just confusing to read. I would prefer something more explicit, e.g. with an `imm` temporary variable.

rir/src/compiler/pir_tests.cpp
```diff
@@ -142,6 +152,51 @@ bool compileAndVerify(const std::string& input) {
...
+            }
+        } 
+        ++it;
+    }
+}
+
+extern "C" SEXP Rf_NewEnvironment(SEXP, SEXP, SEXP);
+SEXP parseCompileToRir(std::string);
+
+bool canRemoveEnvironmentOnBinaryWithTypeFeedback(const std::string& input) {
```
mhyee, line 176: Nitpicks for this function:

- The name is really long, but maybe that's OK because it's only used once
- Blank lines would improve the readability

rir/src/compiler/pir_tests.cpp
```diff
@@ -142,6 +152,51 @@ bool compileAndVerify(const std::string& input) {
...
+    std::array<SEXP,2> types;
+    SEXP res = p(Rf_allocVector(INTSXP, 1));
+    INTEGER(res)[0] = 1;
+    types[0] = res;
+    types[1] = res;
+    auto execEnv = p(Rf_NewEnvironment(R_NilValue, R_NilValue, R_GlobalEnv));
+    auto rirFun = p(parseCompileToRir(input));
+    SET_CLOENV(rirFun, execEnv);
+    Rf_defineVar(Rf_install("removeEnvInBinopTest"), rirFun, execEnv);
+    if (TYPEOF(rirFun) == CLOSXP) {
```
mhyee, line 187: `isValidClosureSEXP` already checks that the type is CLOSXP, and it unpacks the first Function from the DispatchTable.

So maybe rewrite to:

```c
rir::Function* srcFunction = isValidClosureSEXP(rirfun);
assert(srcFunction != nullptr);
insertTypeFeedbackForBinops(srcFunction, types);
```

Or maybe that's too terse?

Tangent: I'm starting to dislike `isValidClosureSEXP`. It's from before we had the `DispatchTable`, and we had a whole bunch of `isValid___SEXP` functions, and I cleaned up most of them in #198, but I think `isValidClosureSEXP` is the only one I couldn't remove.

mhyee (commented):
I missed the boat, but here are some more comments. Most of them are really minor or can be ignored.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
charig: For the logging true, but for consistency I'd keep it. We may change the LOGGING macro later.

For the ifdef yes, I am not sure but seems so. @o- , Is there any reason for verifying on "release mode"?

mhyee: I think that was a typo, and it was fixed in a recent commit/PR. Anyway, I guess my comment about the LOGGING macro is now outdated, so 🤷‍♂️ .

PR #214, opened 2018-09-10
fix some logging brokenness

JanJecmen:
* files were not working (everything went into the same file)
* files were written into .pir directory, if that didn't exist, nothing would be logged (c++ doesn't deal with directories I think) - one idea would be to have the directory as a debug option and create the directory from R
* files were opened for the whole logger lifetime, now use stringstreams instead and only write at the end
* contradicting options - both print into files and stdout would only print into files, now it does both
* allocator was not implemented
* use raii to clean up the logger

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, debugFlags = pir.debugFlags()) {
-    .Call("pir_compile", what, debugFlags)
+pir.compile <- function(what, debugFlags) {
+    .Call("pir_compile", what, if (missing(debugFlags)) NULL else debugFlags)
 }
 
 pir.tests <- function() {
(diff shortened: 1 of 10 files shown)


rir/R/rir.R
```diff
@@ -20,8 +20,8 @@ rir.compile <- function(what) {
 }
 
 # optimizes given rir compiled closure
-pir.compile <- function(what, debugFlags = pir.debugFlags()) {
-    .Call("pir_compile", what, debugFlags)
+pir.compile <- function(what, debugFlags) {
+    .Call("pir_compile", what, if (missing(debugFlags)) NULL else debugFlags)
```
charig, line 24: Null or R_NilValue?

rir/R/rir.R
```diff
@@ -32,28 +32,28 @@ pir.tests <- function() {
...
+          PrintOptimizationPasses, PrintCSSA, PrintAllocator, PrintFinalPir,
+          PrintFinalRir,
           # wants a dummy parameter at the end for technical reasons
           NULL)
 }
 
 # sets the default debug options for pir compiler
 pir.setDebugFlags <- function(debugFlags = pir.debugFlags()) {
-    .Call("pir_setDebugFlags", debugFlags)
+    invisible(.Call("pir_setDebugFlags", debugFlags))
```
charig, line 56: Why invisible is needed here?

charig (approved):

rir/R/rir.R
JanJecmen: it's not needed, it just hides the returned NULL so that it isn't printed in the console

rir/R/rir.R
JanJecmen: `NULL` in R ~ `R_NilValue` in C

charig: He, sorry I read it like c code...

o-:
thanks a lot

PR #216, opened 2018-09-12
print labels in disassembly

JanJecmen:
This attempts to simulate the disassembly format that CodeEditor used, ie. print labels in the code and in the jump instructions (instead of the offsets we have now - I was lazy to do the math in my head all the time)

- removes unused opcode for label
- in disassembly, prints labels and jump targets


o-:
nice one

PR #217, opened 2018-09-12
Attribute test fix in extract2_1_

JanJecmen:
- `getAttrib` is covered by the broader test for the presence of attributes
- the test `ATTRIB(val)` checked for `nullptr`, should be `R_NilValue`

This caused the fast path to always be skipped


o-:
nice catch.

(that's how you do it: add a bogus test to skip the optimization. then let others debug why it fails when you enable optimizations. only in this case the optimization happened to work :) )

PR #218, opened 2018-09-12
rewrite the logger

o-:
* have a LogStream class that captures the stream of one particular
  function that we can pass around.
* use OO to distinguish between our 3 different logging modes.
* some best effort for keeping track of function names.
* print in 3 steps: 1. early pir, 2. optimizations, 3. final pir.
  this allows us to at least get some output if we crash
  in the middle.
* fix all the logging to stdout.
* don't create unnecessary stringstreams.

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -15,6 +15,7 @@
 #include "ir/Compiler.h"
 
 #include <memory>
+#include <string>
 
 using namespace rir;
 
@@ -108,8 +109,41 @@ REXPORT SEXP pir_debugFlags(
     return res;
 }
 
-static pir::DebugOptions PirDebug(
-    getenv("PIR_VERBOSE") ? std::stoul(getenv("PIR_VERBOSE"), nullptr, 0) : 0);
+static long getInitialDebugOptions() {
+    auto verb = getenv("PIR_DEBUG");
+    if (!verb)
+        return 0;
+    std::istringstream in(verb);
+
+    pir::DebugOptions flags;
+    while (!in.fail()) {
+        std::string opt;
+        std::getline(in, opt, ',');
+        if (opt.empty())
+            continue;
+
+        bool success = false;
+
+#define V(flag)                                                                \
+    if (opt.compare(#flag) == 0) {                                             \
+        success = true;                                                        \
+        flags = flags | pir::DebugFlag::flag;                                  \
+    }
+        LIST_OF_PIR_DEBUGGING_FLAGS(V)
+#undef V
+        if (!success) {
+            std::cerr << "Unknown PIR debug flag " << opt << "\n"
+                      << "Valid flags are:\n";
+#define V(flag) std::cerr << "- " << #flag << "\n";
+            LIST_OF_PIR_DEBUGGING_FLAGS(V)
+#undef V
+            exit(1);
+        }
+    }
+    return flags.to_ulong();
+}
+
+static pir::DebugOptions PirDebug(getInitialDebugOptions());
 
 REXPORT SEXP pir_setDebugFlags(SEXP debugFlags) {
     if (TYPEOF(debugFlags) != INTSXP || Rf_length(debugFlags) < 1)
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -32,17 +32,9 @@ REXPORT SEXP rir_disassemble(SEXP what, SEXP verbose) {
         if (!t->available(entry))
             continue;
         Function* f = t->at(entry);
-        std::stringstream output;
-        output << "= vtable slot <" << entry << "> (" << f << ", invoked "
-               << f->invocationCount << ") =\n";
-        f->body()->disassemble(output);
-        for (auto c : *f) {
-            if (c != f->body()) {
-                output << "\n [Prom (index " << c->index << ")]\n";
-                c->disassemble(output);
-            }
-        }
-        std::cout << output.str();
+        std::cout << "= vtable slot <" << entry << "> (" << f << ", invoked "
+                  << f->invocationCount << ") =\n";
+        f->disassemble(std::cout);
     }
 
     return R_NilValue;
@@ -127,7 +119,7 @@ REXPORT SEXP pir_setDebugFlags(SEXP debugFlags) {
     return R_NilValue;
 }
 
-SEXP pirCompile(SEXP what, pir::DebugOptions debug) {
+SEXP pirCompile(SEXP what, const std::string& name, pir::DebugOptions debug) {
 
     if (!isValidClosureSEXP(what)) {
         Rf_error("not a compiled closure");
@@ -142,16 +134,19 @@ SEXP pirCompile(SEXP what, pir::DebugOptions debug) {
 
     PROTECT(what);
 
+    bool dryRun = debug.includes(pir::DebugFlag::DryRun);
+    bool preserveVersions = debug.includes(pir::DebugFlag::PreserveVersions);
     // compile to pir
     pir::Module* m = new pir::Module;
-    pir::Rir2PirCompiler cmp(m, debug);
-    cmp.compileClosure(what,
+    pir::StreamLogger logger(debug);
+    pir::Rir2PirCompiler cmp(m, logger);
+    cmp.compileClosure(what, name,
                        [&](pir::Closure* c) {
-                           cmp.optimizeModule();
+                           cmp.optimizeModule(logger, preserveVersions);
 
                            // compile back to rir
-                           pir::Pir2RirCompiler p2r(debug, cmp.getLogger());
-                           p2r.compile(c, what);
+                           pir::Pir2RirCompiler p2r(logger);
+                           p2r.compile(c, what, dryRun);
                        },
                        [&]() {
                            if (debug.includes(pir::DebugFlag::ShowWarnings))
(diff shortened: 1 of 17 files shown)


rir/src/runtime/Function.h
```diff
@@ -98,6 +98,8 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
         return 0;
     }
 
+    void disassemble(std::ostream&);
```
mhyee, line 101: Missing a file. `Function::disassemble` isn't defined anywhere.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -13,145 +13,181 @@ namespace pir {
...
 }
 
-void StreamLogger::compilationInit(rir::Function* function) {
-    header("Start compiling:", function, getLog(function), true);
+FileLogStream::~FileLogStream() { fstream.close(); }
+
+LogStream& StreamLogger::get(Closure* cls) { return get(cls->rirVersion()); }
+
+LogStream& StreamLogger::begin(SEXP cls, const std::string& name) {
+    return begin(DispatchTable::unpack(BODY(cls))->at(0), name);
```
mhyee: `DispatchTable::unpack(BODY(cls))->first()`

rir/src/compiler/pir_tests.cpp
```diff
@@ -49,7 +49,8 @@ SEXP compileToRir(const std::string& context, const std::string& expr, SEXP supe
 
 std::unordered_map<std::string, pir::Closure*>
 compileRir2Pir(SEXP env, pir::Module* m){
-    pir::Rir2PirCompiler cmp(m, pir::DebugOptions() | pir::DebugFlag::PrintFinalPir);
+    pir::StreamLogger logger((pir::DebugOptions()));
```
mhyee: Extra parens?

rir/src/runtime/Function.cpp
```diff
@@ -0,0 +1,14 @@
+#include "Function.h"
+
+namespace rir {
+void Function::disassemble(std::ostream& out) {
+    body()->disassemble(out);
+    for (auto c : *this) {
+        if (c != body()) {
+            out << "\n [Prom " << std::hex << ((uintptr_t)c - (uintptr_t)this)
```
mhyee: This is outdated, since `Code` has been refactored. Use `c->index` instead of doing pointer subtraction.

(The disassembling code was originally taken from `api.cpp`, so maybe it got lost in a merge.)

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -85,32 +106,31 @@ void Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
...
 
 #ifdef ENABLE_SLOWASSERT
             assert(Verify::apply(f));
 #endif
         });
     }
-#ifndef ENABLE_SLOWASSERT
+#ifdef ENABLE_SLOWASSERT
     module->eachPirFunction([&](Module::VersionedClosure& v) {
+        LOGGING(logger.get(v.current()).pirOptimizationsFinished(v.current()));
```
mhyee: `LOGGING` can be removed, because it's redundant with `ENABLE_SLOWASSERT`.

rir/src/compiler/pir_tests.cpp
o-: there is some parser ambiguity without them. I am not sure I understand why, but they are needed.

o-: on the other hand I added the PrintFinalPir flag again, since that runs the printing code, which sometimes revealed bugs in the past. and with this re-added the () are not needed anymore :)

mhyee: Sounds like the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). Braces instead of parens would also have fixed it.

rir/src/compiler/pir_tests.cpp
```diff
@@ -49,7 +49,8 @@ SEXP compileToRir(const std::string& context, const std::string& expr, SEXP supe
 
 std::unordered_map<std::string, pir::Closure*>
 compileRir2Pir(SEXP env, pir::Module* m){
-    pir::StreamLogger logger((pir::DebugOptions()));
+    pir::StreamLogger logger(pir::DebugOptions() |
```
mhyee: Maybe it's worth using the braces, in case the `PrintFinalPir` flag is ever removed.

Also, maybe I missed it, but why override and force this flag to be enabled?

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -13,145 +13,192 @@ namespace pir {
...
-                     << std::setfill('=') << std::setw(15) << "" << std::left
-                     << std::setw(39) << header << std::setfill(' ') << "\n";
+void LogStream::failed(const std::string& msg) {
+    if (options.includes(DebugFlag::ShowWarnings)) {
+        preparePrint();
+        out << "Failed: " << msg << " in " << id << "\n";
+    }
+}
+
+void LogStream::highlightOn() { out << "\033[1;31m"; }
```
mhyee: I generally prefer defining these escape codes as constants, so it's easier to tell what colour it's supposed to be.

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -13,145 +13,192 @@ namespace pir {
...
+
+void LogStream::highlightOff() { out << "\033[0m"; }
+
+void LogStream::header() {
+    highlightOn();
+    out << "┌";
```
mhyee: Ah, fun with the box-drawing characters. :)

mhyee (commented):
I think there's a few READMEs that might need updating? It would be nice to have some examples of how to use the logger.

Especially from the REPL. I realized that I basically stopped using the debug output when I couldn't figure out what to write for `pir.compile(f, ???)`. I didn't want to type out a lot of text, when before I could just wrte `TRUE`.

rir/src/compiler/pir_tests.cpp
o-: so that the test executes the disassembly. sometimes in the past, printing pir would have segv, so it's good to execute it somewhere in the tests.

rir/src/api.cpp
```diff
@@ -142,16 +168,19 @@ SEXP pirCompile(SEXP what, pir::DebugOptions debug) {
...
+    cmp.compileClosure(what, name,
                        [&](pir::Closure* c) {
-                           cmp.optimizeModule();
+                           cmp.optimizeModule(logger, preserveVersions);
 
                            // compile back to rir
-                           pir::Pir2RirCompiler p2r(debug, cmp.getLogger());
-                           p2r.compile(c, what);
+                           pir::Pir2RirCompiler p2r(logger);
+                           p2r.compile(c, what, dryRun);
```
charig, line 183: Why add the dryRun and preserveVersions as parameters if the compiler already have the logger? To run those modes even in "release mode"?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
```diff
@@ -11,21 +11,20 @@ namespace pir {
...
-    void compileFunction(rir::Function*, FormalArgs const&, MaybeCls success,
-                         Maybe fail);
-    void optimizeModule();
-
-    StreamLogger& getLogger() { return logger; }
+    void compileClosure(SEXP, const std::string& name, MaybeCls success,
+                        Maybe fail);
+    void compileFunction(rir::Function*, const std::string& name,
+                         FormalArgs const&, MaybeCls success, Maybe fail);
+    void optimizeModule(StreamLogger& logger, bool preserveVersions = false);
```
charig, line 20: Why a different logger as parameter?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -85,32 +110,31 @@ void Rir2PirCompiler::compileClosure(rir::Function* srcFunction,
...
 
-void Rir2PirCompiler::optimizeModule() {
+void Rir2PirCompiler::optimizeModule(StreamLogger& logger,
+                                     bool preserveVersions) {
     LOGGING(size_t passnr = 0);
     for (auto& translation : translations) {
         module->eachPirFunction([&](Module::VersionedClosure& v) {
             auto f = v.current();
-            if (debug.includes(DebugFlag::PreserveVersions))
+            if (preserveVersions)
```
charig, line 119: Why surpass the preserve versions defintion of the logger?

charig (approved):

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
o-: yeah, that was an artifact of refactoring that i fixed in #219

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
o-: see above, same reason as dryRun

rir/src/api.cpp
o-: The dryRun flag feels in the wrong place, and it certainly has nothing to do with the logger. asking the logger if we are in dry run, is clearly not something the logger should be concerned with. I think in the long run we might want to remove the flag from debug options, that was sort of a preparation for this.

o-: also, as you might have noticed, I pushed the debug options into the LogStream. The idea was, that in the future we might have different logging options per function. For example, we could have a exclude/include mechanism that allows you to match for function names.

Overall the debugOptions flow like this:

```
PirDebug   ->   Logger   ->   LogStream
```

The PirDebug is the current authoritative version, the logger contains the version as it was when a compilation run was started, the log stream contains the version for one function. It's not immediately clear to me which of the 3 would be the place to ask for the compiler to find out if we are in dryRun. So for now I made it an explicit choice of the compiler client.

PR #219, opened 2018-09-12
use cppcheck on travis and enable more compiler warnings

o-:
based on #218 since both are quite invasive


.travis.yml
```diff
@@ -73,6 +79,7 @@ addons:
       - rsync
       - xorg-dev
       - valgrind
+      - cppcheck
```
mhyee, line 82: I think you have to update `install_osx.sh` so that `brew` installs `cppcheck`.

.travis.yml
```diff
@@ -93,6 +100,7 @@ before_script:
   - make -j2
 
 script:
+  - $CPPCHECK
```
mhyee, line 103: Does `cppcheck` generate any output if it passes? Right now, the Travis logs just show `$CPPCHECK`, so we don't know if the environment variable was empty of if it did anything at all.

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ set(R_COMMAND ${R_HOME}/bin/R)
 add_definitions(-g3)
 set(CMAKE_CXX_FLAGS_RELEASE "-O2")
 set(CMAKE_CXX_FLAGS_DEBUG "-O0 -DENABLE_SLOWASSERT")
-set(CMAKE_CXX_FLAGS "-Wall -std=c++14 -fno-rtti")
+set(CMAKE_CXX_FLAGS "-Wall -Wuninitialized -Wundef -Winit-self -Wcast-align -Woverloaded-virtual -Wctor-dtor-privacy -Wmissing-include-dirs -Wstrict-overflow=5 -Werror -std=c++14 -fno-rtti")
```
mhyee, line 16: Do you also need to update `.ycm_extra_conf.py`?

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -74,7 +74,7 @@ class StaticAnalysis {
 
         BB* bb = i->bb();
         size_t segment = 0;
-        AbstractState state = mergepoint[bb->id][segment];
+        const AbstractState& state = mergepoint[bb->id][segment];
```
mhyee, line 77: OK, this was a gotcha we missed...

rir/src/compiler/translations/pir_translator.h
```diff
@@ -10,7 +10,7 @@ namespace pir {
 class RirCompiler;
 class PirTranslator {
   public:
-    PirTranslator(std::string name) : name(name) {}
+    PirTranslator(const std::string& name) : name(name) {}
```
mhyee, line 13: Another gotcha. I guess `cppcheck` can actually be useful! &lt;/sarcasm&gt;

rir/src/interpreter/interp.cpp
```diff
@@ -376,6 +375,7 @@ void warnSpecial(SEXP callee, SEXP call) {
     // TODO: turn this on?
     return;
 
+#if 0
```
mhyee: Why is this disabled? Can we get a comment so that this doesn't confuse us in the future?

rir/src/interpreter/interp.cpp
```diff
@@ -1072,6 +1074,9 @@ static void cachedSetVar(SEXP val, SEXP env, Immediate idx, Context* ctx,
     UNPROTECT(1);
 }
 
+#pragma GCC diagnostic push
```
mhyee, line 1078: Will this cause problems for clang?

rir/src/interpreter/interp.cpp
```diff
@@ -990,20 +991,21 @@ RIR_INLINE void incPerfCount(Code* c) {
     }
 }
 
-static int debugging = 0;
 void debug(Code* c, Opcode* pc, const char* name, unsigned depth,
            Context* ctx) {
     return;
+#if 0
```
mhyee: Another place where I think we should have a comment.

rir/src/interpreter/interp.cpp
```diff
@@ -2622,9 +2627,11 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
 
             RCNTXT* cntxt = (RCNTXT*)RAW(val);
 
-            // (ab)use the same buffe to store the current pc
-            Opcode** oldPc = (Opcode**)(cntxt + 1);
-            *oldPc = pc;
+            {
+                // (ab)use the same buffe to store the current pc
```
mhyee: Might as well fix the typo here: buffe -> buffer

mhyee (commented):
I forget, but why aren't we running `cppcheck` as a pre-commit hook? It would be nice to see the issues and fix them locally, without waiting for Travis.

And if we're doing that, an explicit make/ninja task to run `cppcheck` (and `clang-format`) might also be nice.

.travis.yml
o-: no need to run cppcheck on more than one instance, so i just run it on one linux box

.travis.yml
o-: no output = good
otherwise all problems are listed separately

rir/src/compiler/util/builder.cpp
```diff
@@ -86,7 +86,7 @@ Builder::Builder(Closure* fun, Value* closureEnv)
         args[i] = this->operator()(new LdArg(i));
     auto mkenv = new MkEnv(closureEnv, fun->argNames, args.data());
     add(mkenv);
-    env = mkenv;
+    this->env = mkenv;
```
JanJecmen, line 89: and we were so pedantic to Guido about these previously ;]

o-:
> I forget, but why aren't we running cppcheck as a pre-commit hook? It would be nice to see the issues and fix them locally, without waiting for Travis.

in the past people where upset by pre commit hooks, that's why I stopped adding them. Also cppcheck is quite long-running. I think I found a solution (basically allowing you to skip the check while it's already running) that might work. But still, if you commit through some horrific git gui, then of course you will be stuck until it finishes. Anyway, next time somebody goes crazy because commiting takes 5 seconds I'll point out that you asked for the hook :P 

> And if we're doing that, an explicit make/ninja task to run cppcheck

`tools/cppcheck`

> (and clang-format) might also be nice.

`tools/git-clang-format`

we don't want to run clang format on everything. only on the commit diff. Because
1. not the whole codebase is clang format conform
2. I absolutely hate it when commits are full of completely unrelated whitespace changes

rir/src/ir/Compiler.cpp
```diff
@@ -391,10 +391,7 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
                         cs << BC::dup() << BC::setShared();
 
                         // Now load index and target
-                        cs << (superAssign ? BC::ldvarSuper(target)
-                                           : BC::ldvar(target));
-                        if (superAssign)
-                            cs << BC::setShared();
+                        cs << BC::ldvar(target);
```
JanJecmen, line 394: When we later remove the `if (superAssign) return false;` above, I bet this will be forgotten?

rir/src/ir/BC_inc.h
```diff
@@ -140,33 +140,16 @@ class BC {
...
-    }
-
-    BC operator=(BC other) {
-        bc = other.bc;
-        immediate = other.immediate;
-        return other;
-    }
+    BC& operator=(BC other) = delete;
+    BC(const BC& other) = delete;
+    BC(const BC&& other)
```
JanJecmen: 1/ shouldn't it be `BC(BC&& other)` for move ctor?
2/ move assignment is missing

rir/src/ir/BC_inc.h
```diff
@@ -140,33 +140,16 @@ class BC {
...
+        : bc(other.bc), immediate(other.immediate),
+          immediateCallArguments(std::move(other.immediateCallArguments)),
```
JanJecmen: `std::swap`?

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -19,41 +21,112 @@ class BC;
 
 namespace pir {
 
-class StreamLogger {
-    static uint64_t logId;
+class StreamLogger;
+
+class LogStream {
+  private:
+    LogStream(const LogStream&) = delete;
```
JanJecmen: deleted functions public?

JanJecmen (approved):

rir/src/compiler/util/builder.cpp
o-: it's one of the 2 false positives that cppcheck produced

rir/src/ir/Compiler.cpp
o-: well, it will break fast :)

rir/src/ir/BC_inc.h
o-: hmm, good point. I wonder why it even works

rir/src/ir/BC_inc.h
o-: I don't think that's what we want here

JanJecmen: I was thinking just for the two vectors. Why wouldn't it work? (just curious, `move` is fine too)

rir/src/ir/BC_inc.h
```diff
@@ -141,12 +141,20 @@ class BC {
 
     BC() : bc(Opcode::invalid_) {}
 
-    BC& operator=(BC other) = delete;
     BC(const BC& other) = delete;
-    BC(const BC&& other)
+    BC& operator=(BC other) = delete;
```
JanJecmen: oh and this one doesn't trigger the cppcheck? `const BC& other`

o-:
In general, the cppcheck is not extremely impressive, but it did find some gotchas, that would have been annoying to debug and it also detected the uninitialized type feedback array size. So I guess overall it might be worth it...

rir/src/interpreter/interp.cpp
o-: apparently not. I think they support most GCC pragmas that people use....

.travis.yml
mhyee: My worry is that no output could also mean "nothing happened," e.g. if we misconfigure and `$CPPCHECK` is not properly set.

mhyee:
> Also cppcheck is quite long-running. I think I found a solution (basically allowing you to skip the check while it's already running) that might work. But still, if you commit through some horrific git gui, then of course you will be stuck until it finishes. Anyway, next time somebody goes crazy because commiting takes 5 seconds I'll point out that you asked for the hook :P

Maybe I'll be the one so annoyed that I'll delete the hook! In all seriousness, I like catching errors/warnings as early as possible, but there's a reason we don't run the full test suite. So let's wait and see how annoying it actually is.

> In general, the cppcheck is not extremely impressive, but it did find some gotchas, that would have been annoying to debug and it also detected the uninitialized type feedback array size. So I guess overall it might be worth it...

It would also have caught one of the uninitialized members I had in my recent PR, and saved some debugging time.

And if we want more firepower, there's always [Coverity](https://docs.travis-ci.com/user/coverity-scan/)...

o-:
> Maybe I'll be the one so annoyed that I'll delete the hook! In all seriousness, ...

already happened to me :) I moved it to pre-push

> And if we want more firepower, there's always Coverity...

oh, I didn't know that one. nice

o-:
haha, there is already quite a discrepancy between cppcheck on travis than on my machine...

PR #221, opened 2018-09-14
improve PIR logger output

o-:
* Better names
* Always print which is the main function being compiled

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -173,9 +173,11 @@ SEXP pirCompile(SEXP what, const std::string& name, pir::DebugOptions debug) {
     // compile to pir
     pir::Module* m = new pir::Module;
     pir::StreamLogger logger(debug);
+    logger.title("Compiling " + name);
     pir::Rir2PirCompiler cmp(m, logger);
     cmp.compileClosure(what, name,
                        [&](pir::Closure* c) {
+                           logger.flush();
                            cmp.optimizeModule(preserveVersions);
 
                            // compile back to rir
@@ -210,35 +212,30 @@ REXPORT SEXP pir_tests() {
     return R_NilValue;
 }
 
-// startup ---------------------------------------------------------------------
-
-SEXP pirOpt(SEXP fun) {
+SEXP pirOptDefaultOpts(SEXP closure, SEXP name) {
+    std::string n = "";
+    if (TYPEOF(name) == SYMSXP)
+        n = CHAR(PRINTNAME(name));
     // PIR can only optimize closures, not expressions
-    if (isValidClosureSEXP(fun) && DispatchTable::check(BODY(fun)))
-        return pirCompile(fun, "", PirDebug);
+    if (isValidClosureSEXP(closure))
+        return pirCompile(closure, n, PirDebug);
     else
-        return fun;
+        return closure;
+}
+
+SEXP pirOptDefaultOptsDryrun(SEXP closure, SEXP name) {
+    std::string n = "";
+    if (TYPEOF(name) == SYMSXP)
+        n = CHAR(PRINTNAME(name));
+    // PIR can only optimize closures, not expressions
+    if (isValidClosureSEXP(closure))
+        return pirCompile(closure, n, PirDebug | pir::DebugFlag::DryRun);
+    else
+        return closure;
 }
 
 bool startup() {
-    auto pir = getenv("PIR_ENABLE");
-    if (pir && std::string(pir).compare("off") == 0) {
-        initializeRuntime(rir_compile, [](SEXP f) { return f; });
-    } else if (pir && std::string(pir).compare("force") == 0) {
-        initializeRuntime(
-            [](SEXP f, SEXP env) { return pirOpt(rir_compile(f, env)); },
-            [](SEXP f) { return f; });
-    } else if (pir && std::string(pir).compare("force_dryrun") == 0) {
-        initializeRuntime(
-            [](SEXP f, SEXP env) {
-                return pirCompile(rir_compile(f, env), "",
-                                  PirDebug | pir::DebugFlag::DryRun);
-            },
-            [](SEXP f) { return f; });
-    } else {
-        // default on
-        initializeRuntime(rir_compile, pirOpt);
-    }
+    initializeRuntime();
     return true;
 }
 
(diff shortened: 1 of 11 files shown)


rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -198,5 +193,29 @@ void LogStream::section(const std::string& title) {
...
 
+void StreamLogger::beginPhase(const std::string& msg) {
+    if (!options.includes(DebugFlag::PrintIntoFiles) &&
+        (options.intersects(PrintDebugPasses) ||
+         options.includes(DebugFlag::ShowWarnings))) {
+        std::cout << BLUE;
+        int l = 36 - (int)msg.length() / 2;
+        int r = l - msg.length() % 2;
+        std::cout << "\n╞";
+        for (int i = 0; i < l; ++i)
```
JanJecmen, line 204: `cout << setw(l) << "=";`

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -198,5 +193,29 @@ void LogStream::section(const std::string& title) {
...
+            std::cout << "═";
+        std::cout << "╡  " << CLEAR << msg << BLUE << "  ╞";
+        for (int i = 0; i < r; ++i)
```
JanJecmen, line 207: same

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -290,8 +294,10 @@ bool Rir2Pir::compileBC(
         Promise* prom = insert.function->createProm(promiseCode->src);
         {
             Builder promiseBuilder(insert.function, prom);
-            if (tryCompilePromise(promiseCode, promiseBuilder))
+            if (!tryCompilePromise(promiseCode, promiseBuilder)) {
```
JanJecmen, line 297: Was this broken? W/O the `!`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -552,6 +560,7 @@ bool Rir2Pir::compileBC(
     case Opcode::movloc_:
     case Opcode::isobj_:
     case Opcode::check_missing_:
+        log.unsupportedBC("Unsupported BC (are you recompiling?)", bc);
         assert(false && "Recompiling PIR not supported for now.");
```
JanJecmen, line 564: Replace the asserts with just signalling failure?

JanJecmen (approved):
lgtm

rir/src/api.cpp
```diff
@@ -210,35 +212,30 @@ REXPORT SEXP pir_tests() {
     return R_NilValue;
 }
 
-// startup ---------------------------------------------------------------------
+SEXP pirOptDefaultOpts(SEXP closure, SEXP name) {
+    std::string n = "";
+    if (TYPEOF(name) == SYMSXP)
+        n = CHAR(PRINTNAME(name));
+    // PIR can only optimize closures, not expressions
+    if (isValidClosureSEXP(closure) && DispatchTable::check(BODY(closure)))
```
charig: @mhyee pointed me in a previous PR that `DispatchTable::check` already test if it is a valid closure...

rir/src/compiler/debugging/stream_logger.cpp
```diff
@@ -198,5 +193,29 @@ void LogStream::section(const std::string& title) {
     highlightOff();
 }
 
+void StreamLogger::beginPhase(const std::string& msg) {
+    if (!options.includes(DebugFlag::PrintIntoFiles) &&
```
charig: Why the phase is not included if we print into files?

rir/src/api.cpp
```diff
@@ -173,9 +173,11 @@ SEXP pirCompile(SEXP what, const std::string& name, pir::DebugOptions debug) {
     // compile to pir
     pir::Module* m = new pir::Module;
     pir::StreamLogger logger(debug);
+    logger.beginPhase("Compiling " + name);
     pir::Rir2PirCompiler cmp(m, logger);
     cmp.compileClosure(what, name,
                        [&](pir::Closure* c) {
+                           logger.endPhase();
```
charig: I do not get exactly what is the idea of a phase. Why the phase ends just before pir starts to optimize? Why the optimization is not another phase?

rir/src/compiler/debugging/stream_logger.cpp
o-: which file? There is one file per function. But the phase is a global thing...

rir/src/api.cpp
o-: this is just `flush()` that I renamed to phase. The idea is to flush all streams after rir2pir, after optimizations and after pir2rir, so you get some output, if it crashes in one of the steps.

should I name it back to flush? I think that was probably the better name

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: yes

rir/src/compiler/debugging/stream_logger.cpp
o-: ehm, that does not work.
It's `cout << std::setfill('=') << std::setw(l) << "" << std::setfill(' ') << "\n";`
honestly I am not convinced that this is nicer, plus it does not work in the case of the box drawing characters, since `setfill` only accepts an 8bit char.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: I think we still want this to fail, because how the compiler is invoked, it should actually not happen currently that we recompile pir optimized code

rir/src/api.cpp
o-: it doesn't

o-: the other way round, isValidClosureSEXP checks if it's a dispatch table in the closure

o-:
gonna merge this, since the current logger does not switch back the console color from red, which is kinda annoying ;)

PR #223, opened 2018-09-14
use simple ret_ if there is no way of leaking the return statement

o-:


o-:
that does not seem to work. I need to investigate. My assumption was that `return` as a top-level statement in a function doesn't do anything funky, but apparently there is something going on.

JanJecmen:
I think your idea is fine, it's just a missed case in the verifier:
(btw should `deopt_` also be covered by `isReturn()`?)

```diff
diff --git a/rir/src/ir/CodeVerifier.cpp b/rir/src/ir/CodeVerifier.cpp
index 734c01a..d956555 100644
--- a/rir/src/ir/CodeVerifier.cpp
+++ b/rir/src/ir/CodeVerifier.cpp
@@ -46,7 +46,7 @@ class State {

     void advance() {
         BC bc = BC::advance(&pc);
-        if (bc.bc == Opcode::return_ || bc.bc == Opcode::deopt_)
+        if (bc.isReturn() || bc.bc == Opcode::deopt_)
             ostack = 0;
         else
             ostack -= bc.popCount();
@@ -201,8 +201,7 @@ void CodeVerifier::calculateAndVerifyStack(Code* code) {
             BC cur = BC::decode(pc);
             i.advance();
             max.updateMax(i);
-            if (cur.bc == Opcode::ret_ || cur.bc == Opcode::return_ ||
-                cur.bc == Opcode::deopt_) {
+            if (cur.isReturn() || cur.bc == Opcode::deopt_) {
                 i.checkClear();
                 break;
             } else if (cur.bc == Opcode::br_) {
```

o-:
@JanJecmen thanks for looking at it. That is indeed one of the problems. The next problem is, that I do not empty the stack before `ret_`. But there is a problem: in `ir/Compiler.cpp` we do not actually keep track of how high the stack is, from what I can see :(

JanJecmen:
Oh right, because until now we didn't care, the `return_` gets the correct stack pointer from the target context...

o-:
let's try to just do this in pir. keeping track of stack height in rir would be a bit of a painful change (though maybe eventually we still want it..)

o-:
(I also decided to remove some callback confusion in the code...)

o-:
hmm, also actually the verifier should verify that the stack is empty on ret_, so I have to revert that again

rir/src/ir/BC_inc.h
```diff
@@ -229,7 +229,8 @@ class BC {
     bool isPure() { return isPure(bc); }
 
     bool isReturn() const {
-        return bc == Opcode::ret_ || bc == Opcode::return_;
+        return bc == Opcode::ret_ || bc == Opcode::return_ ||
+               bc == Opcode::deopt_;
```
JanJecmen: Maybe should now be named `isExit` or similar

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
```diff
@@ -18,50 +18,27 @@ class Rir2Pir {
...
-                               [&](Value* v) {
-                                   finalize(v, insert);
-                                   return true;
-                               },
-                               []() { return false; });
+        if (auto res = tryTranslate(srcCode, insert)) {
+            finalize(res, insert);
+            return true;
+        }
+        return false;
```
JanJecmen: Maybe move to `.cpp` to be consistent

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
```diff
@@ -75,6 +52,17 @@ class Rir2Pir {
     bool compileBC(const BC& bc, Opcode* pos, rir::Code* srcCode, RirStack&,
                    Builder&, std::unordered_map<Value*, CallFeedback>&,
                    std::unordered_map<Value*, TypeFeedback>&) const;
+    virtual bool inPromise() const { return false; }
```
JanJecmen, line 48: Wasn't sure why this is private, I like this answer: https://stackoverflow.com/a/2170696
(originally I thought you cannot override private because you don't see it)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
```diff
@@ -18,50 +18,27 @@ class Rir2Pir {
...
     bool tryCompile(rir::Code* srcCode, Builder& insert)
         __attribute__((warn_unused_result)) {
-        return translate<bool>(srcCode, insert,
-                               [&](Value* v) {
-                                   finalize(v, insert);
-                                   return true;
-                               },
-                               []() { return false; });
+        if (auto res = tryTranslate(srcCode, insert)) {
+            finalize(res, insert);
```
JanJecmen: Can `finalize` fail? Ie. if the `InsertCast` pass fails, we just abort.. I guess in general I am not sure when an abort is appropriate and when just signalling failure is fine

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -696,9 +698,12 @@ void Rir2Pir::translate(rir::Code* srcCode, Builder& insert,
...
-                log.warn("Cannot compile Function. Unsupported return bc");
-                fail();
-                return;
+                if (inPromise()) {
+                    log.warn("Cannot compile Function. Unsupported return bc "
+                             "in promise");
+                    return nullptr;
+                }
+                break;
             default:
```
JanJecmen: Is aborting on `deopt_` what we want?

JanJecmen (commented):
Looks good, I think removing the callbacks makes it a bit easier to read.
But, I don't see the actual optimization in Compiler.cpp now ;)

o-:
>  But, I don't see the actual optimization in Compiler.cpp now ;)

It's a bit hidden, but basically this fallthrough case here:
https://github.com/reactorlabs/rir/pull/223/files#diff-a3b8fcdd57c845ed78e9242b24c2ca2eR705

we will not abort compilation on `return_` if we are outside of the promise, and we will then return tos later in the function. I will make sure to make it more explicit!

rir/src/compiler/translations/rir_2_pir/rir_2_pir.h
o-: Finalize should not fail.

tryCompile fails = unsupported feature
finalize fails = bug

for example the compiler should always generate code such that insertcast succeeds. We do insert cast lazily, because sometimes it's hard to do it eager, as you do not know all the usages of a value when you create it

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: good catch

JanJecmen:
Ah, sneaky

PR #224, opened 2018-09-17
why did we do this?

o-:
I see no reason, plus it's horrible for performance


o-:
I guess it looks fine :)

JanJecmen:
This is why, I guess: https://github.com/wch/r-source/blob/tags/R-3-5-1/src/main/eval.c#L651
Are we sure it's safe?

PR #225, opened 2018-09-17
Named arguments for PIR Calls

o-:
Introduce a new NamedCall PIR instruction that supports named
arguments. Currently this named call instruction is a completely
separate instruction. Therefore for existing optimizations and
analysis, this new instruction is completely opaque.

What we gain from this commit is that named calls can be compiled
to pir and back to rir.

### DIFF
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -93,9 +93,9 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
                 envs.merge(nextFun.result());
                 handled = true;
             }
-        } else {
-            assert(CallBuiltin::Cast(i) || CallSafeBuiltin::Cast(i));
         }
+        // Not handled:
+        // CallBuiltin, CallSafeBuiltin, CallImplicit, CallNamed
     }
 
     // Keep track of closures
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -93,9 +93,12 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
                 envs.merge(nextFun.result());
                 handled = true;
             }
+        } else {
+            // TODO: support for NamedCall
+            assert((CallBuiltin::Cast(i) || CallSafeBuiltin::Cast(i) ||
+                    NamedCall::Cast(i)) &&
+                   "New call instruction not handled?");
         }
-        // Not handled:
-        // CallBuiltin, CallSafeBuiltin, CallImplicit, CallNamed
     }
 
     // Keep track of closures
--- a/rir/src/compiler/pir/instruction.cpp
+++ b/rir/src/compiler/pir/instruction.cpp
@@ -3,6 +3,7 @@
 
 #include "../util/visitor.h"
 #include "R/Funtab.h"
+#include "utils/Pool.h"
 #include "utils/capture_out.h"
 
 #include <algorithm>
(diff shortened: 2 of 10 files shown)


o-:
sorry for mixing the two things. But I also implemented `CallImplicit`, so we can generate `call_implicit_` bytecodes, which are faster than the `mkarg, call` combo. Unfortunately I intermingeled the two PR's with some call printing refactoring and can't easily separate. So I am just gonna add it here.

rir/src/ir/CodeVerifier.cpp
```diff
@@ -330,7 +330,7 @@ void CodeVerifier::verifyFunctionLayout(SEXP sexp, ::Context* ctx) {
                 }
             }
             if (*cptr == Opcode::named_call_) {
-                uint32_t nargs = *reinterpret_cast<Immediate*>(cptr + 5);
+                uint32_t nargs = *reinterpret_cast<Immediate*>(cptr + 1);
```
mhyee, line 333: Whoa, was this a bug before? I guess we couldn't really test it until now...

rir/src/ir/BC.h
```diff
@@ -287,7 +287,7 @@ BC BC::call(size_t nargs, const std::vector<SEXP>& names, SEXP ast) {
     std::vector<PoolIdx> nameIdxs;
     for (auto n : names)
         nameIdxs.push_back(Pool::insert(n));
-    return BC(Opcode::named_call_, im, {}, nameIdxs);
+    return BC(Opcode::named_call_, im, {}, std::move(nameIdxs));
```
mhyee, line 290: Also update line 275 above?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -356,11 +375,59 @@ CallInstruction* CallInstruction::CastCall(Value* v) {
...
+void Call::printArgs(std::ostream& out) {
+    printCallArgs(out, this, callerEnv());
+}
+
+void NamedCall::printArgs(std::ostream& out) {
+    size_t nargs = nCallArgs();
+    out << "(";
+    for (size_t i = 0; i < nargs; ++i) {
+        if (names.size() > i && names.at(i) != R_NilValue)
+            out << CHAR(PRINTNAME(names.at(i))) << "=";
```
mhyee: Spaces before and after `=`?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -93,9 +93,9 @@ void TheScopeAnalysis::apply(AS& envs, Instruction* i) const {
                 envs.merge(nextFun.result());
                 handled = true;
             }
-        } else {
-            assert(CallBuiltin::Cast(i) || CallSafeBuiltin::Cast(i));
         }
+        // Not handled:
+        // CallBuiltin, CallSafeBuiltin, CallImplicit, CallNamed
```
mhyee: Why was the assert changed to a comment?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -356,11 +375,59 @@ CallInstruction* CallInstruction::CastCall(Value* v) {
...
+
+void Call::printArgs(std::ostream& out) {
+    printCallArgs(out, this, callerEnv());
+}
+
+void NamedCall::printArgs(std::ostream& out) {
+    size_t nargs = nCallArgs();
+    out << "(";
+    for (size_t i = 0; i < nargs; ++i) {
+        if (names.size() > i && names.at(i) != R_NilValue)
```
mhyee: `i < names.size() && names[i]`. Don't really need `names.at(i)` to check bounds when you're already checking bounds. (And the flipped comparison is a bit unconventional.)

Probably the other `names.at(i)` can be changed to `names[i]`.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -1063,17 +1083,51 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
                 auto newDeopt = new ScheduledDeopt();
                 newDeopt->consumeSafepoints(deopt);
                 auto newDeoptPos = bb->insert(it, newDeopt);
                 next = newDeoptPos + 2;
+            } else if (auto call = Call::Cast(*it)) {
+                // Lower calls to call implicit
+                std::vector<Promise*> args;
+                if (allLazy(call, args))
+                    call->replaceUsesAndSwapWith(
+                        new CallImplicit(call->callerEnv(), call->cls(), args,
```
mhyee: Should this be `std::move(args)`? Same comment for below.

mhyee:
That should have been an "Approve PR" (but here are some comments).

mhyee (approved):
The previous review should have been an approval.

rir/src/ir/CodeVerifier.cpp
o-: yeah, we didn't really use that instruction

PR #227, opened 2018-09-18
Colors

o-:
This PR is only for the last commit. It is based on #225.

cleanup instruction printing and moar colors

mark environment with colors
* yellow: read
* red: write
* magenta: leak

and make sure to only write colors on actual ttys

### DIFF
--- a/rir/src/compiler/pir/instruction.cpp
+++ b/rir/src/compiler/pir/instruction.cpp
@@ -4,6 +4,7 @@
 #include "../util/visitor.h"
 #include "R/Funtab.h"
 #include "utils/Pool.h"
+#include "utils/Terminal.h"
 #include "utils/capture_out.h"
 
 #include <algorithm>
@@ -45,35 +46,57 @@ void printPaddedInstructionName(std::ostream& out, const std::string& name) {
     out << std::left << std::setw(maxInstructionNameLength + 1) << name << " ";
 }
 
-void Instruction::printArgs(std::ostream& out) {
-    if (nargs() > 0) {
-        for (size_t i = 0; i < nargs(); ++i) {
-            arg(i).val()->printRef(out);
-            if (i + 1 < nargs())
-                out << ", ";
-        }
-    }
-}
-
-void Instruction::print(std::ostream& out) {
+void printPaddedTypeAndRef(std::ostream& out, Instruction* i) {
     std::ostringstream buf;
-    buf << type;
+    buf << i->type;
     out << std::left << std::setw(7) << buf.str() << " ";
     buf.str("");
-    if (type != PirType::voyd()) {
-        printRef(buf);
+    if (i->type != PirType::voyd()) {
+        i->printRef(buf);
         out << std::setw(5) << buf.str() << " = ";
-        buf.str("");
     } else {
         out << "        ";
     }
+}
 
+void Instruction::printArgs(std::ostream& out, bool tty) {
+    size_t n = nargs();
+    size_t env = hasEnv() ? envSlot() : n + 1;
+
+    for (size_t i = 0; i < n; ++i) {
+        if (i != env) {
+            arg(i).val()->printRef(out);
+            if (i + 1 < n && (i + 1) != env)
+                out << ", ";
+        }
+    }
+    if (hasEnv())
+        out << ", ";
+}
+
+void Instruction::print(std::ostream& out, bool tty) {
+    printPaddedTypeAndRef(out, this);
     printPaddedInstructionName(out, name());
-    printArgs(buf);
-    out << std::setw(50) << buf.str();
+    printArgs(out, tty);
+    printEnv(out, tty);
+}
 
-    if (leaksEnv())
-        out << " ; env leak";
+void Instruction::printEnv(std::ostream& out, bool tty) {
+    if (hasEnv()) {
+        if (tty) {
+            if (leaksEnv())
+                ConsoleColor::magenta(out);
+            else if (changesEnv())
+                ConsoleColor::red(out);
+            else
+                ConsoleColor::yellow(out);
+        }
+
+        env()->printRef(out);
+
+        if (tty)
+            ConsoleColor::clear(out);
+    }
 }
 
 void Instruction::printRef(std::ostream& out) {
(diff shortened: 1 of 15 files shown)


o-:
btw, if you ask yourself, why not just check if `out` is a tty, for example directly in `instruction::print`: the problem is that we have the buffered LogStream, that passes a `strstream` as `out` and only later dumps it into stdout.

rir/src/compiler/pir/instruction.cpp
```diff
@@ -45,35 +46,55 @@ void printPaddedInstructionName(std::ostream& out, const std::string& name) {
...
-        out << " ; env leak";
+void Instruction::printEnv(std::ostream& out, bool tty) {
+    if (hasEnv()) {
+        if (tty && leaksEnv())
+            ConsoleColor::magenta(out);
+        else if (changesEnv())
+            ConsoleColor::red(out);
+        else
+            ConsoleColor::yellow(out);
+
```
JanJecmen: The second and third conditions fire even if `!tty`
Also, don't we still want the textual version for non-terminal output?

JanJecmen: And a nitpick, isn't leaking worse than writing? So leak could be red (danger!) and write only magenta (ie. not so bad)

rir/src/compiler/pir/instruction.cpp
```diff
@@ -155,60 +176,55 @@ void LdConst::printArgs(std::ostream& out) {
...
     out << ", ";
-    env()->printRef(out);
 }
 
-void LdVarSuper::printArgs(std::ostream& out) {
-    out << CHAR(PRINTNAME(varName)) << ", ";
-    env()->printRef(out);
+void LdVarSuper::printArgs(std::ostream& out, bool tty) {
+    out << CHAR(PRINTNAME(varName));
+    out << ", ";
```
JanJecmen, line 216: the commas could be written in `printEnv` so you don't have to do it by hand in all instructions that have env

rir/src/compiler/pir/instruction.cpp
```diff
@@ -231,16 +247,10 @@ void Phi::printArgs(std::ostream& out) {
...
+void PirCopy::print(std::ostream& out, bool tty) {
+    printPaddedTypeAndRef(out, this);
     std::ostringstream buf;
-    buf << type;
-    out << std::left << std::setw(7) << buf.str() << " ";
-    buf.str("");
-    printRef(buf);
-    out << std::setw(5) << buf.str() << " = ";
-    buf.str("");
     arg(0).val()->printRef(buf);
```
JanJecmen, line 255: `buf` is never written to `out`

JanJecmen (approved):
How about passing, instead of the naked ostreams and bools, the logstreams? And then have the `operator<<` on logstream check if tty or not. But perhaps it's too much rewriting for a too low gain...

o-:
thanks

> How about passing, instead of the naked ostreams and bools, the logstreams?

the problem is, that then you cannot print to stdout anymore without a logstream. I think it's kind of a bad idea to couple the printing of instructions with the logging infrastructure. What if you want to print in the debugger?

That's why I looked into having logstream inherit from std::ostream, but it so nasty that I did not even try,

rir/src/compiler/pir/instruction.cpp
o-: I want to be able to use printEnv in other places. For example if you look at MkEnv it does:
`out << "env="; Instruction::printEnv();`

rir/src/compiler/pir/instruction.cpp
o-: yes, but at least in my color schema magenta looks more scary. not to you?

JanJecmen:
Fair point. Looks good then

mhyee:
Sorry for being difficult, but having run this, I'm not sure I like *only* colouring the environments. I don't immediately remember what the colours mean, and if you redirect output, then the colours are lost.

And this is unrelated, but am I the only one getting filenames with question marks when I run with `PrintIntoFiles`?

o-:
> Sorry for being difficult, 

no problem

> but having run this, I'm not sure I like only colouring the environments. I don't immediately remember what the colours mean, 

I don't know. It's what I am currently interested in. But now we  have the infrastructure, so we can do other things, eg. bold instructions or I don't know. suggestions?

> and if you redirect output, then the colours are lost.

yeah, good point.  are you suggesting a textual alternative? I think as long as colors are just adding and there is no information that is just in the colors it should be fine.

> And this is unrelated, but am I the only one getting filenames with question marks when I run with PrintIntoFiles?

question mark means we where not able to find the name. That still happens quite often, depending how you invoke the compiler and how the function was called.

Or are you saying you get less function names in files than in stdout? that i think would be a bug. do you have an example?

mhyee:
I'd like to see some text, so that the information is available even without colours. I think the previous text was fine.

I don't mind the colours either way. Actually, they're probably good, to make things stand out more.

> question mark means we where not able to find the name. That still happens quite often, depending how you invoke the compiler and how the function was called.
>
> Or are you saying you get less function names in files than in stdout? that i think would be a bug. do you have an example?

I wasn't being clear. My problem is generating files with weird symbols in the filenames, like `?`, `>`, and `@`:

```
pir-function-00001-?_0x33e9538.log
pir-function-00007-execute>ball@576_0x3d58248.log
```

I guess it's not *actually* causing problems, but it's a bit annoying.

o-:
> I'd like to see some text, so that the information is available even without colours. I think the previous text was fine.

ok, but which instructions leak is currently static, so you know by the kind of instruction. so technically speaking it's redundant

mhyee:
> ok, but which instructions leak is currently static, so you know by the kind of instruction. so technically speaking it's redundant

Redundancy is good. Redundancy is good.

Seriously though, I think the benefits of having an additional visual aid/reminder (so I don't have to go well, this is a force, and let's see, oh yeah, force leaks the environment) outweigh the costs of including some extra text. And colour doesn't count as an additional visual aid because it still needs a legend (what was magenta again, oh yeah, it's a leak), and that's only if we don't print into files.

PR #228, opened 2018-09-19
check gnur testsuite return code and actually fail if they don't pass + fix all the errors this exposed

o-:


o-:
The gnur testrunner only returns non-zero result when running with TEST_MC_CORES > 1. That's why the package tests failed in the past with TEST_MC_CORES > 1, since they where actually always failing :)

To be sure we always fail when we should (even when the cores are set to 1) this PR adds an additional script that greps the test logs for ERROR. This script has the additional advantage that it prints the full test error in the log, where it is not mixed with output from other concurrently running tests.

A number of fixes where neccessary after "enabling" those tests for real:

* install all the latex crap on travis
* start a virtual framebuffer for the tests
* creating a R context before calling dispatch in the subassign bytecodes
* preventing the closures mentioned in the typefeedback from being collected. For this i added a generic "extra pool" to the code objects. This is a list that you can add stuff that needs to have the same lifetime as the code object.
* #234 
* fixed handling of missing arguments passed to ldarg
* standardGeneric cannot be inlined (like usemethod)

Since now all the issues are resolved, this PR scales TEST_MC_CORES to the number of available CPUs, to make the tests run faster. Overall it unfortunately now takes much longer, since installing all the latex crap and running the previously failing latex tests takes so much time.

This PR also makes the promise in the `MkArg` instruction private. This turned out to be a red herring, and there was nothing wrong with the updating of this field. I still left the changes in the PR, since it might prevent future errors.

I also removed a redundant entry in the travis build matrix. the tests are taking already long enough...

tools/gnur-make-tests
```diff
@@ -31,6 +31,7 @@ if [[ "$DISPLAY" == "" && "$XVFB" != "" ]]; then
 fi
 
 CORES=`nproc 2> /dev/null || echo 1`
+let CORES=$CORES*2
```
mhyee, line 34: Why is the number of cores being doubled?

rir/src/runtime/Code.h
```diff
@@ -49,14 +49,15 @@ struct Code : public RirRuntimeObject<Code, CODE_MAGIC> {
...
+    static constexpr size_t NumLocals = 2;
 
     Code() = delete;
 
     Code(FunctionSEXP fun, size_t index, SEXP ast, unsigned codeSize,
          unsigned sourceSize, bool isDefaultArg, size_t localsCnt);
 
   private:
-    FunctionSEXP function_;
+    SEXP locals_[NumLocals];
```
mhyee, line 66: I think it would help to have a comment here, about how this is the "extra pool" of objects we want the Code object to keep alive.

And to be more specific, it looks like `locals_[0]` is the associated function object, and `locals_[1]` is a linked list of all the extra stuff?

rir/src/ir/RuntimeFeedback.h
```diff
@@ -24,20 +26,10 @@ struct CallFeedback {
...
-        if (numTargets < MaxTargets) {
-            int i = 0;
-            for (; i < numTargets; ++i)
-                if (targets[i] == callee)
-                    break;
-            if (i == numTargets)
-                targets[numTargets++] = callee;
-        }
-    }
+    std::array<SEXP, MaxTargets> targets;
```
mhyee, line 33: Make this private? If you're enforcing that accesses are only through `getTarget`.

mhyee (approved):
Thanks! The summary comment was really helpful!

In fact, if you do a rebase, could you add that comment to one of the commits, so that it's preserved in the history?

tools/gnur-make-tests
o-: well, turns out that there is quite some infrastructure involved in those package tests. like packaging the package, shelling out, compiling latex, writing output to disk, diffing outputs and so on. So the cores are not really saturated at all if we have just one process per core.

rir/src/ir/RuntimeFeedback.h
o-: yeah, problem is then the compiler complains about memcopying the struct with private fields. which we really want to....

PR #229, opened 2018-09-20
fix broken call instruction printing

o-:


o-:
@JanJecmen I think at least one of them you already caught, but I failed to fix it...

just some small fixes to the printing, merging tbr

PR #230, opened 2018-09-20
don't convert named calls to static calls

o-:
Before we can convert named_call_implicit to static call, we need the
static version of arg position matching.

This case was introduced by accident


o-:
tbr

PR #231, opened 2018-09-20
don't deopt on constants

o-:


rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -437,11 +449,8 @@ bool Rir2Pir::compileBC(
         if (typeFeedback[rhs].numTypes > 0 &&                                  \
             typeFeedback[lhs].numTypes > 0 &&                                  \
             !typeFeedback[rhs].observedObject() &&                             \
-            !typeFeedback[lhs].observedObject()) {                             \
-            Value* leftIsObj = insert(new IsObject(lhs));                      \
-            insert.conditionalDeopt(leftIsObj, srcCode, pos, stack, false);    \
-            Value* rightIsObj = insert(new IsObject(rhs));                     \
-            insert.conditionalDeopt(rightIsObj, srcCode, pos, stack, false);   \
+            !typeFeedback[lhs].observedObject() && deoptIfObject(lhs) &&       \
+            deoptIfObject(rhs)) {                                              \
```
JanJecmen: What about the mixed case? When `lhs` gives true and `rhs` false? Won't we have an unnecessary deopt then?

o-: I know. but ok, gonna write it differently, you are right

o-:
@JanJecmen ok, this is now ready. I rebased on top of #228 since they conflict anyway
(so you only need to check the last commit here)

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -125,6 +125,51 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
     return mergepoints;
 }
 
+static bool valueMaybeObject(Value* v) {
+    if (auto c = LdConst::Cast(v)) {
+        return OBJECT(c->c);
```
JanJecmen: Maybe more readable `return Rf_isObject(c->c)`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -125,6 +125,51 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
...
+    }
+    return true;
+}
+
+static bool valueSeenObject(
+    Value* v,
+    const std::unordered_map<Value*, rir::TypeFeedback>& typeFeedback) {
+    if (auto c = LdConst::Cast(v)) {
+        return OBJECT(c->c);
```
JanJecmen: Here too

rir/src/ir/RuntimeFeedback.h
```diff
@@ -69,7 +69,9 @@ struct TypeFeedback {
         }
     }
 
-    bool observedObject() {
+    bool empty() const { return numTypes > 0; }
```
JanJecmen: 1/ don't see this called anywhere
2/ shouldn't it be `numTypes == 0`?

JanJecmen (approved):
Sorry, I missed this one.
I like the templated function way more than the original macro!

o-:
ok, I am gonna redo this on top of #241 I think

PR #232, opened 2018-09-20
Fix missing case for force_dryrun.

mhyee:
It's just a few lines after the end of the diff, but `c->closureOptimizer` was not initialized in the `force_dryrun` case.


o-:
good catch

PR #233, opened 2018-09-21
Miscellaneous printing fixes

mhyee:
Print to ostream parameter, not std::cout.

Also other miscellaneous fixes (see second commit).

---

Before, we would get something like:

```
╞══════════════════════════╡  Compiling bitwShiftL  ╞══════════════════════════╡
bitwiseShiftL
┌──────────────────────────────────────────────────────────────────────────────┐
│ bitwShiftL_0x2a6ecc8                                                         │
├────── Final PIR Version
Closure 0x2792ad0(0x2a6ecc8)
BB0
  val^    %0.0  = LdArg            0
  val^    %0.1  = LdArg            1
  env     e0.2  = MkEnv            a=%0.0, n=%0.1, parent=<environment: namespace:base>
  val     %0.3  = Force            %0.0, e0.2
  val^?   %0.4  = LdVar            n, e0.2
  val     %0.5  = Force            %0.4, e0.2
  val^    %0.6  = CallBuiltin      (%0.3, %0.5) e0.2
  val     %0.7  = Force            %0.6, e0.2
  void            Return           %0.7
```

Note the stray `bitwiseShiftL` at the top, and how `CallBuiltin` is calling some unnamed function.

The correct output is:

```
╞══════════════════════════╡  Compiling bitwShiftL  ╞══════════════════════════╡
┌──────────────────────────────────────────────────────────────────────────────┐
│ bitwShiftL_0x2a6ecc8                                                         │
├────── Final PIR Version
Closure 0x2792ad0(0x2a6ecc8)
BB0
  val^    %0.0  = LdArg            0
  val^    %0.1  = LdArg            1
  env     e0.2  = MkEnv            a=%0.0, n=%0.1, parent=<environment: namespace:base>
  val     %0.3  = Force            %0.0, e0.2
  val^?   %0.4  = LdVar            n, e0.2
  val     %0.5  = Force            %0.4, e0.2
  val^    %0.6  = CallBuiltin      bitwiseShiftL(%0.3, %0.5) e0.2
  val     %0.7  = Force            %0.6, e0.2
  void            Return           %0.7
```

### DIFF
--- a/rir/src/interpreter/runtime.cpp
+++ b/rir/src/interpreter/runtime.cpp
@@ -2,7 +2,7 @@
 #include "api.h"
 #include "interp.h"
 
-#include <sstream>
+#include <iomanip>
 
 SEXP envSymbol;
 SEXP callSymbol;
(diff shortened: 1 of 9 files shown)


JanJecmen (commented):
Good one.
I did a search for `std::cout` and there are a couple of places that are dubious, but maybe they can go into a separate pr...

- runtime.cpp: `printFunction` creates `stringstream` and then dumps it into `cout`, could print directly into `cout`.. Also, it mixes `Rprintf` and `cout`
- type.cpp: the pir type prints always into `cout` (it looks as if it's not used at all though)
- scope.cpp: `TheScopeAnalysis::print` mixes `out` and `cout` printing (hidden behind `if (false)`)

Probably leftovers from debugging? Anyway, if you feel it's not necessary to change these, this PR is fine as is.

mhyee:
I fixed the issues you mentioned, and made a few other ones as well.

I skipped over the things that were more annoying to change, or were only in commented out code.

rir/src/interpreter/runtime.cpp
```diff
@@ -28,27 +28,45 @@ Function* isValidClosureSEXP(SEXP closure) {
...
-    Rprintf("  Size:            %u\n", f->size);
-    Rprintf("  Origin:          %p %s\n", f->origin(), f->origin() ? "" : "(unoptimized)");
-    Rprintf("  Next:            %p\n", f->next());
-    Rprintf("  Code objects:    %u\n", f->codeLength);
-    Rprintf("  Fun code addr:   %x (hex)\n", f->body());
-    Rprintf("  Invoked:         %u\n", f->invocationCount);
-    Rprintf("  Signature:       %p\n", f->signature);
-    if (f->signature)
-        f->signature->print();
+void printFunction(Function* f, std::ostream& out) {
```
mhyee, line 31: This actually looks uglier, but the advantage is that we can control where the printing happens, rather than trying to hack Rprintf.

o-:
thanks for doing another pass here!

PR #236, opened 2018-09-26
fix inliner

o-:
this code was broken. we would never find the safepoint after the call,
since we would start searching at the (side-effecting) call instruction
itself. for now let's be less clever. we need a better story for sp
anyway.

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -419,7 +419,7 @@ class VarLenInstructionWithEnvSlot
         Super::pushArg(env, RType::env);
     }
 
-    void pushArg(Value* a, PirType t) override {
+    void pushArg(Value* a, PirType t) override final {
         assert(a);
         assert(args_.size() > 0);
         assert(args_.back().type() == RType::env);
@@ -427,7 +427,7 @@ class VarLenInstructionWithEnvSlot
         args_.push_back(args_.back());
         args_[args_.size() - 2] = InstrArg(a, t);
     }
-    void popArg() override {
+    void popArg() override final {
         assert(args_.size() > 1);
         assert(args_.back().type() == RType::env);
         args_[args_.size() - 2] = args_[args_.size() - 1];
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -419,7 +419,7 @@ class VarLenInstructionWithEnvSlot
         Super::pushArg(env, RType::env);
     }
 
-    void pushArg(Value* a, PirType t) override final {
+    void pushArg(Value* a, PirType t) override {
         assert(a);
         assert(args_.size() > 0);
         assert(args_.back().type() == RType::env);
@@ -427,7 +427,7 @@ class VarLenInstructionWithEnvSlot
         args_.push_back(args_.back());
         args_[args_.size() - 2] = InstrArg(a, t);
     }
-    void popArg() override final {
+    void popArg() override {
         assert(args_.size() > 1);
         assert(args_.back().type() == RType::env);
         args_[args_.size() - 2] = args_[args_.size() - 1];
@@ -1118,15 +1118,21 @@ class VLIE(Safepoint, Effect::Any, EnvAccess::Leak) {
             pushArg(v);
     }
 
+    void updateNext(Safepoint* s) {
+        assert(inlined);
+        assert(arg(stackSize).type() == NativeType::safepoint);
+        arg(stackSize).val() = s;
+    }
+
     void next(Safepoint* s) {
         assert(!inlined);
         inlined = true;
         pushArg(s, NativeType::safepoint);
     }
 
-    Safepoint* next() {
+    Safepoint* next() const {
         if (inlined) {
-            auto r = Cast(arg(nargs() - 2).val());
+            auto r = Cast(arg(stackSize).val());
             assert(r);
             return r;
         } else {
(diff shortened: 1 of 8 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -1097,7 +1098,7 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
     };
 #endif
 
-    std::vector<FrameInfo*> synthesizeFrames;
+    std::deque<FrameInfo*> synthesizeFrames;
```
JanJecmen, line 1100: Is this just that if we have frames `a, b, c` then we want to synthesize them in the FIFO order? Was LIFO simply wrong? I mean, it kinda makes sense when we call eval as we go...

rir/tests/pir_regression.R
```diff
@@ -54,3 +54,30 @@ rir.compile(function(){
...
+
+# inlined safepoints:
+f <- rir.compile(function(x) g(x))
+g <- rir.compile(function(x) h(x))
+h <- rir.compile(function(x) 1+i(x))
+i <- rir.compile(function(x) 40-x)
+
+stopifnot(f(-1) == 42)
+
+hc1 = .Call("rir_invocation_count", h)
```
JanJecmen, line 67: This relies on the default that PIR_ENABLE is `on`, if this changes to say `force` in the future, this will break, no?
We could do `compiler::enableJIT(0)` and then be explicit about what is rir and what pir...

rir/src/interpreter/interp.cpp
```diff
@@ -10,6 +7,10 @@
 #include "ir/RuntimeFeedback_inl.h"
 #include "runtime.h"
 
+#include <alloca.h>
```
JanJecmen: I know it's a clang-format change, but it looks this is unused

rir/src/interpreter/interp.cpp
```diff
@@ -2711,13 +2713,14 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
 
 eval_done:
     while (!synthesizeFrames.empty()) {
```
JanJecmen, line 2702: I am unclear on this though: when we deopt, we jump here and want to redo everything unoptimized. We get the unoptimized version of the first frame, and evaluate it. Let's say that the second frame is for an inlined function, then by evaluating the unoptimized caller, we actually call the second function. Then we get back here and synthesize the second frame, and evaluate it, for the second time? What is the piece that I am missing?

rir/src/compiler/pir/instruction.h
```diff
@@ -419,15 +419,15 @@ class VarLenInstructionWithEnvSlot
         Super::pushArg(env, RType::env);
     }
 
-    void pushArg(Value* a, PirType t) override final {
```
JanJecmen: Why not final anymore? There is no further overriding in this PR

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -854,16 +854,6 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
-
-                auto dt = DispatchTable::unpack(mkfuncls->code);
-
-                if (dt->capacity() > 1 && !dt->available(1)) {
-                    Pir2Rir pir2rir(compiler, mkfuncls->fun, nullptr, dryRun,
-                                    compiler.logger.get(mkfuncls->fun));
-                    auto rirFun = pir2rir.finalize();
-                    if (!dryRun)
-                        dt->put(1, rirFun);
-                }
```
JanJecmen, line 866: Why is this not necessary any more?

JanJecmen (approved):

rir/tests/pir_regression.R
o-: it does work with all the modes. the trick is that the pir.compile below will recompile it with the typefeedback

rir/src/interpreter/interp.cpp
o-: yeah, it was just plain wrong.

rir/src/interpreter/interp.cpp
o-: Think about it this way:

o-: Original Function:
```
- Inlinee Frame
- Caller Frame
- rest
```

Optimized Function
```
- Combined Frame
- rest
```

On deoptimization we convert it to this:
```
- Callee Frame   # synthesizeFrame = [CallerFrame]
- rest
```

we directly rewrite the top fram, and push every frame but the top frame into the synthesize Frame. Then when the callee returns, instead of returning to `rest` we do this:

```
- Caller Frame  # synthesizeFrame = []
- rest
```

rir/src/compiler/translations/pir_2_rir.cpp
o-: this is creating an endless loop if we have a function static call itself...

rir/src/interpreter/interp.cpp
JanJecmen: Thanks, great explanation. I guess I missed that the evals take a pc, so they don't run from the start.
Now I see only one problem, if the callee frame touches the caller frame, but that we already discussed earlier

rir/src/compiler/translations/pir_2_rir.cpp
JanJecmen: So where does this take place now? Or do we just use the unoptimized version?

o-: no replacement. we call through the dispatch table anyway and will just optimize it later

PR #237, opened 2018-09-26
move instructions out of edge-split nodes

o-:
We keep our pir cfg in edge-split format. What this means is that
every BB can be a merge block, or a split block, but never both.
This avoids pathological graphs like:

       a     b
     /  \  /
    c    d

Where instructions from d cannot be hoisted without being captured by
an unrelated path and instructions from a cannot be pushed down,
without being captured by an unrelated path.

Our cleanup respects this condition. But if a BB is only there to
guarantee edge-split form, we should move all instructions to the
next BB. Many optimizations are BB local, thus it's not a good idea
to have a BB split over two.

### DIFF
--- a/rir/src/compiler/opt/cleanup.cpp
+++ b/rir/src/compiler/opt/cleanup.cpp
@@ -129,15 +129,11 @@ class TheCleanup {
         CFG cfg(function);
         std::unordered_map<BB*, BB*> toDel;
         Visitor::run(function->entry, [&](BB* bb) {
-            // Remove unnecessary splits
-            if (bb->isJmp() && cfg.hasSinglePred(bb->next0) &&
-                /* this condition keeps graph in split-edge: */
-                cfg.hasSinglePred(bb) &&
-                cfg.immediatePredecessors(bb)[0]->isJmp()) {
+            // If bb is a jump to non-merge block, we merge it with the next
+            if (bb->isJmp() && cfg.hasSinglePred(bb->next0)) {
                 BB* d = bb->next0;
-                while (!d->isEmpty()) {
+                while (!d->isEmpty())
                     d->moveToEnd(d->begin(), bb);
-                }
                 bb->next0 = d->next0;
                 bb->next1 = d->next1;
                 d->next0 = nullptr;


rir/src/compiler/opt/cleanup.cpp
```diff
@@ -130,20 +130,24 @@ class TheCleanup {
         std::unordered_map<BB*, BB*> toDel;
         Visitor::run(function->entry, [&](BB* bb) {
             // Remove unnecessary splits
-            if (bb->isJmp() && cfg.hasSinglePred(bb->next0) &&
+            if (bb->isJmp() && cfg.hasSinglePred(bb->next0)) {
                 /* this condition keeps graph in split-edge: */
```
charig: From what I understand, this second condition is conservative. I mean *bb* could have 2 predecessors, each just being a jump and I think that does not mean the edge can not be removed without breaking the condition. If I am wrong ok, otherwise a comment clarifying would help.

Beyond that, it would be great to have both `ifs` following the same order for the conditions:
1. isJmp() && 
2. hasSinglePred()

Then, maybe 1 && 2 could be added as a method to cfg. I am not sure the proper name for that, maybe isIntermediary?

charig (approved):

rir/src/compiler/opt/cleanup.cpp
o-: now you confused myself. while the condition is conservative regarding whether the graph is still edge-split, it is of course a necessary condition for the question whether I can dump instructions into the block. I mean, if not `cfg.hasSinglePred(bb->next0)` that basically means:
```
   bb      other
    \    /
      next0
```

so if i merge `bb` with `next0`, or if i dump instructions into `next0`, in both cases bad things will happen.

so the first `if` is a necessary and sufficient condition wheter I can flow instructions from `bb` to `next0`, and the second if is a neccessary and sufficient condition on the question whether i can merget `bb` and `next0` into one, without breaking edge-split.

o-: ignore what I wrote, I think it's still not quite accurate

charig (approved):

PR #238, opened 2018-09-27
Smaller type feedback struct

o-:
after many iterations I think this PR finally strikes a balance between fast decoding of bytecodes and still having the option for bytecodes to refer to external data.

The high level idea is, that you do `BC::decode(pc)` and this will create a BC object, that captures all the required information of a BC, and that also lets you dump that bc to a `CodeStream` again.

This nice idea breaks down, if the BC refers to external data, or gets painfully slow if the BC it is variable length.

So in a recent PR i removed all the external dependencies and made BC contain everything necessary. But now I wanted to move the recorded call targets of the `record_call_` bytecode into the code object constant pool. This means that decoding a BC also needs to access the code object (to read out the actual targets).

The solution in this PR is twofold:

1. there are two decodes now: `decode_shallow` only reads local and fixed lenght information. This is enough get most of the bc and for example know its lenght and so on.
2. the full `decode` allocates (only if needed, ie. currently for `call_implicit`, `call_named` and `record_call`) an additional `ExtraInformation` struct, that contains all the variable length, or external data. This is needed if you want to print the BC in full, or write it again to a code stream.

As a concrete usecase this allows us to store the recorded call targets in the code object extra pool. When decoding a `record_call_`, we just read the targets from the extra pool and store them in a `RecordCallExtraInformation` struct that is associated with the BC object.

By moving the extra information into a heap allocated object, we shave the old `callArgumentNames` and `implicitCallArguments` from all the BC objects that don't need them. This should roughtly halve the size of most BC objects..

### DIFF
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
@@ -100,7 +100,7 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
     std::unordered_map<Opcode*, std::vector<Opcode*>> incom;
     // Mark incoming jmps
     for (auto pc = srcCode->code(); pc != srcCode->endCode();) {
-        BC bc = BC::decode(pc);
+        BC bc = BC::decode_shallow(pc);
         if (bc.isJmp()) {
             incom[bc.jmpTarget(pc)].push_back(pc);
         }
@@ -108,7 +108,7 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
     }
     // Mark falltrough to label
     for (auto pc = srcCode->code(); pc != srcCode->endCode();) {
-        BC bc = BC::decode(pc);
+        BC bc = BC::decode_shallow(pc);
         if (!bc.isUncondJmp() && !bc.isExit()) {
             Opcode* next = BC::next(pc);
             if (incom.count(next))
@@ -132,7 +132,8 @@ namespace pir {
 
 bool Rir2Pir::compileBC(
     const BC& bc, Opcode* pos, rir::Code* srcCode, RirStack& stack,
-    Builder& insert, std::unordered_map<Value*, CallFeedback>& callFeedback,
+    Builder& insert,
+    std::unordered_map<Value*, std::vector<SEXP>>& callFeedback,
     std::unordered_map<Value*, TypeFeedback>& typeFeedback) const {
     Value* env = insert.env;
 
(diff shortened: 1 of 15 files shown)


o-:
travis fail is fixed by #236, just need to merge that one first

rir/src/runtime/Code.cpp
```diff
@@ -91,7 +91,7 @@ void Code::disassemble(std::ostream& out) const {
             out << ":\n";
         }
 
-        BC bc = BC::decode(pc);
+        BC bc = BC::decode(pc, this);
```
mhyee, line 94: It's not obvious why this is `decode` and not `decode_shallow`. It took me some time to realize that the code below calls `bc.print`, which calls `printImmedateArgs`, which calls `callExtra` and `callFeedbackExtra`.

Is there a better way to make things like this clearer? A run-time error helps, but requires that we have appropriate coverage (and this has bitten us before).

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -779,7 +782,7 @@ Value* Rir2Pir::tryTranslate(rir::Code* srcCode, Builder& insert) const {
                 for (int i = 0; i < 2 && n < end; ++i, n = BC::next(n))
                     ;
                 if (n < end) {
-                    auto nextbc = BC::decode(n);
+                    auto nextbc = BC::decode(n, srcCode);
```
mhyee: And this looks like it could be `decode_shallow`?

rir/src/ir/RuntimeFeedback_inl.h
```diff
@@ -12,11 +12,11 @@ void CallFeedback::record(Code* caller, SEXP callee) {
     if (numTargets < MaxTargets) {
         int i = 0;
         for (; i < numTargets; ++i)
-            if (targets[i] == callee)
+            if (caller->getEntry(targets[i]) == callee)
```
mhyee: `caller->getExtraPoolEntry`

This is pretty nasty. Maybe we should make `getEntry` and `setEntry` private? It looks like only [Function.h](https://github.com/reactorlabs/rir/blob/8984e0ea0c77c227b3c56c3aca5b4ff1ae0ae3ee/rir/src/runtime/Function.h#L67) would break, but we could add a `setFunction` to `Code` (which already has a `function()` accessor).

mhyee (commented):
I think it's great that we can cut down on the size of BC objects.

But I worry that we're paying for this with extra complexity. It's not always obvious when `decode` or `decode_shallow` should be used, and the extra indirection is a bit hard to follow (e.g. reading stuff from the actual immediates and then storing into the extra information). I don't know if there's a better way.

rir/src/ir/RuntimeFeedback_inl.h
o-: whaa, good catch :) yeah, there needs to be some safeguard. I agree

o-:
> But I worry that we're paying for this with extra complexity. It's not always obvious when decode or decode_shallow should be used

I understand your concerns. but I currently see no easier solution and I think the failure modes a kind of straight forward here.

also it's not primarily about making the size of BC smaller (misleading branch name, i know) but about having the ability to attach additional "external stuff" to a BC, like the call targets.

o-:
ok, look at the commit hash, we have to merge this one :)

mhyee (approved):

PR #241, opened 2018-10-02
Speculation: from rir2pir translation to a pir optimization phase

charig:


rir/src/compiler/opt/cleanup.cpp
```diff
@@ -1,8 +1,8 @@
-#include "cleanup.h"
 #include "../pir/pir_impl.h"
 #include "../util/cfg.h"
 #include "../util/visitor.h"
 #include "R/r.h"
+#include "phase_definitions.h"
```
o-: I think the proper term would be "pass" and not phase

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -0,0 +1,58 @@
...
+namespace rir {
+namespace pir {
+
+void ElideEnvSpec::apply(Closure* function) const {
+
+    // Elide environments of binary operators in which both operators are
+    // primitive values
+    Visitor::run(function->entry, [&](BB* bb) {
+        auto ip = bb->begin();
+        Safepoint* lastSafepoint;
```
o-: this might be uninitialized, should set it to nullptr;

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -0,0 +1,58 @@
...
+        while (ip != bb->end()) {
+            Instruction* i = *ip;
+
+            // We assume that always before a binop that enables speculation
+            // there will be a safepoint inserted only for that purpose
+            if (Safepoint::Cast(i)) {
+                lastSafepoint = Safepoint::Cast(i);
+                ip++;
+                continue;
+            }
```
o-, line 36: it would be more robust if you add:
```
if (i->hasEffect())
  lastSafepoint = nullptr;
```
and then later on only do the optimization if you have a safepoint available

o-: because if there is an effect between deopt and safepoint it will happen twice when we deopt

rir/src/compiler/pir/instruction.h
```diff
@@ -195,6 +195,11 @@ class Instruction : public Value {
                "subclass must override env() if it uses env");
         assert(false && "this instruction has no env");
     }
+    virtual void elideEnv() {
```
o-: you could just write `arg(envSlot()).val() = Env::elided();` here, instead of a virtual method. if the instruction has no EnvSlot then it will complain there.

rir/src/interpreter/interp.cpp
```diff
@@ -2331,37 +2331,36 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
                      (vectorT == VECSXP)) &&
                     (XLENGTH(val) == 1 || vectorT == VECSXP)) { // 3
 
-                        int idx_ = -1;
+                    int idx_ = -1;
```
o-, line 2334: changes here and lines below looks accidental

rir/src/ir/RuntimeFeedback.h
```diff
@@ -13,7 +15,7 @@ struct Code;
 #pragma pack(push)
 #pragma pack(1)
 
-struct CallFeedback {
+struct ObservedCalles {
```
o-, line 17: is there a deeper reason for the renaming? runtime type feedback is kind of the established umbrella term, not?

rir/src/ir/RuntimeFeedback.h
```diff
@@ -78,11 +80,31 @@ struct TypeFeedback {
         return false;
     }
 };
-static_assert(sizeof(TypeFeedback) == sizeof(uint32_t),
+static_assert(sizeof(ObservedValues) == sizeof(uint32_t),
               "Size needs to fit inside a record_ bc immediate args");
 
+namespace pir {
```
o-: the pir namespace is for the compiler. if you really want the ProfiledValues struct to be in the pir namespace, you should put it under `/compiler` folder

rir/src/ir/RuntimeFeedback.h
```diff
@@ -78,11 +80,31 @@ struct TypeFeedback {
...
+
+struct ProfiledValues {
```
o-: This struct should be written as:
```
struct ProfiledValues {
    std::unordered_map<Value*, ObservedCalles> calles;
    std::unordered_map<Value*, ObservedValues> types;
}
```
That does effectively do the same thing, but with automatic ressource management.

rir/src/compiler/pir/closure.h
```diff
@@ -47,8 +47,8 @@ class Closure : public Code {
 
     std::vector<SEXP> argNames;
     std::vector<Promise*> defaultArgs;
-
     std::vector<Promise*> promises;
+    ProfiledValues* runtimeFeedback;
```
o-: Again, here you should really just write: `ProfiledValues runtimeFeedback;` there is no need for new and delete, since this object has the exact same lifetime as the Closure.

rir/src/ir/RuntimeFeedback.h
```diff
@@ -78,11 +80,31 @@ struct TypeFeedback {
...
+    std::unordered_map<Value*, ObservedCalles>* calles;
```
o-: maybe `callTargets`?

o- (commented):
lgtm (modulo comments and travis errors)

rir/src/interpreter/interp.cpp
charig: I think it is only clang-format...Could remove it from the PR if it hurts.

rir/src/ir/RuntimeFeedback.h
charig: Mmmm. I see the runtime feedback as the process. The struct is something different. Maybe the new struct I call now `ProfiledValues` could be named `RuntimeFeedback` but I am not sure. Anyway, I like it like this but no problem to rollback this naming

README.md
```diff
@@ -56,6 +56,8 @@ To print intermediate debug information, you have a number of options:
 
 We periodically [benchmark](documentation/benchmarking.md) the performance of the optimizer
 
+We periodically [benchmark](documentation/benchmarking.md) the performance of the optimizer
```
mhyee, line 59: This line got duplicated?

rir/src/ir/RuntimeFeedback.h
```diff
@@ -13,7 +14,7 @@ struct Code;
 #pragma pack(push)
 #pragma pack(1)
 
-struct CallFeedback {
+struct ObservedCalles {
```
mhyee, line 17: Typo: `ObservedCalles` -> `ObservedCallees`

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -0,0 +1,91 @@
...
+class PHASE(ScopeResolution, "Dead store elimination");
+
+/*
+ * ElideEnv removes envrionments which are not needed. It looks at all uses of
+ * a `MkEnv` instruction. If the environment does not leak, and none of the
+ * uses have any effect (besides changing the unnecessary environment), then it
+ * can be removed.
+ *
+ */
+
```
mhyee, line 36: Nitpick: remove this blank line.

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -0,0 +1,91 @@
...
+class PHASE(ElideEnv, "Elide environments not needed");
+
+/*
+ * This pass searches for dominating force instructions.
+ *
+ * If we identify such an instruction, and we statically know which promise is
+ * being forced, then it inlines the promise code at the place of the
+ * dominating force, and replaces all subsequent forces with it's result.
```
mhyee, line 44: Typo: it's ->its

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -0,0 +1,91 @@
...
+ *
+ */
+class PHASE(ForceDominance, "Eliminate redudant force instructions");
+
+/*
+ * DelayInstr tries to schedule instruction right before they are needed.
```
mhyee, line 50: Typo: instruction -> instructions

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -0,0 +1,69 @@
...
+                continue;
+            } else {
+                if (i->hasEffect()) {
+                    lastFrameState = nullptr;
+                    ip++;
+                    continue;
+                }
+            }
+
+            if ((Lte::Cast(i) || Gte::Cast(i) || Lt::Cast(i) || Gt::Cast(i) ||
```
mhyee, line 38: This looks ugly to me, but I don't know if there's a better way...

mhyee (commented):
Skimmed over this and found a few typos.

rir/src/compiler/opt/elide_env_spec.cpp
charig: You mean the concatenation of `||`? If that is the case, yes, agree. I'll try with a macro to check it it looks nicer...

charig:
Thanks for the comments @mhyee. Since it is already merged, I will address the typos in a following PR that builds on top.

PR #242, opened 2018-10-02
Colors

charig:
I hate red sorry. Promise it is my only whim :). In case red is more convenient for everyone I can just  change it locally.


o-:
closing this, since it's way outdated. But in general i am fine with changing the colors. So if you want to, just go ahead and push the changes.

Note: We already have red.
But feel free to eg. change red->orange and cyan->red.

PR #243, opened 2018-10-03
Redo the removal of nops

JanJecmen:
Just a proposal, not sure we want it. Basically, when removing those pesky `br_ 0` instructions, the problem was that the code object was preallocated and 5 nops were left after the return for every removed branch.

So I rewrote it such that it now uses a buffer and then memcopies to a properly sized sexp.

Opinions?

### DIFF
--- a/rir/src/ir/CodeStream.h
+++ b/rir/src/ir/CodeStream.h
@@ -22,7 +22,6 @@ class CodeStream {
     typedef unsigned PcOffset;
     PcOffset pos = 0;
     unsigned size = 1024;
-    unsigned nops = 0;
 
     FunctionWriter& function;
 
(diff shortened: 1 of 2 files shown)


o-:
hmm, I spent quite some effort to remove this memcpy.
I think it's not really an issue that there are some nops at the end of the function...
Alternatively, what I do with the nop removal: I just count the number of nops, and then allocate the code object too small, by the number of nops that I am going to remove. you could do the same thing here. instead of passing the number of nops to the function writer, we could just pass a general number of deletable instructions. then just for every 0 you patch in a branch, you increment this by 5.
or, alternatively, instead of patching 0 into the branch, just patch the whole thing with 5 nops, then you don't need to change the function writer at all

JanJecmen:
Yes, I remember the trouble you went into here... So, I will do it in the codestream as you suggested. I did this because it seemed more general, to remove anything we want and only allocate after. I don't like the buffer either :)

o-:
but it can be quite general, if you just add an "erase(pos)" method to the codestream, that overrides a full instruction with nops

JanJecmen:
true.. something like that is already there

PR #246, opened 2018-10-11
New optimization model

charig:
Implements: https://www.o1o.ch/lab/entry/46361g4zp9nveus4tl9vpzobk2rlpe7o2fdfdxvlaqi1f5kopcyqpjcftjqboifr.html.
The only change is the renaming of `assume() -> expectNot()`.

@o-: Please add the spec to the documentation folder so I can iterate it (I do not want to steal authorship).

Missing/todo things:
1. A pass that moves the operands foward for readability. For instance an addition like `1 + 1` translates to:

```
env     e0.0  = MkEnv            parent=<environment: R_GlobalEnv>
  real$   %0.1  = LdConst          [1] 1
  real$   %0.2  = LdConst          [1] 1
  fs      %0.3  = FrameState       0x7fb208e0cff0+45: [%0.1, %0.2], env=e0.0
  val     %0.4  = Add              %0.1, %0.2, e0.0
```

Then to: 

```
BB0
  real$   %0.0  = LdConst          [1] 1
  real$   %0.1  = LdConst          [1] 1
  void            Checkpoint        -> BB2 (by default) | BB1 (if coming from expect)
BB2
  env     e2.0  = MkEnv            parent=<environment: R_GlobalEnv>
  val     %2.1  = Add              %0.0, %0.1, e2.0
```

The operands are in another BB and this makes harder the readability. The problem right now is that the FrameState uses the operands and our verifier asks that the operands exists before their usage. Even moving the FrameState to the deopt branch the verifier for sure fails. 

2. Passes that remove additional basic blocks when the code finally do not find any specualtion after all the opt passes? 

3. Passes that try to split specualtive operations with no effects in between to the same framestate?

4. More tests.


rir/src/compiler/opt/cleanup.cpp
```diff
@@ -7,8 +7,8 @@
 #include <unordered_map>
 #include <unordered_set>
 
-namespace {
```
o-: that's not the same thing, anonymous namespace is on purpose, to avoid polluting the actual namespace

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -34,56 +34,64 @@ class PHASE(ScopeResolution, "Dead store elimination");
...
  */
-class PHASE(Inline, "Inline closure");
+class PASS(Inline, "Inline closures");
 
 /*
- * This phase goes through every operation that for the general case needs
- * an environment, but it could elide it for some particular inputs. The phase
+ * This PASS goes through every operation that for the general case needs
+ * an environment, but it could elide it for some particular inputs. The PASS
  * then analyzes the profiling information and if all the observed inputs are
```
o-: out of date comment

rir/src/compiler/pir/instruction.cpp
```diff
@@ -168,6 +168,20 @@ Value* Instruction::baseValue() {
     return this;
 }
 
+bool Instruction::maySpecialize() {
+    bool may = false;
+    if (false) {
+        // Do nothing
+    }
+#define V(Name)                                                                \
```
o-, line 172: you could:
```
if (Name::Cast(this))
  return true;
```

rir/src/compiler/transform/insert_cast.cpp
```diff
@@ -47,7 +47,12 @@ void InsertCast::apply(BB* bb) {
                 }
                 c->bb_ = bb;
                 arg.val() = c;
-                ip = bb->insert(ip, c) + 1;
+                if (instr->maySpecialize() && ip != bb->begin() &&
+                    FrameState::Cast(*(ip - 1))) {
```
o-: that is strange, can we remove it?

rir/src/compiler/pir/instruction_list.h
```diff
@@ -62,6 +62,8 @@
     V(Subassign2_1D)                                                           \
     V(ForSeqSize)                                                              \
     V(FrameState)                                                              \
+    V(Checkpoint)                                                              \
+    V(ExpectNot)                                                               \
```
o-: would you mind `assumeNot`. why I like assume: the idea is that the optimizer can *assume* that the thing is correct after the instruction (not just expect it to be true, but really do as if it where true).

rir/src/compiler/util/visitor.h
```diff
@@ -166,16 +168,35 @@ class VisitorImplementation {
                 done.set(cur->next1);
             }
 
+            if (!predicate(cur))
+                return false;
+
+            if (processNewNodes) {
```
o-: i find it strange that with `processNewNodes` we would visit the next BBs that where there before the visit *and* the ones that are next after the visit. I think we should either say, the next BBs are what they originally where, or what they they are after visiting the BB, but not both. So maybe just add an `if (!processNewNodes)` around 154-169.

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -1090,45 +1094,63 @@ static bool allLazy(CallType* call, std::vector<Promise*>& args) {
...
             auto next = it + 1;
-            if (FrameState::Cast(*it) || Deopt::Cast(*it)) {
+            if (FrameState::Cast(*it)) {
+                next = bb->remove(it);
+            } else if (Checkpoint::Cast(*it)) {
                 next = bb->remove(it);
+                // If a bb ending with a checkpoint is emptied we must
+                // "inform" the cleanup that it could replace it with the
+                // true branch
+                if (bb->isEmpty())
```
o-: next1 should always be set to null, because with the checkpoint gone, this is not a branching BB anymore. so not setting next1 to null actually breaks the cfg invarianst (only BB ending in Branch can have next1 set)

rir/src/compiler/opt/adapt_spec.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "../util/visitor.h"
+#include "R/r.h"
+#include "pass_definitions.h"
+
+#include <unordered_map>
+
+namespace rir {
+namespace pir {
+
+void AdaptForSpec::apply(Closure* function) const {
```
o-: I don't like that name. I would prefer something like  "addCheckpoints"? (dual to elideCheckpoints)

rir/src/compiler/opt/adapt_spec.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    Visitor::run(function->entry, [&](BB* bb) {
+        auto ip = bb->begin();
+        while (ip != bb->end()) {
+            Instruction* i = *ip;
+            Instruction* next = *(ip + 1);
```
o-: dereferencing an iterator outside of the collection is undefined behavior. So you are only allowed to do `*(ip+1)` *after* you checked `ip+1 != bb->end()`

o-: alternative:
```
if (bb->empty())
  return;
auto ip = bb->begin();
auto ipp = ip+1;
while (ipp != bb->end()) {
  auto next = *ipp;
  if (auto f = FrameState(*ip)) {
    if (next->maySpecialize() ...)
      ...
    ip = ipp;
    ipp++;
}
```

rir/src/compiler/pir/instruction.cpp
```diff
@@ -446,5 +460,22 @@ void CallImplicit::printArgs(std::ostream& out, bool tty) {
 
 FrameState* Deopt::frameState() { return FrameState::Cast(arg<0>().val()); }
 
+void Checkpoint::printArgs(std::ostream& out, bool tty) {
+    FixedLenInstruction::printArgs(out, tty);
+    out << " -> BB" << bb()->trueBranch()->id << " (by default) | BB"
+        << bb()->falseBranch()->id << " (if coming from expect)";
+}
+
+BranchInstruction* BranchInstruction::CastBranch(Value* v) {
```
o-: I think the verifier still has the explicit list of branch instructions instead of using this interface

rir/src/compiler/pir/instruction.cpp
```diff
@@ -446,5 +460,22 @@ void CallImplicit::printArgs(std::ostream& out, bool tty) {
...
+    switch (v->tag) {
```
o-, line 466:
```
case Tag::Branch:
case Tag::Checkpoint:
  return this;
```
should have the same semantics

rir/src/compiler/pir/instruction.h
```diff
@@ -585,12 +586,18 @@ class FLIE(StVar, 2, Effect::None, EnvAccess::Write) {
     void printArgs(std::ostream& out, bool tty) override;
 };
 
-class FLI(Branch, 1, Effect::None, EnvAccess::None) {
+// Common interface to all branch instructions
+class BranchInstruction {
```
o-: can we call this `BranchingInstruction` to avoid confusion with the actual `Branch` Instructions

o- (approved):
:tada:

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -34,56 +34,64 @@ class PHASE(ScopeResolution, "Dead store elimination");
...
+ * the version operation without an environment, it avoids creating the
+ * environment and add the corresponding guard to deoptimize in case an
+ * incompatible input appears at run time.
+ */
+class PASS(ElideEnvSpec, "Speculate on values to elide environments");
 
-class PHASE(Cleanup, "...");
-class PHASE(CleanupFrameState, "...");
+class PASS(Cleanup, "...");
+class PASS(CleanupFrameState, "...");
```
o-: I know off-topic, but maybe also fill in the ..., since they look a bit strange on the disassembly

rir/src/compiler/opt/adapt_spec.cpp
charig: Ok. I changed it to insertCheckpoints. But we do not have an elideCheckpoint pass...

rir/src/compiler/opt/adapt_spec.cpp
charig: Thanks

rir/src/compiler/pir/instruction.cpp
charig: I do not get this one

rir/src/compiler/pir/instruction_list.h
charig: For the record. I am not sure assume is the proper name. I think the proper name is a mixture between expect and assume. It is true that since that instruction, the opt can assume the condition to hold. But at the same time the semantics of the operation requires to test the condition. It is like a conditional assume. That is usually a guard. Maybe guard is a name to reconsider. Anyway, I followed your suggestion. Since you are going to work on top, maybe you come up to a better name meanwhile.

PR #247, opened 2018-10-25
added rir.printInvocation

Jakobeha:
I think this is the first assignment you gave me, not sure if I'm doing this right, I added rir.printInvocation which simply prints how many times a function was called (technically how many times each "slot" was called).

I think you gave me another, "harder" assignment, but I don't see it in the e-mail. Can you remind me what it is again?


mhyee:
Thanks! I have a few comments.

1. We're probably not going to merge this into our code, because it's a "starter" project (and we have another student doing it too). Same will probably apply to the second assignment.

2. I haven't actually run your code, but it looks right.

3. For future work, it will be easier if you create a separate branch in your own repository, so that your `master` branch is clean with respect to the `reactorlabs` repo. Then you can easily fetch new changes without worrying about merge conflicts.

4. For some reason, all the files had their permissions changed. You'll probably want to look into it and fix it.

As for the harder assignment, I called it "step 2." It should be in the email, but I've pasted it below:

> Add a function .printInvocation that is compiled to a RIR bytecode
> Need to add RIR and PIR bytecodes, update the compilers, update the interpreter
> 
> Part of this will be similar to the PR: https://github.com/reactorlabs/rir/pull/176
> But the RIR bytecode and interpreter were already implemented.
> 
> Start looking in
> https://github.com/reactorlabs/rir/blob/master/rir/src/ir/insns.h#L437
> https://github.com/reactorlabs/rir/blob/master/rir/src/interpreter/interp.cpp#L2693

Jakobeha:
I finished step 2:

Now if you put ".printInvocation" in a function, whenever the function is run it'll print the invocation count. It'll even print it before the function gets compiled ("Invocation count: 0 (not compiled)").

rir/src/runtime/Function.cpp
```diff
@@ -10,4 +10,9 @@ void Function::disassemble(std::ostream& out) {
         }
     }
 }
+
+void Function::printInvocation(std::ostream& out) {
```
mhyee, line 14: It looks like this isn't used anywhere, so you could just delete it.

mhyee (commented):
I only have one comment, other than what I said above. Looks good!

I'll get back to you later about what you could work on next.

mhyee:
Superseded by #270.

Issue #215, opened 2018-09-10
Remove the code->function pointer


mhyee:
As of #198, here's what the pointers look like:

![2018_09_06 11_00 office lens](https://user-images.githubusercontent.com/547059/45308832-64238400-b4f0-11e8-875d-b5cc4414ba54.jpg)

The problem is that if a promise escapes, it keeps its Code object alive, which keeps its Function alive, which keeps all associated Code objects (i.e. the code object for the function, its promises, the promises' promises, etc.) alive.

Ideally, we would remove the Code object's pointer to its function. But the compiler and interpreter use the pointer to find a function's implicit arguments and promises, and changing this seems like it would be very invasive.

PR #251, opened 2018-10-31
remove code->function pointer

o-:
this removes the pointer from code objects to their function. Thus
this prevents escaping promises from keeping functions live.

To that end we change the function format from

    Function
      Code* promises[]    // list of body, formal args and all (nexted) promises

    Code
      justBC

to:

    Function
      Code* defaultArgs[]
      Code* body

    Code
      Code* promises[]
      justBC

Note: this leads to nested promises. For example `function() a(b(c,d))` is:

    function -> a -> b --> c
                       \-> d

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -34,7 +34,7 @@ REXPORT SEXP rir_disassemble(SEXP what, SEXP verbose) {
             continue;
         Function* f = t->at(entry);
         std::cout << "= vtable slot <" << entry << "> (" << f << ", invoked "
-                  << f->invocationCount << ") =\n";
+                  << f->invocationCount() << ") =\n";
         f->disassemble(std::cout);
     }
 
@@ -203,7 +203,7 @@ REXPORT SEXP rir_invocation_count(SEXP what) {
 
     SEXP res = Rf_allocVector(INTSXP, dt->capacity());
     for (size_t i = 0; i < dt->capacity(); ++i) {
-        INTEGER(res)[i] = dt->available(i) ? dt->at(i)->invocationCount : 0;
+        INTEGER(res)[i] = dt->available(i) ? dt->at(i)->invocationCount() : 0;
     }
 
     return res;
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -194,6 +194,7 @@ SEXP pirCompile(SEXP what, const std::string& name, pir::DebugOptions debug) {
     return what;
 }
 
+// Used in test infrastructure for counting invocation of different versions
 REXPORT SEXP rir_invocation_count(SEXP what) {
     if (!isValidClosureSEXP(what)) {
         Rf_error("not a compiled closure");
@@ -202,9 +203,8 @@ REXPORT SEXP rir_invocation_count(SEXP what) {
     assert(dt);
 
     SEXP res = Rf_allocVector(INTSXP, dt->capacity());
-    for (size_t i = 0; i < dt->capacity(); ++i) {
+    for (size_t i = 0; i < dt->capacity(); ++i)
         INTEGER(res)[i] = dt->available(i) ? dt->at(i)->invocationCount() : 0;
-    }
 
     return res;
 }
--- a/rir/src/interpreter/runtime.cpp
+++ b/rir/src/interpreter/runtime.cpp
@@ -26,49 +26,6 @@ Function* isValidClosureSEXP(SEXP closure) {
     return nullptr;
 }
 
-// for now, we will have to rewrite this when it goes to GNU-R proper
-
-void printFunction(Function* f, std::ostream& out) {
-    out << "Function object (" << static_cast<void*>(f) << "):\n";
-    out << std::left << std::setw(20) << "  Magic:" << std::hex << f->info.magic
-        << std::dec << " (hex)\n";
-    out << std::left << std::setw(20) << "  Size:" << f->size << "\n";
-    out << std::left << std::setw(20) << "  Origin:";
-    if (f->origin()) {
-        out << static_cast<void*>(f->origin()) << "\n";
-    } else {
-        out << "(nil) (unoptimized)\n";
-    }
-    out << std::left << std::setw(20) << "  Next:";
-    if (f->next()) {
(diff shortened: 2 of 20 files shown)


o-:
closes #215

rir/src/api.cpp
```diff
@@ -203,7 +203,7 @@ REXPORT SEXP rir_invocation_count(SEXP what) {
 
     SEXP res = Rf_allocVector(INTSXP, dt->capacity());
     for (size_t i = 0; i < dt->capacity(); ++i) {
-        INTEGER(res)[i] = dt->available(i) ? dt->at(i)->invocationCount : 0;
+        INTEGER(res)[i] = dt->available(i) ? dt->at(i)->invocationCount() : 0;
```
mhyee, line 207: I didn't realize this function existed. (And it's part of what we asked our undergraduate helpers to work on.) We don't expose it in `rir.R`, but it looks like we use it internally for the `pir_regression.R` tests.

I know this is unrelated, but add a comment about how this is only used as a test helper?

rir/src/interpreter/runtime.cpp
```diff
@@ -26,49 +26,6 @@ Function* isValidClosureSEXP(SEXP closure) {
     return nullptr;
 }
 
-// for now, we will have to rewrite this when it goes to GNU-R proper
-
-void printFunction(Function* f, std::ostream& out) {
```
mhyee, line 31: Thanks for removing this. I wonder where this came from originally, and why we no longer call it...

mhyee: There is an `extern void printFunction(Function* f)` in `interp.cpp` that should also be removed.

rir/src/runtime/Function.h
```diff
@@ -50,108 +50,65 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
...
               // GC area includes the SEXPs before the Code objects array
-              CODEOBJ_OFFSET + codeVec.size()),
-          size(functionSize), signature(nullptr), invocationCount(0),
-          deopt(false), markOpt(false), codeLength(codeVec.size()),
-          origin_(nullptr), next_(nullptr) {
-        for (size_t i = 0; i < codeLength; ++i) {
-            Code* c = Code::unpack(codeVec[i]);
-            // Set c->function_ to point to this Function's container
-            c->setEntry(0, container());
+              NUM_PTRS + defaultArgs.size()),
```
mhyee, line 49: Comment above needs to be updated.

rir/src/runtime/Function.h
```diff
@@ -50,108 +50,65 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
...
+          size(functionSize), signature(nullptr), deopt(false), markOpt(false),
+          numArgs(defaultArgs.size()) {
+        origin(nullptr);
+        next(nullptr);
+        for (size_t i = 0; i < numArgs; ++i) {
             // Set codeObjects[i] to point to Code object c's container
-            setEntry(CODEOBJ_OFFSET + i, c->container());
+            setEntry(NUM_PTRS + i, defaultArgs[i]);
```
mhyee, line 55: Comment above needs to be updated.

rir/src/runtime/Function.h
```diff
@@ -50,108 +50,65 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
...
-        for (Code* c : *this) {
-            if (c == code)
-                return idx;
-            ++idx;
-        }
-        assert(false);
-        return 0;
-    }
+    Code* body() { return Code::unpack(getEntry(2)); }
+    void setBody(SEXP body) { setEntry(2, body); }
```
mhyee: Should this be `void setBody(SEXP body)` or `void body(SEXP body)`? Below, you use `origin(...)` and `next(...)` as setters.

rir/src/runtime/Function.h
```diff
@@ -50,108 +50,65 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
     friend class FunctionCodeIterator;
```
mhyee, line 39: The comment above (lines 24-46) need to be updated.

rir/src/runtime/Function.h
```diff
@@ -50,108 +50,65 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
...
     // !!! Furthermore, you need to update                 !!!
     // !!!     the CODEOBJ_OFFSET constant.                !!!
 
-    FunctionSEXP origin_; /// Same Function with fewer optimizations,
-                          //   NULL if original
-    FunctionSEXP next_;
-
-    CodeSEXP codeObjects[]; /// Pointers to CodeSEXPs (Code objects embedded
-                            //   inside EXTERNALSXP)
+    CodeSEXP locals[NUM_PTRS];
```
mhyee, line 93: The comment above needs to be updated. Also, it would be useful to explain what each of the entries in `locals` is used for, i.e. 0 is origin, 1 is next, 2 is the body code object.

rir/src/ir/BC_inc.h
```diff
@@ -213,6 +213,21 @@ class BC {
                bc == Opcode::push_code_;
     }
 
+    void addPromargsTo(std::vector<FunIdx>& proms) {
```
mhyee: `addPromArgsTo`? Or maybe `addToPromArgs`? Because we're adding `bc` to the `proms` vector.

mhyee (approved):

rir/src/ir/Compiler.cpp
```diff
@@ -820,8 +820,9 @@ void compileCall(Context& ctx, SEXP ast, SEXP fun, SEXP args) {
 
         // (1) Arguments are wrapped as Promises:
         //     create a new Code object for the promise
-        size_t prom = compilePromise(ctx, *arg);
-        callArgs.push_back(prom);
+        Code* prom = compilePromise(ctx, *arg);
```
JanJecmen, line 823: @o- I think here and then later down with the default args, these promises can get collected before they are written into the extra pool.

PR #252, opened 2018-10-31
fix unprotected promises before functions are finalized

o-:
a GC could release the compiled promises, since they are only reachable
through the function.

thanks @JanJecmen for pointing this out

### DIFF
--- a/rir/src/ir/Compiler.cpp
+++ b/rir/src/ir/Compiler.cpp
@@ -908,15 +908,13 @@ SEXP Compiler::finalize() {
 
     FunctionSignature* signature = new FunctionSignature();
 
-    std::vector<SEXP> defaultArgs;
-
     // Compile formals (if any) and create signature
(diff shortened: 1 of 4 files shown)


mhyee:
Have you tried running with `GCTORTURE`?

rir/src/ir/Compiler.cpp
```diff
@@ -913,10 +913,10 @@ SEXP Compiler::finalize() {
     // Compile formals (if any) and create signature
     for (auto arg = RList(formals).begin(); arg != RList::end(); ++arg) {
         if (*arg == R_MissingArg) {
-            defaultArgs.push_back(nullptr);
```
JanJecmen, line 916: cpp check complains about unused `defaultArgs`

JanJecmen (approved):
looks good

o-:
> Have you tried running with GCTORTURE?

didn't trigger the bug. guess our compiler does not alloc much.
but good idea. I created #253

PR #253, opened 2018-11-01
run tests with gctorture on travis

o-:
this change runs `./bin/tests` once with gc torture.

also clean up our test matrix a bit. we do not need to run all of
the tests with valgrind, one run is enough.


o-:
tbr

mhyee:
Looked at this and it seems fine. (Guess I can't approve if it's already merged.)

mhyee (commented):
Looks fine. (Guess I can't approve if it's already merged.)

PR #254, opened 2018-11-01
improve scheduling of branches in pir2rir

o-:
1. emit brtrue instead of brfalse
2. order branches such that the false branch is fallthrough
3. schedule deopt branches last

long term we might want some heuristic when to emit brtrue and when
to emit brfalse. for now brtrue is much better, since all our
deopts are on the condition `isobj_`.


JanJecmen (approved):
Seems ok

PR #255, opened 2018-11-01
make controlflow a property of an instruction

o-:
this commit makes explicit the notion of controlflow of an instruction

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -88,6 +88,13 @@ enum class Effect : uint8_t {
     Any,
 };
 
+// Controlflow of instruction.
+enum class Controlflow : uint8_t {
+    None,
+    Exit,
+    Branch,
+};
+
 class Instruction : public Value {
   public:
     struct InstructionUID : public std::pair<unsigned, unsigned> {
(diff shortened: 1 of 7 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -242,6 +256,11 @@ class InstructionImplementation : public Instruction {
     bool hasEnv() const final {
         return mayAccessEnv() && env() != Env::elided();
     }
+    bool exits() const final { return Description.Exits; }
+    bool branches() const final { return Description.Branches; }
+    bool hasControlflow() const final {
```
charig: changesFlow?

rir/src/compiler/pir/instruction.h
```diff
@@ -88,6 +88,13 @@ enum class Effect : uint8_t {
     Any,
 };
 
+// Controlflow of instruction.
+enum class Controlflow : uint8_t {
+    None,
+    Exit,
```
charig, line 94: This may be a sensible list for our purposes. But it is not clear why `exit` is a cf for returns & deopt but `calls` are not.

charig (approved):

rir/src/compiler/pir/instruction.h
o-: good point. so far I see no issue in adding a call property later, except the hasControlflow would be wrong. so resolving your above comment should fix both issues, right?

PR #258, opened 2018-11-02
add framestate to the call instructions

o-:
for inlining we need a framestate that corresponds to the call instruction.
we used to just put a framestate after the call and then let the inliner
linearly search for it. This commit places the framestate right before
the call (which is more correct) and adds it as an argument to the call
instruction. thanks to that it is always unambiguous which framestate belongs
to the call.

### DIFF
--- a/rir/src/compiler/opt/inline.cpp
+++ b/rir/src/compiler/opt/inline.cpp
@@ -33,6 +33,7 @@ class TheInliner {
                 Closure* inlinee = nullptr;
                 Value* staticEnv = nullptr;
 
+                FrameState* callerFrameState = nullptr;
                 if (auto call = Call::Cast(*it)) {
                     auto mkcls = MkFunCls::Cast(call->cls()->baseValue());
                     if (!mkcls)
@@ -41,6 +42,7 @@ class TheInliner {
                     if (inlinee->argNames.size() != call->nCallArgs())
                         continue;
                     staticEnv = mkcls->lexicalEnv();
+                    callerFrameState = call->frameState();
                 } else if (auto call = StaticCall::Cast(*it)) {
                     inlinee = call->cls();
                     // if we don't know the closure of the inlinee, we can't
@@ -54,6 +56,7 @@ class TheInliner {
                         R_IsNamespaceEnv(inlinee->closureEnv()->rho))
                         continue;
                     staticEnv = inlinee->closureEnv();
+                    callerFrameState = call->frameState();
                 } else {
                     continue;
                 }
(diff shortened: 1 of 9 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -892,34 +892,130 @@ UNOP(Length);
...
+    void printEnv(std::ostream& out, bool tty) override final{};
+};
+
 // Common interface to all call instructions
 class CallInstruction {
   public:
     virtual size_t nCallArgs() = 0;
     virtual void eachCallArg(Instruction::ArgumentValueIterator it) = 0;
     static CallInstruction* CastCall(Value* v);
+    virtual void clearFrameState(){};
```
JanJecmen, line 989: Not sure I understand why only `Call` and `StaticCall` can clear the frame state?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -272,12 +272,14 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, rir::Code* srcCode,
 
         auto ast = bc.immediate.callFixedArgs.ast;
         auto insertGenericCall = [&]() {
-            if (bc.bc == Opcode::named_call_implicit_)
+            if (bc.bc == Opcode::named_call_implicit_) {
```
JanJecmen, line 275: Why do the named calls not have a frame state?

rir/src/compiler/opt/inline.cpp
```diff
@@ -120,11 +105,6 @@ class TheInliner {
                                 auto cloneSp =
                                     FrameState::Cast(copyFromFs->clone());
 
-                                // Remove the inlinee result from the
```
JanJecmen, line 123: Is this removed because now the framestate is before the call, so the result of the call is not on its stack?

JanJecmen (approved):
Looks good

rir/src/compiler/pir/instruction.h
o-: we currently only inline call and static call. so I added the framestate only to those two

rir/src/compiler/opt/inline.cpp
o-: yes

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: because we do not inline them

rir/src/compiler/pir/instruction.h
o-: so for the others there is nothing to clear...

PR #259, opened 2018-11-03
constand folding and dead branch removal

o-:


rir/src/compiler/opt/constantfold.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace rir {
+namespace pir {
+
+#define FOLD(Instruction, Operation)                                           \
+    do {                                                                       \
+        if (auto instr = Instruction::Cast(i)) {                               \
+            if (auto lhs = LdConst::Cast(instr->arg<0>().val())) {             \
+                if (auto rhs = LdConst::Cast(instr->arg<1>().val())) {         \
+                    auto res = Rf_eval(Rf_lang3(Operation, lhs->c, rhs->c),    \
```
JanJecmen, line 30: I think this might be a protect bug, since `res` is neither protected nor inserted into the constant pool?

Edit: And are similar issues possible elsewhere in pir? This call to `eval` might collect anything that pir has created and not protected

JanJecmen (commented):
Good job! Maybe add a comment to the removal of the branch instruction, please, took me a while to realize what's going on there

PR #261, opened 2018-11-12
fix framestate after static call

o-:
the framestate for the static call accidentially included the callee,
since the pop was after the insertion of the framestate.

this was only revealed when i started optimizing functions later.
thus this commit also introduces 3 different optimization trigger
values on travis, to get more coverage.


o-:
tbr

charig:
The WARMUP env var is very useful for testing. Thanks!

PR #267, opened 2018-11-13
The big PR

o-:
various fixes and cleanups. see the individual commit messages


o-:
@JanJecmen @mhyee what do you suggest, how do we handle this PR? do you want me to run you through it?

mhyee:
With a PR this big, maybe we should just schedule a code walk and then merge. That way, we would at least know what sort of things have changed.

I don't think any of us have the time (or willpower) to review this line-by-line.

JanJecmen:
Agreed, how about Monday?

o-:
yeah, absolutely. but I mean it might also show the limitations of the reviewing, maybe we just don't have enough time to do it? but anyway. I am happy to go thru it on mon

rir/src/compiler/opt/cleanup_checkpoints.cpp
```diff
@@ -0,0 +1,35 @@
...
+                    assert(bb->next1->isExit() &&
+                           "deopt blocks should be just one BB");
+                    bb->next1 = nullptr;
+                }
+            }
+        });
+        BBTransform::removeBBs(code, toDelete);
+    };
+    apply(function);
+    for (auto& p : function->promises)
```
charig: function->eachPromise(apply)

charig: Won't the compiler run this pass in the promises automatically? Because if that is the case we can avoid the closure here that harms code readability.

rir/src/compiler/opt/cleanup_checkpoints.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace rir {
+namespace pir {
+
+void CleanupCheckpoints::apply(RirCompiler&, Closure* function) const {
+    auto apply = [](Code* code) {
+        std::unordered_set<BB*> toDelete;
+        Visitor::run(code->entry, [&](BB* bb) {
+            if (bb->isEmpty())
+                return;
+            if (auto cp = Checkpoint::Cast(bb->last())) {
```
charig, line 17: I guess you add this here instead of in cleanup because cleaning checkpoints imply changing the cfg. I wonder if 1. It is time to rename cleanup to something like basicCleanup and 2. it may be worth to merge cleanupCheckpoints with cleanupFramestate in a cleanupSpeculation phase?

rir/src/compiler/opt/cleanup_framestate.cpp
```diff
@@ -7,25 +8,13 @@
...
 
-void CleanupFrameState::apply(RirCompiler&, Closure* function) const {
+void CleanupFramestate::apply(RirCompiler&, Closure* function) const {
     auto apply = [](Code* code) {
         Visitor::run(code->entry, [&](Instruction* i) {
             if (auto call = CallInstruction::CastCall(i)) {
                 call->clearFrameState();
             }
         });
-
```
charig, line 17: This gots removed because a later cleanup will remove it? If that is the case I see it a little brittle.

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -108,24 +109,18 @@ void Constantfold::apply(RirCompiler& cmp, Closure* function) const {
...
     }
-    // Dead code can still appear as phi inputs in live blocks
-    Visitor::run(function->entry, [&](Instruction* i) {
-        if (auto phi = Phi::Cast(i)) {
-            phi->removeInputs(deleted);
-        }
-    });
-    // Remove the actual branch instruction
+
+    BBTransform::removeBBsWithChildren(dom, function, toDelete);
```
charig: removeBranch or removeBranchSince?

rir/src/compiler/opt/inline.cpp
```diff
@@ -61,6 +61,9 @@ class TheInliner {
                     continue;
                 }
 
+                if (inlinee == function && fuel < 5)
```
charig: This means we do not inline recursive functions? Maybe a comment on that?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -37,7 +37,13 @@ class TheScopeResolution {
                 else if (ldfun)
                     ld = ldfun;
 
-                if (sld) {
+                if (ldfun && ldfun->guessedBinding() &&
```
charig: Line 31: Instruction* ld = LdVar::Cast(i); The rhs is unneeded?

rir/src/compiler/pir/instruction.h
```diff
@@ -498,18 +499,31 @@ class FLI(LdConst, 0, Effect::None, EnvAccess::None) {
...
-        : FixedLenInstructionWithEnvSlot(RType::closure, env), varName(name) {
+        : FixedLenInstructionWithEnvSlot(RType::closure, {{PirType::any()}},
+                                         {{Tombstone::closure()}}, env),
+          varName(name) {
         assert(TYPEOF(name) == SYMSXP);
     }
 
+    void clearGuessedBinding() { arg<0>().val() = Tombstone::closure(); }
+
+    void guessedBinding(Value* val) { arg<0>().val() = val; }
```
charig, line 526: I do not fully get the name here. Guessed sounds to me like sth that we are not 100% sure. Why not just binding and in case we do not know just return some special `UNKNOWN` value.

rir/src/compiler/pir/instruction.h
```diff
@@ -498,18 +499,31 @@ class FLI(LdConst, 0, Effect::None, EnvAccess::None) {
     void printArgs(std::ostream& out, bool tty) override;
 };
 
-class FLIE(LdFun, 1, Effect::Any, EnvAccess::Write) {
+class FLIE(LdFun, 2, Effect::Any, EnvAccess::Write) {
```
charig, line 509: What is the advantage of having the eventual binding as an instruction argument instead of a class instance variable?

rir/src/compiler/pir/singleton_values.h
```diff
@@ -44,13 +44,28 @@ class Missing : public SingletonValue<Missing> {
...
+            out << "fs";
+        else
+            assert(false);
+    }
+    static Tombstone* closure() {
+        static Tombstone cls(RType::closure);
+        return &cls;
+    }
+    static Tombstone* framestate() {
+        static Tombstone cls(NativeType::frameState);
```
charig: fs?

rir/src/compiler/pir_tests.cpp
```diff
@@ -317,7 +317,7 @@ bool checkPir2Rir(SEXP expected, SEXP result) {
 
 bool testPir2Rir(const std::string& name, const std::string& fun,
                  const std::string& args, bool useSame = false,
-                 bool verbose = false) {
+                 bool verbose = true) {
```
charig: ?

rir/src/utils/configurations.cpp
```diff
@@ -51,13 +51,22 @@ void Configurations::defaultOptimizations() {
     };
 
     optimizations.push_back(new pir::insertCheckpoints);
-    for (int j = 0; j < 3; ++j) {
-        for (int i = 0; i < 3; ++i) {
+
+    for (int i = 0; i < 3; ++i) {
+        for (int j = 0; j < 2; ++j) {
+            addDefaultOpt();
```
charig: Why two times the default opts? This phase ordering is becoming tricky. 
Maybe we should start documenting somehow the reason whenever we change sth here.

rir/src/utils/configurations.cpp
```diff
@@ -51,13 +51,22 @@ void Configurations::defaultOptimizations() {
...
             addDefaultOpt();
             optimizations.push_back(new pir::Cleanup());
             optimizations.push_back(new pir::Inline());
         }
-        optimizations.push_back(new pir::CleanupFrameState());
+        // This pass removes unused checkpoints
```
charig: It may be the case that some interleavings with other opt enable to remove more speculative things. Unless I am wrong, or we have another good reason, I would still run these two in each outer iteration.

rir/src/compiler/analysis/scope.cpp
```diff
@@ -162,6 +162,21 @@ ScopeAnalysis::ScopeAnalysis(Closure* function) {
     if (false)
         analysis.print(std::cout);
 
+    // if (function->entry->isExit()) {
```
charig: debugging stuff?

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -38,23 +38,23 @@ void ElideEnvSpec::apply(RirCompiler&, Closure* function) const {
...
+                bool maybeObjR = opRight->type.maybeObj();
                 bool assumeNonObjL =
                     profile->hasTypesFor(opLeft) &&
                     !profile->types.at(opLeft).observedObject();
                 bool assumeNonObjR =
                     profile->hasTypesFor(opRight) &&
                     !profile->types.at(opRight).observedObject();
 
-                if (assumeNonObjL && assumeNonObjR) {
+                if ((!maybeObjL && !maybeObjR) ||
```
charig: Why maybeObj is a condition for not speculating?

rir/src/compiler/pir/instruction.h
```diff
@@ -147,6 +147,8 @@ class Instruction : public Value {
     void replaceUsesIn(Value* val, BB* target);
     bool unused();
 
+    virtual void updateType(){};
```
charig, line 161: Couldn't we just add a generic version that ors the argument values?

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -50,14 +50,15 @@ class StaticAnalysis {
     }
 
   public:
+    Closure* closure;
```
charig, line 70: Do we still need the entry?

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -54,11 +63,8 @@ void ElideEnvSpec::apply(RirCompiler&, Closure* function) const {
                         ip = insertExpectAndAdvance(bb, opLeft, ip);
                     if (maybeObjR)
                         ip = insertExpectAndAdvance(bb, opRight, ip);
-                    i->type.setNotObject();
+                    i->type = i->type.notObject();
```
charig: It seemed clearer before

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -37,8 +37,19 @@ class TheScopeResolution {
                 else if (ldfun)
                     ld = ldfun;
 
-                if (ldfun && ldfun->guessedBinding() &&
-                    ldfun->guessedBinding()->type.isA(PirType::closure())) {
+                if (auto call = CallInstruction::CastCall(i)) {
+                    if (auto cls = call->tryGetCls()) {
+                        // If we know the target of the call and the analysis
+                        // a more precise type of the function we can update
```
charig: the analysis... provides?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -37,8 +37,19 @@ class TheScopeResolution {
...
+                        // callsite.
+                        if (analysis.funTypes.count(cls) &&
+                            !i->type.isA(analysis.funTypes.at(cls))) {
```
charig: The test costs the same as the update. I'd just update to simplify

rir/src/compiler/pir/instruction.h
```diff
@@ -547,7 +547,7 @@ class FLIE(LdVar, 1, Effect::None, EnvAccess::Read) {
 class FLI(ForSeqSize, 1, Effect::Error, EnvAccess::None) {
   public:
     explicit ForSeqSize(Value* val)
-        : FixedLenInstruction(PirType(RType::integer).scalar(),
+        : FixedLenInstruction(PirType(RType::integer).scalar().notObject(),
```
charig, line 555: Shouldn't notObject() be part of scalar()?

rir/src/compiler/pir/type.h
```diff
@@ -142,6 +144,12 @@ struct PirType {
         return *this;
     }
 
+    RIR_INLINE bool merge(const PirType& other) {
```
charig, line 157: Are we using the RIR_INLINE annotation? I'd vote for removing

rir/src/compiler/pir/instruction.cpp
```diff
@@ -81,6 +81,21 @@ void Instruction::print(std::ostream& out, bool tty) {
     printEnv(out, tty);
 }
 
+void Phi::removeInputs(const std::unordered_set<BB*>& del) {
+    auto ii = input.begin();
```
charig: bb

rir/src/compiler/pir/instruction.cpp
```diff
@@ -81,6 +81,21 @@ void Instruction::print(std::ostream& out, bool tty) {
...
+    auto ai = args_.begin();
```
charig: arg

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -11,62 +11,72 @@ namespace rir {
...
 
+    Visitor::run(function->entry, [&](BB* bb) {
         if (!checkpoints.count(bb))
             return;
+        auto cp = checkpoints.at(bb);
 
         auto ip = bb->begin();
         while (ip != bb->end()) {
             Instruction* i = *ip;
+            auto next = ip + 1;
```
charig, line 49: unused?

rir/src/compiler/pir/instruction.h
```diff
@@ -1271,15 +1271,22 @@ class Deopt : public FixedLenInstruction<Tag::Deopt, Deopt, 1, Effect::None,
...
+    bool assumeTrue = true;
+    Assume(Value* test, Value* checkpoint)
         : FixedLenInstruction(PirType::voyd(),
                               {{NativeType::test, NativeType::checkpoint}},
                               {{test, checkpoint}}) {}
 
     Checkpoint* checkpoint() { return Checkpoint::Cast(arg(1).val()); }
+    void checkpoint(Checkpoint* cp) { arg(1).val() = cp; }
     Value* condition() { return arg(0).val(); }
+    Assume* Not() {
```
charig, line 1299: Capital is on purpose?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -291,8 +291,8 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos, rir::Code* s
                 [&](Closure* f) {
                     Value* expected = insert(new LdConst(monomorphic));
                     Value* t = insert(new Identical(callee, expected));
-
-                    insert.conditionalDeopt(t, srcCode, pos, stack, true);
+                    auto cp = insert.addCheckpoint(srcCode, pos, stack);
```
charig, line 291: In the original version, at the rir level we insert only the framestate because it is the only thing that could not be inserted at the pir level. Later, PIR decides what to do. If it needs to speculate it adds the checkpoints, etc. Personally, I prefer much more that design. The optimizer makes all the decision about how to model speculation based on the only needed information from RIR. I agree that we have to add a couple of phases at the pir level later moving things here and there in the CFG, but still, from my POV is a much more clear model.

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -135,8 +157,20 @@ class StaticAnalysis {
                 assert(mergepoint[id].size() > 0);
                 AbstractState state = mergepoint[id][segment];
 
+                if (debug >= DebugLevel::BB) {
```
charig: I'd subclassify the logger and move all this conditions because it really complicates the readability of the real code which already has its complexity

rir/src/compiler/analysis/scope.h
```diff
@@ -29,7 +29,7 @@ class ScopeAnalysis {
         return !allStoresObserved.count(i->env()) && !observedStores.count(i);
     }
     AbstractREnvironmentHierarchy finalState;
-    explicit ScopeAnalysis(Closure* fun);
+    ScopeAnalysis(Closure* fun, LogStream& log);
```
charig, line 33: +1 that we can now remove that weird explicit

rir/src/compiler/debugging/stream_logger.h
```diff
@@ -63,6 +63,21 @@ class LogStream {
         printedAnything = false;
     }
 
+    template <typename T>
+    friend LogStream& operator<<(LogStream& log, T dump) {
```
charig, line 68: Why a template and not just a string?

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -246,7 +246,7 @@ class TheCleanup {
 namespace rir {
 namespace pir {
 
-void Cleanup::apply(RirCompiler&, Closure* function) const {
+void Cleanup::apply(RirCompiler&, Closure* function, LogStream&) const {
```
charig, line 262: Now that the compiler is a parameter, maybe we can have an analysis logger as part of the compiler and simplify this api

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -148,7 +146,7 @@ class ForceDominanceAnalysisResult {
             if (deopt)
                 return false;
         }
-        return exit.count(a) && exit.at(a) != ForcedBy::ambiguous();
+        return true;
```
charig: I do not get what was the bug? There is also no test to help me. Besides, apparently the change open more opportunities for inlining promises, why that would kill opts?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -43,41 +43,58 @@ using namespace rir::pir;
...
+                    changed = true;
+                }
             }
         }
-        for (auto& e : other) {
-            if (!count(e.first)) {
-                (*this)[e.first] = ambiguous();
+        for (auto& e : other.forcedBy) {
+            if (!forcedBy.count(e.first) && declared.count(e.first)) {
+                forcedBy[e.first] = ambiguous();
```
charig: I do not fully get why it is ambigous if one state defines the arg and the other force it.

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -46,22 +46,29 @@ using namespace rir::pir;
...
+    std::unordered_set<Value*> escaped;
 
     static Force* ambiguous() {
         static Force f(Nil::instance(), Env::nil());
         return &f;
     }
 
     void declare(MkArg* arg) { declared.insert(arg); }
+
+    void sideeffect() {
```
charig: forceEscaped

rir/src/compiler/pir/instruction.h
```diff
@@ -87,6 +87,7 @@ enum class Effect : uint8_t {
     Error,
     Print,
     Write,
+    Force,
```
charig, line 95: Ok, but I do not see the force effect in any instruction

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -236,33 +236,52 @@ struct AbstractLoad {
...
-                } else if (at(i).merge(other.at(i))) {
+                } else if (envs.at(i).merge(other.envs.at(i))) {
                     changed = true;
                 }
             } else {
-                (*this)[i].taint();
+                envs[i].taint();
                 changed = true;
             }
+        for (auto& a : other.aliases) {
```
charig: a -> alias

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -66,18 +66,21 @@ class StaticAnalysis {
     const DebugLevel debug;
 
     Closure* closure;
+    Code* code;
     BB* entry;
 
-    StaticAnalysis(std::string name, Closure* cls, LogStream& log,
+    StaticAnalysis(std::string name, Closure* cls, Code* code, LogStream& log,
```
charig: We do not need the code. I'd prefer a flag or the promise id

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -165,20 +171,26 @@ struct ForcedBy {
 
 class ForceDominanceAnalysis : public StaticAnalysis<ForcedBy> {
   public:
-    explicit ForceDominanceAnalysis(Closure* cls, LogStream& log,
+    explicit ForceDominanceAnalysis(Closure* cls, Code* code, LogStream& log,
                                     DebugLevel debug = DebugLevel::None)
-        : StaticAnalysis("ForceDominance", cls, log, debug) {}
+        : StaticAnalysis("ForceDominance", cls, code, log, debug) {}
 
     void apply(ForcedBy& d, Instruction* i) const override {
```
charig: d -> state

rir/src/compiler/analysis/scope.cpp
```diff
@@ -264,8 +287,10 @@ void TheScopeAnalysis::apply(ScopeAnalysisState& state, Instruction* i) const {
             if (!CallSafeBuiltin::Cast(i))
                 state.mayUseReflection = true;
         }
-        if (!handled)
+        if (!handled) {
+            effect.max(AbstractResult::Tainted);
```
charig: At least for consistency I'd put the effect after that state merge

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -53,61 +53,82 @@ struct ForcedBy {
...
+        if (!escaped.count(val)) {
+            escaped.insert(val);
+            return true;
+        }
+        return false;
     }
 
-    void escape(Value* val) { escaped.insert(val); }
+    AbstractResult merge(const ForcedBy& other) {
+        AbstractResult res;
```
charig, line 94: I do not get where it is initialized

rir/src/ir/Compiler.cpp
```diff
@@ -748,25 +748,24 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
         }
     }
 
+    // SEXP builtin = fun->u.symsxp.value;
```
charig, line 756: ???

rir/src/compiler/analysis/scope.cpp
```diff
@@ -88,29 +88,29 @@ struct ScopeAnalysisState {
...
     Value* promiseEnv = nullptr;
 
-    TheScopeAnalysis(Closure* cls, const std::vector<SEXP>& args,
-                     LogStream& log, DebugLevel debug = DebugLevel::None)
-        : Super("Scope", cls, cls, log, debug), args(args), depth(0) {}
-    TheScopeAnalysis(Closure* cls, const std::vector<SEXP>& args,
-                     Value* staticClosureEnv,
+    TheScopeAnalysis(Closure* cls, const std::vector<SEXP>& argNames,
+                     LogStream& log)
+        : StaticAnalysis("Scope", cls, cls, log), argNames(argNames), depth(0) {
```
charig, line 99: We do not initialize the args?

rir/src/compiler/opt/delay_env.cpp
```diff
@@ -71,31 +70,26 @@ void DelayEnv::apply(RirCompiler&, Closure* function, LogStream&) const {
...
             auto moveMkEnvToDeoptBranch = [&](BB* deoptBranch,
                                               BB* fastPathBranch) {
                 auto newEnvInstr = envInstr->clone();
-                it = bb->insert(it, newEnvInstr);
-                envInstr->replaceUsesIn(newEnvInstr, fastPathBranch);
-                // Closure wrapper in MkEnv can be circular
-                Replace::usesOfValue(newEnvInstr, envInstr, newEnvInstr);
+                deoptBranch->insert(deoptBranch->begin(), newEnvInstr);
+                envInstr->replaceUsesIn(newEnvInstr, deoptBranch);
                 it = bb->moveToBegin(it, fastPathBranch);
```
charig, line 84: The cleanup will do it, but maybe it is worth to remove the old env instruction here. Otherwise, change the name from moveMkEnv to copyMkEnv or sth like that

rir/src/compiler/analysis/scope.cpp
```diff
@@ -146,17 +146,27 @@ void TheScopeAnalysis::drillDown(const ScopeAnalysisState& s, Value* v,
...
+            auto argi = Instruction::Cast(args[lda->id]);
+            if (argi) {
+                auto arg = AbstractPirValue(argi, argi);
+                return drillDown(s, argi, found, [&]() { found(arg); });
+            }
+        }
+    }
+
     auto i = Instruction::Cast(v);
     if (s.returnValues.count(i)) {
```
charig: s -> state; i -> instruction

rir/src/compiler/analysis/scope.cpp
```diff
@@ -146,17 +146,27 @@ void TheScopeAnalysis::drillDown(const ScopeAnalysisState& s, Value* v,
     // if for a particular value we have already a result available from the
     // above.
 
-    bool done = false;
+    if (auto lda = LdArg::Cast(v)) {
```
charig, line 146: The function comment should be before the definition

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -78,10 +80,23 @@ void Constantfold::apply(RirCompiler& cmp, Closure* function,
...
             FOLD(Neq, symbol::Ne);
             FOLD(Pow, symbol::Pow);
+
+            if (auto assume = Assume::Cast(i)) {
+                auto condition = assume->arg<0>().val();
+                if (auto isObj = IsObject::Cast(condition))
+                    if (!isObj->arg<0>().val()->type.maybeObj())
+                        next = bb->remove(ip);
+
+                FOLD2(condition, Identical,
```
charig, line 87: I do not get this FOLD2 here. 
How do we know the condition is the current ip? Why don't we rely on dead code elimination that should be more reliable? 
Why don't we just have a FOLD(Identical)?

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -106,6 +121,10 @@ void Constantfold::apply(RirCompiler& cmp, Closure* function,
                     branchRemoval.emplace(bb, a == b);
                 });
             }
+
+            if (auto isObj = IsObject::Cast(condition))
```
charig, line 128: I think we are removing branch's bbs, but not the branch execution and its condition which we should. Am I wrong?

rir/src/compiler/pir/instruction.h
```diff
@@ -1225,7 +1225,7 @@ class VLI(Phi, Effect::None, EnvAccess::None) {
     void pushArg(Value* a) override { assert(false && "use addInput"); }
     void addInput(BB* in, Value* arg) {
         input.push_back(in);
-        args_.push_back(InstrArg(arg, arg->type));
+        args_.push_back(InstrArg(arg, PirType::any()));
```
charig, line 1242: Do we have a test failing on this?

rir/src/compiler/pir/instruction.h
```diff
@@ -58,15 +58,20 @@ class BB;
 class Closure;
 class Phi;
 
-struct InstrArg : public std::pair<Value*, PirType> {
-    InstrArg(Value* v, PirType t) : std::pair<Value*, PirType>(v, t) {
+struct InstrArg {
+  private:
+    PirType t;
```
charig: type; value

rir/src/compiler/pir/type.cpp
```diff
@@ -8,8 +8,10 @@ namespace pir {
 
 void PirType::print(std::ostream& out) { out << *this << "\n"; }
 
-PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
-    switch (TYPEOF(e)) {
+void PirType::mergeSexptype(SEXPTYPE sexptype) {
```
charig: mergeSexptype -> mergeType | addType

rir/src/compiler/pir/type.cpp
```diff
@@ -81,5 +87,28 @@ PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
...
+    if (other.numTypes == 0 || other.numTypes == ObservedValues::MaxTypes)
+        return;
+
+    bool noObj = true;
+    bool allScalar = true;
+
+    for (size_t i = 0; i < other.numTypes; ++i) {
+        const auto& record = other.seen[i];
+        if (record.object)
+            noObj = false;
```
charig: Why not setting the TypeFlags here, for instance, if other is obj and current no?

rir/src/compiler/pir/value.h
```diff
@@ -22,6 +22,7 @@ class BB;
 class Value {
   public:
     PirType type;
+    PirType typeFeedback;
```
charig: I am not sure that adding a type to every value is better than just having the profiling as an independent structure. I agree that inlining requires some mappings when merging the values because we copy the instructions but it should not be so complicated

rir/src/compiler/opt/elide_env.cpp
```diff
@@ -15,10 +15,27 @@ void ElideEnv::apply(RirCompiler&, Closure* function, LogStream&) const {
...
+            if (i->envOnlyForObj()) {
+                needed = false;
+                i->eachArg([&](Value* v) {
+                    if (!needed && v != i->env() && v->type.maybeObj()) {
+                        needed = true;
+                    }
+                });
+                if (!needed) {
+                    i->elideEnv();
+                    i->type = i->type.notObject();
```
charig: Why we set this flag to the instruction?

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -246,25 +246,12 @@ class AbstractREnvironmentHierarchy {
...
-                        res.taint();
-                    }
-                } else {
-                    res.max(envs.at(i).merge(other.envs.at(i)));
-                }
-            } else {
-                envs[i].taint();
-                res.taint();
-            }
+            if (envs.count(e.first))
```
charig, line 250: Why not? Because it is based on a forward analysis?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -237,21 +237,21 @@ AbstractResult TheScopeAnalysis::apply(ScopeAnalysisState& state,
...
         if (!handled) {
             state.returnValues[i].merge(AbstractPirValue::tainted());
-            effect.max(AbstractResult::Tainted);
+            effect.taint();
         }
     } else if (CallInstruction::CastCall(i) && depth < maxDepth) {
         auto calli = CallInstruction::CastCall(i);
         if (auto call = Call::Cast(i)) {
-            auto trg = drillDown(call->cls()->followCastsAndForce(), i).val;
+            auto trg = call->cls()->followCastsAndForce();
```
charig: trg?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -119,6 +119,10 @@ class TheScopeAnalysis : public StaticAnalysis<ScopeAnalysisState> {
     typedef std::function<void(AbstractLoad)> LoadMaybe;
     RIR_INLINE void tryLoad(const ScopeAnalysisState& envs, Value* i,
                             LoadMaybe) const;
+    typedef std::function<void(const AbstractPirValue&)> AbstractValMaybe;
+    typedef std::function<void()> Maybe;
+    inline void drillDown(const ScopeAnalysisState& envs, Value* i,
```
charig: why inline?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -135,42 +139,50 @@ void TheScopeAnalysis::tryLoad(const ScopeAnalysisState& s, Value* i,
...
+    tryLoad(s, v, [&](AbstractLoad l) {
+        done = true;
+        auto& res = l.result;
+        if (res.isSingleValue()) {
+            return drillDown(s, res.singleValue().val, found,
+                             [&]() { found(res); });
+        }
+        found(res);
+    });
+    if (!done)
```
charig, line 177: tryLoad would always set done to true?

charig (approved):
A couple of meta comments about my review:

1. It is a huge PR with lot of important code so I would not complain if we just merge it as is. But...

2. Some methods are full of one letter names like s(state), i(instruction/index), v(value), a (alias). Some of these methods model really complex algorithms and the manual conversion of the letters to the whole word makes them more complex to understand/review. If possible I would try to go through, at least the ones in the more complex methods, before merging.  

3.  I didn't go through the last commit: `slightly more fine grained named controls`. The reason is that I have no idea about the setShared and namedVar construction. Where can I find information about these semantics?

4. I had serious problems to follow the scope analysis changes. Maybe it was just because I did the review on the web and never looked at the whole code. Also I couldn't use too much time since the PR is huge. But I had the impression that this phase is getting quite complex and already has lot of assumptions encoded. 
From my POV it should be refactor. For instance, it is definitely unclear for me what the `drillDown` method does, even if it has a comment. I am absolutely sure the name of the method is wrong but my point is beyond naming.

rir/src/compiler/opt/cleanup_checkpoints.cpp
o-: the compiler invokes every pass with the closure object. if the pass looks just at the body or also promises is up to the pass.

rir/src/compiler/opt/cleanup_checkpoints.cpp
o-: I split them into 3 passes since I want to be able to run them independently. Cleanup is run a couple of times, the other two are only run once. In general, at some point we remove checkpoints and then run the optimizations again without them, since that might give us more opportunities to optimize.

rir/src/compiler/opt/cleanup_framestate.cpp
o-: any instruction which has no effect and is unused is always removed by the normal cleanup phase. So there is no point in having a dedicated framestate cleanup phase. The only thing that needs to be done is remove framestates from calls, since there they are just metadata and not actual input to the instruction. Thus they generate sort of a "fake" dependency.

The reason we actually did explicitly remove fs in the past was that it was marked as effectful (which makes no sense for a metadata only instruction). after fixing that, fs gets removed automatically.

rir/src/compiler/opt/scope_resolution.cpp
o-: I don't understand this comment. In this case ld is not used, but its used eg on line 94?

o-: I don't understand this comment. In this case ld is not used, but its used eg on line 94?

rir/src/compiler/pir/instruction.h
o-: > Guessed sounds to me like sth that we are not 100% sure.

which is exactly what this is.
for example in:
```
f <- function() 1
(function(f) f())(1)
```
the function argument is the guessed binding, but the f from the first line is the actual binding.

o-: so the whole point is that we want to remember what we guessed, even when we are not 100% sure

o-: I am not sure about a better name. Maybe `innermostBinding`?

rir/src/compiler/pir/instruction.h
o-: I guess you mean member (ie. one variable per object and not per class).

any `pir::Value*` *must* be in the arguments. The whole compiler relies on the fact that we can iterate over all arguments. Otherwise e.g. the cleanup pass will not see the use and just delete the value.

rir/src/utils/configurations.cpp
o-: the reasons are: the testcases still work :)

rir/src/compiler/opt/elide_env_spec.cpp
o-: not maybe object == it can't be an object, so why speculate, if you know?

o-: I think you got it wrong. it's not a condition for speculating. if we know that both are not objects we just elide right away. but I made it more readable

rir/src/compiler/pir/instruction.h
o-: no.

what about `a[b]`, here the result type is the scalar version of the a type, for `a:b` the result type is a sequence of the union type, and so on. This is highly specific on the actual instruction.

rir/src/compiler/analysis/generic_static_analysis.h
o-: yes, it seems so

rir/src/compiler/pir/instruction.h
o-: they are orthogonal. scalar refers to the size of the vector.
`structure(1, class="foo")` is a scalar S3 object...

rir/src/compiler/pir/type.h
o-: its used to force gcc/clang to inline this function

rir/src/compiler/opt/elide_env_spec.cpp
o-: why?

o-: why do you think so?

rir/src/compiler/pir/instruction.h
o-: `not` is a keyword

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: well, that is unfortunately not true. If you look at the old `insert.conditionalDeopt` it did in fact insert a branch with a deopt. I guess we never cared to update it to use our new checkpoint infrastructure.

I really tried to fit this case into the split design but it was cumbersome and would have required some major rehaul. On the other hand inserting checkpoints early is quite natural and was easy to do.

I guess I just don't see an advantage. of having an extra pass, right after the rir2pir, that inserts checkpoints. Also the whole insert checkpoints pass was really brittle and needed some "interaction at a distance" between 3 passes (rir2pir, insertCheckpoint, elideEnv) to work.

Thinking about it, I feel like the checkpoints should not be inserted by any pass at all. But rather we should insert them on demand. basically at the point where we want to add an assumption, we either use the nearest checkpoint of if that is not possible we insert a new one. that would make a lot more sense.

o-: currently we add checkpoints that we might need in the future and then we remove again those that we did not need. That's kind of stupid, when we could just add them, when we need... I might look into that later

rir/src/compiler/analysis/generic_static_analysis.h
o-: I would like to keep it this way. Because I would like to keep it zero overhead when not activated.

rir/src/compiler/analysis/scope.h
o-: `explicit` is used when the constructor has one argument. Consider
```
class A {};
class B {
  B(A& a) {}
};
```
note `B` has a constructor that takes an `A` as argument. C++ has this "great feature" that you can write:

```
A a;
B b = a;
```

and it will implicitly use the `B(A& a)` constructor to create a new `b` from the `a` rhs. This is often a source of confusion and errors. That's why constructors with one argument should generally be marked explicit. This causes the above snippet to fail at compile time. And you have to write instead explicitly:

```
A a;
B b = B(a);
```

rir/src/compiler/debugging/stream_logger.h
o-: because it's potentially not just used for strings. Think about stuff like `log << 1`, or basically anything that supports an `<<` operator for streams.

rir/src/compiler/opt/cleanup.cpp
o-: I don't see how this would simplify anything here. The analysis logger would then have to find the correct log stream (ie. the one that belongs to the function). Remember that this was one of the main issues with the stream logger manager thingy.

rir/src/compiler/opt/force_dominance.cpp
o-: ah, sorry, i fucked up the rebasing. The actual bugfix is in the next commit :)

It's this one:

https://github.com/reactorlabs/rir/pull/267/commits/0e169f9ff1b4743aba214c36b2b8167dd40c0a51#diff-2ba178b0e828f9905c1be7cbf0c32e49R124

basically the missing `baseValue`. This caused the analysis to discover promises, but then miss some of their uses.

o-: sorry for the confusion

rir/src/compiler/opt/force_dominance.cpp
o-: declared should be called inScope.

so if state A has forced a promise and state B not, then either it's a backjump and B has not even seen the promise (in this case no ambiguity), or both A and B have the promise in scope but one forced it, the other not (in this case it's really ambiguous).

rir/src/compiler/opt/force_dominance.cpp
o-: it seems to me that sideeffect is more declarative. forceEscaped explains what it does, but not really why

rir/src/compiler/pir/instruction.h
o-: reserved for future use. it's a lattice, so Any contains Force

rir/src/compiler/analysis/abstract_value.h
o-: I wrote entry now. We are not iterating over aliases, but over alias pairs..

rir/src/compiler/analysis/generic_static_analysis.h
o-: and what do you do in case of the body? this is supposed to work for promises and body.

rir/src/compiler/opt/force_dominance.cpp
o-: objects are initialized by default.  The empty constructor
`AbstractResult() : kind(None) {}` is called.

rir/src/ir/Compiler.cpp
o-: hah, forgot about this. I disabled some optimizations at the RIR level. Here we would just hardwire function calls to any builtin function in the global space. For example you could not define a function called `add`.

rir/src/compiler/analysis/scope.cpp
o-: again it is initialized to the empty list. The default constructor of a std::vector is the empty vector. See https://en.cppreference.com/w/cpp/language/data_members to see how exactly it works in C++.

rir/src/compiler/opt/delay_env.cpp
o-: no, both of them are still needed. We duplicate the env, one version goes into the deopt branch and the other one in the other branch. Maybe the non deopt one is not used anymore. But we don't know at this point.

o-: will update comment and naming.

rir/src/compiler/opt/constantfold.cpp
o-: ultimatively the issue is that we do not have a native true/false value in pir atm. So native booleans are only used in branches, so instead of first constant folding tests to true/false and then remove branches on true/false, i do both at the same time.

o-: I am gonna put a todo there...

rir/src/compiler/opt/constantfold.cpp
o-: this is correct, because asTest and asLogical are both effectful. asLogical prints a warning if it is not scalar and asTest throws an error if it's NA. That's just how R works. So e.g. it could be that we remove the branch, but still need to show a warning that the branch condition was larger than 1.

rir/src/compiler/pir/instruction.h
o-: yes, through more optimizations being unlocked this started failing in the gnur tests

rir/src/compiler/pir/type.cpp
o-: that's what I do below.

rir/src/compiler/pir/value.h
o-: it was complicated, i tried. that's against this solution?

rir/src/compiler/opt/elide_env.cpp
o-: if the inputs are not objects the results cannot be object.

rir/src/compiler/analysis/abstract_value.h
o-: envs are ssa values, they cannot be ambiguous...

rir/src/compiler/analysis/scope.cpp
o-: no, try load only executes the lambda, if it manages to figure out what is being loaded.

rir/src/compiler/pir/instruction.h
mhyee: FWIW I'm fine with `guessedBinding`, but more comments could help.

o-:
> Where can I find information about these semantics?

the R interpreter uses number, called the named count (something like a reference count), to keep track of how many times a variable is used.

0 means, it's a temporary variable, not used anywhere else. we can just reuse the memory when we don't need it anymore. For example if we do a `+` and say the lhs has a named count of 0, we can just store the result there.

1 means it has only one binding. e.g. if I do `a[1] <- x` I know that i can update the array `a` inplace, if it has a named count of 1.

above 1 means this object is used in several places. due to the pass-by-value semantics of R I need to copy this SEXP if I want to change it.

in RIR/PIR we have two instructions to manipulate named count: `setShared` ensures that named count is at least 2. `ensureNamed` ensures that it's at least 1.

in the future we should have a special pass that computes a static named count and uses that to keep the named count as low as possible, to avoid duplicating memory if  possible.

rir/src/compiler/opt/cleanup_checkpoints.cpp
charig: It is still not very clear the advantages of running these passes independently. Anyway, my main point was that cleanup checkpoint and cleanup framestates could be part of the same pass.

rir/src/compiler/opt/scope_resolution.cpp
charig: The comment was because ld is defined in 31 and I thought ld was being redefined in 35-38 but I missed that line 37 is an `else if` instead of an `if`, sorry.

rir/src/compiler/pir/instruction.h
charig: But the problem I have with the naming here is that I see this as a pure static analysis. I agree the binding may be a temporal value we are not a 100% sure, but after the whole analysis passes if it has a binding we are sure. If not we set the thombstone. It is not sth we are checking at runtime and deoptimizing afterwards, which is what guessed resembles me to. So, maybe I am missing sth. If that is the case, I am sorry and just ignore this comment.

rir/src/utils/configurations.cpp
charig: Indeed. But what about the benchmarks? 
Maybe adding/removing a pass here or there is worth for one benchmark but jeopardizes another one. 
So, maybe start documenting the rationale/need for the change on the number of passes gives us more information for eventual future changes.

rir/src/compiler/opt/elide_env_spec.cpp
charig: It is set to ip + 1 here and to the same value inside the if...

rir/src/compiler/analysis/generic_static_analysis.h
charig: If it is just for the overhead, the define for adding the code or not could be inside the logging functions. I guess the C compiler will not call an empty function...

rir/src/compiler/opt/cleanup.cpp
charig: Yes, this is absolutely not important. Just a matter of preferences I guess. 
I personally do not like to add parameters when they are not absolutely needed. I prefer the simple api `Pass:apply(compiler, function)` which resembles my loving smalltalk `Pass>>applyTo: function using: aCompiler`. Then inside each pass I'd prefer to add 1 line like: `log := compiler.logStreamFor(this)`.

charig (approved):

rir/src/compiler/pir/instruction.h
o-: > the binding may be a temporal value we are not a 100% sure, but after the whole analysis passes if it has a binding we are sure

unfortunately not. Let me give you an annotated example:

```
test <- function(f) f()
test(function()1)
```
so in a kind of PIR syntax this is:
```
test:
  %t1 = ldarg 1
  %t2 = mkenv "f"=%t1
  %t3 = ldfun "f", %t2
        call %t3

main:
  %m1 = mkarg(p1)
  %m2 = ldfun "test"
        call %m2(%m1)
 prom1:
  return mkClosure(...)
```

ok, so first step is of course to inline test into main, so we get:

```
main:
  %1 = mkarg(p1)
  %2 = mkenv "f"=%1
  %3 = ldfun "f", %2
        call %3
 prom1:
  return mkClosure(...)
```

so now we would like to inline the promise `p1`, but how? Our promise inliner pass matches up `mkArg <--> force` pairs and inlines the promise at the location of the force instruction (because that's where it get's actually executed). But as you can see here, there just isn't any force instruction, because of the weird semantics of lookup, the "f" is forced as a sideeffect of `%t3 = ldfun "f", %2`.

How to remedy?

*Step1:* get the desired force instruction

The scope resolution pass creates the following:

```
main:
  %1   = mkarg(p1)
  %2   = mkenv "f"=%1
  %3.1 = force %1
  %3   = ldfun "f", %2, guess = %3.1
          call %3
 prom1:
  return mkClosure(...)
```

(Note, here I simplified to much. Actually the return type of 'prom1' is trivially 'closure', so we would not need to guess at all. But imagine it is just `return (ldvar x)`, we might have no clue at all, what the guess is.)

*Step2:* promise inliner

Voila, we have a force (`%t3.1`) and a mkarg (`%1`) that match up. We can inline the promise code:

```
main:
  %1   = mkarg(p1)
  %2   = mkenv "f"=%1
  %3.1 = mkClosure(...)
  %3   = ldfun "f", %2, guess = %3.1
          call %3
```

*Step3:* promote the guess

Well, now it's easy. `%3.1` has type closure, and so we know that the guess was right. We will replace `%3` by `%3.1` and remove `%3`.

But, there are two other possible outcomes:

1. The guess is wrong:

after promise inlining we might end up with:

 ```
main:
  %1   = mkarg(p1)
  %2   = mkenv "f"=%1
  %3.1 = ldConst 1
  %3   = ldfun "f", %2, guess = %3.1
          call %3
```

so we know the guess was wrong since the innermost binding is just the number 1. This happens if I would call test like this: `f <- function()1; test(1)`.

What we could do in this case (we don't yet) is to replace the guess with the "next best guess" (ie. the next binding).

2. We can't show the guess to be true nor false

after promise inlining we might end up with:

 ```
main:
  %1   = mkarg(p1)
  %2   = mkenv "f"=%1
  %3.1 = call("mystery")
  %3   = ldfun "f", %2, guess = %3.1
          call %3
```

The first binding is something we have no clue what it is. We can neither exclude it, since mystery might return a closure, nor show it will be skipped, since mystery might not return a closure.

In this case the backend of the compiler just forgets the whole guess business. When we lower to rir, the guess is dropped on the floor.

rir/src/compiler/opt/elide_env_spec.cpp
o-: But `ip` has a different value inside the if... Granted here we could just use `ip++` at the end. But I had so many subtle bugs with iterators, that I started using the same robust pattern everywhere:

```
while (ip != end) {
  auto next = ip+1;
  ...
  ip = next;
}
```

You are basically suggesting to use:

```
while (ip != end) {
  ...
  ip++;
}
```

Let me show you how this breaks:

At some point I decide to add a new case to this loop. Namely:
```
...
  ip = bb->remove(ip);
  ip++;
```

so now, what happens if I remove the last element?
the `ip` points to the previously second-last after removal. But since I don't want to process this instruction twice i add the `ip++`, so now `ip` points to the end.

Then at the end of the loop the `ip++` actually increments over the end, the loop condition is true and we get a segv (maybe).

o-: The same thin does not happen with the above pattern. I would add the case like this:
```
  ip = bb->remove(ip);
  next = ip+1
```

and since we are not incrementing next anymore, this will do the right thing

o-:
> I prefer the simple api Pass:apply(compiler, function)

Imho this is confusing simple with easy.

Your api adds a tight coupling between compiler and pass: the compiler relies on a certain api for the pass (apply) and the pass relies on a certain api of the compiler (log).

My api has the requirement only in one direction. The pass does not need to know about the compiler. Thus I would state that it is strictly less complex than the alternative, even if it is a bit more verbose.

rir/src/compiler/pir/instruction.h
charig: Thanks a lot for the explanation!

Although I stated it wrong, your explanation reinforces somehow what I wanted to mean.
The name guess resorts me to a dynamic setting, in which I would execute code, and if the guess was wrong I must act accordingly.
However, this behavior happens within the compiler statically. The argument of the `ldFun` is just a static binding of the name at the point where this instruction would be executed. IMO that is much more clear for somebody reading the code. Then in the lowering phase, you work with the semantics of the ldFun: if we can not ensure the binding is a closure we can not change the `ldFun` for a simple call. 

Summarizing: IMO the naming is misleading. Since it encapsulates a lot of assumptions about why it is actually there for, I thought it was important to discuss. But you and @mhyee think it is the correct one so leave the current naming.

rir/src/compiler/opt/elide_env_spec.cpp
charig: I do not see why in:
```
...
  ip = bb->remove(ip);
  ip++; 
```
you do `ip++` if you know that the end of the loop will do that.

Anyway, ok if we have a new pattern for iterations. I didn't know about it at the moment of the comment.

PR #268, opened 2018-11-19
Add backward visitors

JanJecmen:


rir/src/compiler/util/visitor.h
```diff
@@ -88,13 +112,26 @@ class VisitorImplementation {
...
+                      "Backward visitors need a CFG.");
         run(bb, [action](BB* bb) {
             for (auto i : *bb)
                 action(i);
         });
     }
 
+    static void run(CFG const& cfg, BB* bb, InstrAction action) {
+        static_assert(DIR == Direction::Backward,
+                      "Only backward visitors can take a CFG.");
```
o-: would it maybe make sense to use a subclass for the backwards visitor? Then you can add the cfg to the constructor.

JanJecmen: I was thinking about it, but then I would have to duplicate the `genericRun` so this seemed a bit less ugly... In general, I couldn't find a good way to add the cfg only to a subclass but use it in the superclass (tried crtp too, but everything is static, so that didn't quite work)

JanJecmen: Hmm maybe I could factor out the `scheduleNext` and then the subclass would work I think.. I will give it another go

o-: ah, right, I see

o-: i guess it's fine like that in that case. don't worry

o-:
sorry for stalling here. I think the algorithm is fine. I don't like the api. Ie. it's not very intuitive that the fwd and bwd have the same function names and you pick them by passing a parameter more or less. that looks fishy.

it seems to me that you could factor out the `schedule` function (or lambda) and then share most of the code, maybe?

also, maybe it makes sense to remove all the cruft I added to share the code between `check` and `run`, that seems to complicate things and maybe the bit of code duplication is actually better than all that template trickery.

o-:
anyway, I leave it at your discretion, don't overthink it, we can still iterate on it and maybe accept some code duplication.... Also, if it helps I would make all the functions non-static. You can still have a function api, eg. something like:

```
static void visit(BB* bb, function fun) { Visitor v; v.run(bb, fun); };
static void visitBwd(BB* bb, function fun, CFG& cfg) { BwdVisitor v(cfg); v.run(bb, fun); };
```

JanJecmen:
Another iteration:
- Now the direction is not a template parameter
- We have the api as suggested, `run` vs. `runBackward` to choose the direction
- The `sheduleNext` lambda is factored out and the forward and backward functions use different ones
- Downside is that we pass around a lambda that takes another lambda.. not very readable :(

PR #269, opened 2018-11-22
move loop context setjmp/longjmp out of the interpreter loop

o-:
I had a theory that the setjmp/longjmp in the middle of the loop body deteriorates perf. So I experimented with moving it out into a loopTrampoline function. With this patch we are now using the same strategy as mainline R -- executing contextful loop bodies in their own function. Since most loops have no context it seems better to tank their performance.

Anyway, the theory was wrong, the patch does not improve much. Still putting it here, maybe we decide it's more readable?

### DIFF
--- a/rir/src/R/r.h
+++ b/rir/src/R/r.h
@@ -1,6 +1,8 @@
 #ifndef RIR_R_H
 #define RIR_R_H
 
+#include "common.h"
+
 #include <R.h>
 #define USE_RINTERNALS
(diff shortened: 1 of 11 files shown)


o-:
actually I just pushed another small change (since I worked on that on top of this PR) and at least together they seem to speed up the interpreter a tiny bit

o-:
ok, merging this tbr. not gonna rebase it yet another time

Issue #250, opened 2018-10-30
Factor out some of the duplicate code for handling bytecodes


mhyee:
When adding a new bytecode (e.g. in #176 and #247), there are a lot of files that need to be changes. Some of these changes are non-trivial, but others are trivial (especially for RIR bytecodes).

There may be some opportunity for factoring out the common code. Or maybe there are too many non-trivial cases.

Or maybe there's a way to define all the RIR bytecodes in one file and all the PIR bytecodes in another file, and get most of the code generated. (Caveat is that generating this code can make things more complicated. Right now, adding a new bytecode is pretty rare, and we get helpful error messages if we forget to update a file.)

This task is a bit more open-ended: investigate what parts can be factored out, and propose a refactoring or pull request for it.

mhyee:
Closed by #270.

PR #270, opened 2018-11-25
Simple instructions (#250)

### LINKED ISSUE #249: Deduplicate Symbols.{h,cpp}
Currently [Symbols.h](https://github.com/reactorlabs/rir/blob/master/rir/src/R/Symbols.h) and [Symbols.cpp](https://github.com/reactorlabs/rir/blob/master/rir/src/R/Symbols.cpp) are very similar, and every time we add a new symbol (admittedly, not a common occurrence), we have to edit both files.

We should fix this so we only need to modify one file.

See how [insns.h](https://github.com/reactorlabs/rir/blob/master/rir/src/ir/insns.h) does it with a C macro trick/idiom.

Jakobeha:
This PR allows you to add a new "simple" instruction easier. A "simple" instruction compiles to a single assembly instruction handled in interp.cpp - currently the only "simple" instructions are int3 and printInvocation. Instead of changing multiple files, just declare the instruction in [simple_instruction_list.h](https://github.com/Jakobeha/rir/blob/simple_instr/rir/src/simple_instruction_list.h), and then define it in interp.cpp. simple_instruction_list.h uses "recursive" preprocessor macros like instruction_list.h.

This PR includes my other PR, which adds the printInvocation instruction.

The program should behave almost the same as my previous PR, **except** instead of creating breakpoints in R with `.debug.break()`, you need to write `.int3()`. This makes the naming easier, since ".debug.break()" adds an "int3" instruction. It could be fixed by adding a new field to the definitions in simple_instruction_list.h, or renaming "int3" to "debugBreak".

This PR also fixes #249 with a new file, symbol_list.h. symbol_list.h uses preprocessor macros like instruction_list.h and simple_instruction_list.h, and it's used by Symbols.h and Symbols.cpp.


mhyee:
Thanks for working on this! I'm going to look at it later this week when I have more time, but I have two quick comments for now:

1. There's some issue with the file permissions. If you look at the diff on GitHub, all the files had their permissions changed from 644 to 755, including a lot of unrelated files. Could you fix this?

2. There are some merge conflicts, because the codebase was updated while you were making changes. This means our continuous integration tests can't run, and we wouldn't be able to merge in the changes. Could you please resolve the merge conflicts and update your branch?

mhyee:
Sorry, but it looks like there's still some merge conflicts.

Also, some of the permissions you changed to 644 should actually be 755. This is why the Travis CI build is broken.

rir/src/compiler/pir/instruction_list.h
```diff
@@ -70,7 +74,25 @@
     V(CastType)                                                                \
     V(SetShared)                                                               \
     V(PirCopy)                                                                 \
-    V(Int3)
+    SIMPLE_INSTRUCTIONS(V_SIMPLE_INSTRUCTION, V)                               \
+
+#undef V_SIMPLE_INSTRUCTIONS 
+
+#define BINOP_INSTRUCTIONS(V)                                                  \
```
mhyee: Why was this duplicated?

rir/src/runtime/Function.cpp
```diff
@@ -4,4 +4,5 @@ namespace rir {
 void Function::disassemble(std::ostream& out) {
     body()->disassemble(out);
 }
+
```
mhyee: Unnecessary change.

rir/src/R/Symbols.cpp
```diff
@@ -3,80 +3,9 @@
...
-DECLARE(debugBreak, ".debug.break");
-DECLARE(UseMethod, "UseMethod");
-
-#undef DECLARE
+#define V(name, txt) SEXP name = Rf_install(txt);
+SYMBOLS(V)
+#undef V
 } // namespace symbol
 
 } // namespace rjit
```
mhyee: Since you're already updating this file, could you update the comment to `// namespace rir`?

rir/src/R/Symbols.h
```diff
@@ -2,84 +2,14 @@
...
-DECLARE(debugBreak, ".debug.break");
-DECLARE(UseMethod, "UseMethod");
-
-#undef DECLARE
+#define V(name, txt) extern SEXP name;
+SYMBOLS(V)
+#undef V
 } // namespace symbol
 
 } // namespace rjit
```
mhyee: Same with this comment. `rjit` -> `rir`.

mhyee (changes requested):
Thanks for making the changes, and thanks for your patience.

I left a few small comments in the PR, but I also have two larger comments here.

First, I'm not sure why the build is failing. [cppcheck is complaining](https://travis-ci.org/reactorlabs/rir/jobs/460463334) about some code style, for [a file that you didn't change](https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/pir/tag.cpp). (@o-, any thoughts?)

We'll probably have to add a suppression (refer to the [cppcheck manual, chapter 9](http://cppcheck.sourceforge.net/manual.pdf)). For RIR, we've been using a combination of inline suppressions and command-line suppressions in our [cppcheck script](https://github.com/reactorlabs/rir/blob/master/tools/cppcheck), so pick whichever you think is more appropriate.

Second, I wonder if there are other redundant instructions you could factor out? Right now, the PR makes `int3` and `printInvocation` simple instructions, but we still haven't decided if we want to keep `printInvocation`. So the changes here have factored out the code for one, maybe two, instructions, and adding a new instruction (simple or otherwise) is pretty rare. (Another thing you could look at is the diff: as a *very* loose rule of thumb, a refactor that adds more code than it removes might not be worth it.)

Maybe it makes more sense to do the refactoring for individual file? e.g., `BC.cpp`, `BC.h`, and `BC_inc.h` have a lot of "trivial" code.

Please take a look and let me know what you think. If it doesn't seem worthwhile, the symbols refactoring is still worthwhile and something we could merge in.

Jakobeha:
Hello,

Sorry this took so long, I've been busy with finals and adjusting for winter break, so I haven't worked on this until very recently.

I applied all of the changes you requested:
- Small changes [556df88]
- I also found out what was giving cppcheck an issue, I didn't need to actually suppress anything, I just needed to fix an import path to `simple_instruction_list.h` (for future reference: cppcheck wanted `../../simple_instruction_list.h` for one of the files, every other one works with just `simple_instruction_list.h`) [0f2f232]
- Lastly, I abstracted all of the `BC` constructors which take no arguments and don't do anything, in `BC.cpp`, `BC.h`, and `BC_inc.h` (e.g. `inline static BC <name>();`). Now, you just need to add an entry to `BC_noarg_list.h` instead of changing those 3 files. [1f7feb3]

mhyee (approved):

mhyee:
Thanks for the update, and no worries about being slow. I should also apologize for being busy on my end.

Good catch with the cppcheck issue, and thanks for summarizing your commits!

This is good to merge, except that there's still a minor conflict, which is why the tests are failing. @o- and I thought there would be a clean merge, but `identical_` was renamed to `identical_noforce_` in a257d538febf74751ecf675223c25332ef270614, and the merge commit messed things up a bit.

Once you resolve this and all the tests pass, I'll merge it in.

PR #271, opened 2018-12-03
Cleanup analysis

o-:
@JanJecmen 

since we talked about this, I felt like I can cleanup the whole segments business in the static analysis.

This PR is only for the top commit 142dfc178f251d321ffd5024b0e083f984b909fc all the rest is from #267

### DIFF
--- a/rir/src/compiler/analysis/generic_static_analysis.h
+++ b/rir/src/compiler/analysis/generic_static_analysis.h
@@ -55,16 +55,19 @@ class StaticAnalysis {
   private:
     const std::string name;
 
-    std::vector<std::vector<AbstractState>> mergepoint;
+    struct BBSnapshot {
+        bool seen = false;
+        AbstractState entry;
+        std::unordered_map<Instruction*, AbstractState> extra;
+    };
+    typedef std::vector<BBSnapshot> AnalysisSnapshots;
+    AnalysisSnapshots snapshots;
+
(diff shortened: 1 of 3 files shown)


rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -197,13 +197,8 @@ class StaticAnalysis {
...
 
-                    if (CallInstruction::CastCall(i)) {
-                        segment++;
-                        if (mergepoint[id].size() <= segment) {
-                            mergepoint[id].resize(segment + 1);
-                        }
-                        mergepoint[bb->id][segment] = state;
-                    }
+                    if (res.keepSnapshot)
+                        snapshots[bb->id].extra[i] = state;
```
JanJecmen, line 201: So now you can keep snapshots for any instruction, not just calls? And you pick which ones in the particular analysis?

JanJecmen (approved):
This reads way better than the segments version :+1:

rir/src/compiler/analysis/generic_static_analysis.h
o-: that's the idea. because I realized that e.g. scope analysis should also keep snapshots after analyzing force, but only if we actually dived into the promise.

otoh for many cheaper analysis it does not make sense to keep any snapshots.

PR #272, opened 2018-12-05
massive cleanup of scope analysis

o-:

PR #273, opened 2018-12-06
[wip] towards compiling with assumptions 

o-:
based on #272 (ignore first two commits)

I open this to start a discussion how the api of assumptions could look like.

The idea is that there is an *OptimizationContext* under which code is compiled. The OptimizationContext captures the things the compiler is allowed to take for granted.

Then the next question will be, how does the runtime ensure the context is always matching. But in this PR I only use the context for things we can deduce from contextual information.

## WIP

There is a problem here, namely we store all versions into slot 1 and we always call slot 1, even functions compiled under assumptions.

Here:
https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/translations/pir_2_rir.cpp#L955 
we would potentially end up compiling a function compiled under assumptions into the generic slot1.

To fix this my first impulse woud be to just extend `static_call_` instruction with a fixed slot index.
That should be enough for the case here. Otoh, should we already start to tie it in with the signature mechanism? @mhyee any opinions?


o-:
To understand the proposal just look at  ....

Update: my `git add -p` skills are terrible. Had to fix it. The main commit is now 1bf12697f0455472f1e1df66c9da3fe8a2855b2d

rir/src/compiler/pir/optimization_context.h
```diff
@@ -0,0 +1,43 @@
...
+#include "runtime/Function.h"
+
+namespace rir {
+namespace pir {
+
+struct OptimizationContext {
+    OptimizationContext(Env* environment, const AssumptionsSet& assumptions)
+        : environemnt(environment), assumptions(assumptions) {}
+
+    Env* environemnt;
```
mhyee: environemnt -> environment

mhyee:
> Then the next question will be, how does the runtime ensure the context is always matching. But in this PR I only use the context for things we can deduce from contextual information.

What's the joke about how "context" is used in computer science for whatever stuff we want to keep around but don't know what to name it? More seriously, "use the context for things we can deduce from contextual information" seems like a recursive definition.

> [...]
> To fix this my first impulse woud be to just extend `static_call_` instruction with a fixed slot index.
> That should be enough for the case here. Otoh, should we already start to tie it in with the signature mechanism? @mhyee any opinions?

I don't think we have a good sense of how to tie the assumptions with the signatures? It's all very uncertain right now, so I would go with the simplest solution for now, and expect to change it again (and again) later.

rir/src/runtime/DispatchTable.h
```diff
@@ -18,30 +18,56 @@ typedef SEXP DispatchTableEntry;
...
 
-    void put(size_t i, Function* f) {
-        assert(i < capacity());
-        setEntry(i, f->container());
+    void baseline(Function* f) {
+        assert(f->signature().optimization ==
+               FunctionSignature::BaselineVersion);
+        setEntry(0, f->container());
+        if (size() == 0)
+            size_++;
```
charig, line 36: We are assuming there will always be a baseline compiled. Maybe that is not necessary...

rir/src/runtime/FunctionSignature.h
```diff
@@ -2,13 +2,25 @@
...
+#include "compiler/pir/assumptions.h"
+
 #include <iomanip>
 #include <iostream>
 #include <vector>
 
 namespace rir {
 
 struct FunctionSignature {
+    enum EnvironmentCreation {
```
charig, line 14: Environment
 Provided
 NotProvided

rir/src/runtime/FunctionSignature.h
```diff
@@ -35,50 +47,60 @@ struct FunctionSignature {
...
+            for (auto i = assumptions.begin(); i != assumptions.end(); ++i) {
+                out << *i;
+                if (i + 1 != assumptions.end())
+                    out << ",";
+            }
+            out << "]";
+        }
     }
 
     FunctionSignature() = default;
```
charig, line 88: Shouldn't we remove the default constructor now?

rir/src/runtime/DispatchTable.cpp
```diff
@@ -0,0 +1,3 @@
+#include "DispatchTable.h"
```
charig, line 1: ?

rir/src/compiler/pir/assumptions.h
```diff
@@ -0,0 +1,32 @@
...
+enum class Assumptions {
+    CorrectNumberOfArguments,
+    MaxNumberOfArguments,
+    CorrectOrderOfArguments,
+
+    FIRST = CorrectNumberOfArguments,
+    LAST = CorrectOrderOfArguments
+};
+
+struct AssumptionsSet : public EnumSet<Assumptions> {
```
charig, line 20: Maybe enum is an Assumption  and here we can call it Assumptions

PR #274, opened 2018-12-10
fix BBTransform::clone copies instructions without copying promises

o-:
when copying instructions from one function to another, we would not
copy promises, that where used in MkArgs. This would cause pointers
across different closures and ultimatively to segv, if the old function
is deleted (or the promise in the old function).


o-:
bugfix

PR #278, opened 2018-12-12
new pass schedule

o-:
1. fix the examples: if we compile after only one execution, we
accidentially think that the inner function has a monomorphic target
closure. (which is not true, since it gets recreated for every
invocation)
2. increase inliner limits a bit
3. a new pass schedule, that is still completely arbitrary, but at least
has some justification for the parts :)

### DIFF
--- a/examples/tree_mini.r
+++ b/examples/tree_mini.r
@@ -24,4 +24,5 @@ test <- function() {
 f <- rir.compile(function() test())
 
 f()
+f()
 pir.compile(test, debugFlags=pir.debugFlags(PrintPirAfterOpt=TRUE, ShowWarnings=TRUE))
--- a/examples/tree_mini.r
+++ b/examples/tree_mini.r
@@ -23,6 +23,7 @@ test <- function() {
 
 f <- rir.compile(function() test())
 
+# Run twice to get realistic type feedback
 f()
 f()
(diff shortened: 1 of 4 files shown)


examples/tree_mini.r
```diff
@@ -23,5 +23,6 @@ test <- function() {
 
 f <- rir.compile(function() test())
```
charig, line 25: A comment here saying we need to run twice (at least at the moment) to avoid confusing our profiling with partial data may save a lot of time to other who eventually do not understand the rationale and remove one of the calls

rir/src/compiler/opt/inline.cpp
```diff
@@ -19,8 +19,8 @@ class TheInliner {
     Closure* function;
     explicit TheInliner(Closure* function) : function(function) {}
 
-    static constexpr size_t MAX_SIZE = 1024;
-    static constexpr size_t MAX_INLINEE_SIZE = 128;
+    static constexpr size_t MAX_SIZE = 10000;
```
charig: Just as a comment: I guess more sooner than later it would be quite useful to tweak this values with env vars.

charig (approved):

o-:
excellent points

PR #279, opened 2018-12-13
fix soundness issue in static analysis

o-:
currently only return statements are merged into the exitstate in
the static analysis. while this might be correct for some analysis
it is certainly not correct in general.

for example the dead store optimization looks at result() of the
analysis to determine whether a load is visible. If the load would
be in the deopt branch we would not see it.

I looked at all the places where we currently use result() and it
seems to me, that including the deopt state is the correct choice.

thus let's always include the deopt state for now and we can still
make it more clever in the future.

Also, this revealed a case in the scope analysis, where we did not
consider that a deopt influences the return value of a function.

### DIFF
--- a/rir/src/compiler/analysis/generic_static_analysis.h
+++ b/rir/src/compiler/analysis/generic_static_analysis.h
@@ -206,17 +206,15 @@ class StaticAnalysis {
                 }
 
                 if (bb->isExit()) {
-                    if (!Deopt::Cast(bb->last())) {
-                        if (DEBUG_LEVEL >= AnalysisDebugLevel::Exit) {
-                            log << "===== Exit state is\n";
-                            log(state);
-                        }
-                        if (reachedExit) {
-                            exitpoint.merge(state);
-                        } else {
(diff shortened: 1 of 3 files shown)


o-:
@JanJecmen the discussion yesterday made me actually look at this again. and it seems rather dangerous to ignore the deopt in the analysis. so I fixed it in the forward analysis by always considering the deopt case.

In general, we might have some analysis that only looks at the state at the returns, but not at the state of deopts. but for now this seems to be enough.

it seems to me that in forward mode it is easy to compute both at the same time (ie. just keep a merged state of all returns and deopts vs. a merged state of just the returns). but in backwards mode you can't, because either you start the analysis at all exit points, or just at the returns. So you have to run it twice to get both results, ie. it's a parameter of the whole analysis.

For now I would just start at any instruction marked isExit. That's the safe choice. anything else would need to be justified.

JanJecmen (approved):
lgtm
good point about the forward vs. backward exit points...

PR #280, opened 2018-12-13
add some infrastructure to mark builtins safe

o-:
safe builtins are those who do not access the caller environment in
any way.

The list needs to be extended of course...

@charig I have this lying around from another PR that needs more thinking. I guess it's small, but still worth salvaging

### DIFF
--- a/rir/src/compiler/pir/instruction.cpp
+++ b/rir/src/compiler/pir/instruction.cpp
@@ -1,6 +1,7 @@
 #include "instruction.h"
 #include "pir_impl.h"
 
+#include "../util/safe_builtins_list.h"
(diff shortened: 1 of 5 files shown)


charig (approved):
Nice!
Only one comment: I have the impression that most builtins are safe. And with most I mean > 80-90%.
If this is the case, maybe it would be better to make this just the other way with a list of unsafe builtins.

o-:
> Only one comment: I have the impression that most builtins are safe. And with most I mean > 80-90%.
> If this is the case, maybe it would be better to make this just the other way with a list of unsafe builtins.

this might be the case. but currently I have not yet looked at them...

PR #282, opened 2018-12-13
Assumptions stuff

o-:
(see commits)

### DIFF
--- a/rir/src/compiler/opt/assumptions.cpp
+++ b/rir/src/compiler/opt/assumptions.cpp
@@ -1,0 +1,58 @@
+#include "../pir/pir_impl.h"
+#include "../translations/pir_translator.h"
+#include "../util/cfg.h"
+#include "../util/visitor.h"
+
+#include "R/r.h"
+#include "pass_definitions.h"
+
+#include <unordered_map>
+#include <unordered_set>
+
+namespace rir {
+namespace pir {
+
+void OptimizeAssumptions::apply(RirCompiler&, Closure* function,
+                                LogStream&) const {
+    CFG cfg(function);
+    Visitor::run(function->entry, [&](BB* bb) {
+        auto ip = bb->begin();
+        while (ip != bb->end()) {
+            auto next = ip + 1;
+            auto instr = *ip;
+
+            if (auto assume = Assume::Cast(instr)) {
+                auto cp = assume->checkpoint();
+                auto bb = cp->bb();
+                // We are trying to group multiple assumes into the same
+                // checkpoint by moving up linear sequence of bbs, to find the
+                // topmost possible checkpoint.
+                // TODO: we could also try to move up the assume itself, since
+                // if we move both at the same time, we could even jump over
+                // effectful instructions.
+                while (true) {
+                    if (!cfg.hasSinglePred(bb))
+                        break;
+
+                    bool hasEffect = false;
+                    for (auto i : *bb)
+                        if (i->hasEffect())
+                            hasEffect = true;
+                    if (hasEffect)
+                        break;
+                    bb = cfg.immediatePredecessors(bb).front();
+
+                    if (!bb->isEmpty())
+                        if (auto prevCp = Checkpoint::Cast(bb->last()))
+                            cp = prevCp;
+                }
+                assume->checkpoint(cp);
+            }
+
+            ip = next;
+        }
+    });
+}
+
+} // namespace pir
+} // namespace rir
--- a/rir/src/compiler/opt/assumptions.cpp
+++ b/rir/src/compiler/opt/assumptions.cpp
@@ -25,8 +25,8 @@ void OptimizeAssumptions::apply(RirCompiler&, Closure* function,
                 auto cp = assume->checkpoint();
                 auto bb = cp->bb();
                 // We are trying to group multiple assumes into the same
-                // checkpoint by moving up linear sequence of bbs, to find the
-                // topmost possible checkpoint.
+                // checkpoint by finding for each assume the topmost compatible
+                // checkpoint.
                 // TODO: we could also try to move up the assume itself, since
                 // if we move both at the same time, we could even jump over
                 // effectful instructions.
--- a/rir/src/compiler/opt/delay_instr.cpp
+++ b/rir/src/compiler/opt/delay_instr.cpp
@@ -10,17 +10,15 @@ void DelayInstr::apply(RirCompiler&, Closure* function, LogStream&) const {
     std::vector<MkEnv*> envs;
 
     Visitor::run(function->entry, [&](BB* bb) {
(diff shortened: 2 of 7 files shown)


rir/src/compiler/opt/delay_instr.cpp
```diff
@@ -15,8 +15,12 @@ void DelayInstr::apply(RirCompiler&, Closure* function, LogStream&) const {
             auto i = *ip;
             auto next = ip + 1;
 
-            if (!i->hasEnv() && !i->hasEffect() && !Phi::Cast(i) &&
-                !i->branchOrExit()) {
+            if (LdFun::Cast(i) ||
+                // For LdFun we allow the instruction to be delayed, since
+                // probably nobody cares about when exactly looking up a
```
charig: Mmmm, just wondering, if this is the case, maybe we can just declare them effectfulness and move the comment there?

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -105,6 +105,12 @@ class PASS(CleanupCheckpoints, "Cleanup unused checkpoints");
  */
 class PASS(CleanupFramestate, "Cleanup framestates unused by checkpoints");
 
+/*
+ * Trying to group assumptions, by pushing them up. This well lead to fewer
```
charig, line 109: There is a typo: `well -> will`. But now that I already write this, I would rephrase:
`Tries to group assumptions to reduce the number of checkpoints used overall`

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -297,15 +298,18 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
             }
 
             std::string name = "";
-            if (auto ldfun = LdFun::Cast(callee))
+            if (ldfun)
```
charig, line 301: why not just put the `auto ldfun = LdFun::Cast(callee);` here, in line 300?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -297,15 +298,18 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
...
             Assumptions asmpt(Assumption::CorrectOrderOfArguments);
             asmpt.set(Assumption::CorrectNumberOfArguments);
             compiler.compileClosure(
                 monomorphic, name, asmpt,
                 [&](Closure* f) {
                     Value* expected = insert(new LdConst(monomorphic));
-                    Value* t = insert(new Identical(callee, expected));
+                    Value* given = callee;
+                    if (ldfun)
+                        given = insert(new LdVar(ldfun->varName, ldfun->env()));
```
charig, line 313: I think there are two scenarios here.
1. There is a binding in the environment.
2. There is not.
For case number 2 I do not see how this helps. Even worst, it adds an instruction. For case number 1 well, if I guess you are looking for the case we can prove the binding is a constant and thus we can remove the whole speculation.
In case I am right maybe it would better to put this on pir instead of rir...

rir/src/compiler/opt/assumptions.cpp
```diff
@@ -0,0 +1,58 @@
...
+    Visitor::run(function->entry, [&](BB* bb) {
+        auto ip = bb->begin();
+        while (ip != bb->end()) {
+            auto next = ip + 1;
+            auto instr = *ip;
+
+            if (auto assume = Assume::Cast(instr)) {
+                auto cp = assume->checkpoint();
+                auto bb = cp->bb();
+                // We are trying to group multiple assumes into the same
```
charig, line 27: I was a little confused by the comment. `moving up linear sequence of bbs` made me think we are moving bbs.
I would say: `We are trying to group multiple assumes into the same checkpoint by finding for each assume the topmost compatible checkpoint.`
    
Then, again, just brainstorming...I have the impression this is becoming a kind of pattern we are facing many times. 
Maybe we can add to cfg a function `findTopmostComplyingWith(Instruction inst, lambda condition)` and iterate it till we reach a general version. I mean the code here would be much easier to read if we replace the whole while with

charig (approved):

o-:
@charig so maybe to explain better my idea in the first commit.

Instead of this:
```
    f = ldfun("f")
    c = checkpoint()  -> BB1 | BB2
BB2:
    deopt(stack=[f])
BB1:
    assume (f==someFunction, c)
```

I want this:
```
    c = checkpoint()  -> BB1 | BB2
BB2:
    f1 = ldfun("f")
    deopt(stack=[f1])
BB1:
    f2 = ldvar("f")
    assume (f2==someFunction, c)
```

because f2 has no sideeffect the code is better for optimizations.

(I should probably add that diagram to the code)

I don't think that we can mark ldfun as having no sideeffects in general, since that would certainly break some optimiztions. I think what would be better is to make it even more explicit, ie. just allow ldfun to move over checkpoint, because that is always safe to do. The reason is that we do not allow sideeffects between assume and checkpoint, so the ldfun sideeffect in the example above still happens at the correct time.

charig:
> @charig so maybe to explain better my idea in the first commit.
> 
> Instead of this:
> 
> ```
>     f = ldfun("f")
>     c = checkpoint()  -> BB1 | BB2
> BB2:
>     deopt(stack=[f])
> BB1:
>     assume (f==someFunction, c)
> ```
> I want this:
> 
> ```
>     c = checkpoint()  -> BB1 | BB2
> BB2:
>     f1 = ldfun("f")
>     deopt(stack=[f1])
> BB1:
>     f2 = ldvar("f")
>     assume (f2==someFunction, c)
> ```
> because f2 has no sideeffect the code is better for optimizations.
> 
> (I should probably add that diagram to the code)
> 
> I don't think that we can mark ldfun as having no sideeffects in general, since that would certainly break some optimiztions. I think what would be better is to make it even more explicit, ie. just allow ldfun to move over checkpoint, because that is always safe to do. The reason is that we do not allow sideeffects between assume and checkpoint, so the ldfun sideeffect in the example above still happens at the correct time.

Ok. But then in your code there is a hidden assumption: that the callee `ldFun f` is not used anywhere else and thus, a later delay pass will move it to the deopt branch. If this is not true then we may end up with worst code instead of better. I do not know if ldFun could be used more than once but I assume so.

o-:
> But then in your code there is a hidden assumption: that the calle ldFun f is not used anywhere else and thus, a later delay pass will move it to the deopt branch. If this is not true then we may end up with worst code instead of better. 

that is true. it's very hard to know in advance if an optimization is a net win. e.g. maybe there is a second use, but the second use can be resolved by the same trick...

PR #283, opened 2018-12-13
aslogical should not be confused with astest

o-:
we compiled RIR asbool_ to PIR AsLogical, which is wrong. As logical
will trim the result to a length 1 logical vector (it is used for ||
and && and such), but astest/asbool_ are used for branches and this
one will throw a warning if the size is not 1 and will throw an error
if the result is NA.


o-:
tbr, since bugfix

PR #284, opened 2018-12-13
introduce a true and a false value

o-:
this commit introduces a constant true and false pir value. Since
this is the first commit that introduces a value (that is not an
instruction) that can occur as argument, i had to extend the backend
to support loading of values into the correct argument position.


o-:
The reason why I need this: to be able to test deoptimization, I want to be able to write `assume(false)` and `assumeNot(true)`.

As a bonus it simplifies the constantfolding a lot

PR #286, opened 2018-12-14
Speculate on builtins

o-:
Instead of just speculating on closures, we should also speculatively insert builtin calls (last commit).

Initially I had some bugs in the implementation. I wanted to be more confident in the deoptimization mechanism and therefore implemented the `PIR_DEOPT_CHAOS` mode, that triggers 20% of all deopts, regardless of the assumption (2nd commit). The good news is that I did not find any issues.

For the deoptimization test mode I wanted to be able to emit `assume(false)` and `assumeNot(true)`. For this I added a true and false singleton value to PIR (first commit).

The problem (in the first commit) was, that our register allocation backend is only able to handle arguments from instructions. The only exception where environments, but they where implemented with a hack (since they are always the last argument, they are always tos, so we can just push them without stack shuffling). So I additionally implemented the necessary code to load values to the correct stack location. After all it was simpler than thought. The argument loading code counts how many args are already on the stack. The stack shuffling consists of just `put (numberOfArgsAlreadyOnStack - argumentNumber)`. E.g. if 3 arguments are already on the stack, and you are pushing the second argument, then it needs to be put between the first and second already there. So it's `push val; put 2`.

A nice side-effect of the true/false value is, it simplified constant dead branch removal. since now this is really two steps: first step constant fold the condition to true/false, second: remove branch. And not like before, where we would have to evaluate the the branch condition and then removing in one step.

### DIFF
--- a/rir/src/compiler/opt/constantfold.cpp
+++ b/rir/src/compiler/opt/constantfold.cpp
@@ -4,6 +4,7 @@
 #include "../util/cfg.h"
 #include "../util/visitor.h"
 #include "R/Symbols.h"
+#include "R/r.h"
 #include "pass_definitions.h"
 
 #include <unordered_set>
@@ -13,13 +14,14 @@ namespace {
 using namespace rir::pir;
 
 static LdConst* isConst(Value* instr) {
-    if (auto cst = LdConst::Cast(instr->followCastsAndForce())) {
+    instr = instr->followCastsAndForce();
+    if (auto cst = LdConst::Cast(instr)) {
         return cst;
     }
     return nullptr;
 }
 
-#define FOLD(Instruction, Operation)                                           \
+#define FOLD2_SIMPLE(Instruction, Operation)                                   \
     do {                                                                       \
         if (auto instr = Instruction::Cast(i)) {                               \
             if (auto lhs = isConst(instr->arg<0>().val())) {                   \
@@ -35,9 +37,18 @@ static LdConst* isConst(Value* instr) {
         }                                                                      \
     } while (false)
 
-#define FOLD2(__i__, Instruction, Operation)                                   \
+#define FOLD1(Instruction, Operation)                                          \
     do {                                                                       \
-        if (auto instr = Instruction::Cast(__i__)) {                           \
+        if (auto instr = Instruction::Cast(i)) {                               \
+            if (auto arg = isConst(instr->arg<0>().val())) {                   \
+                Operation(arg->c);                                             \
+            }                                                                  \
+        }                                                                      \
+    } while (false)
+
+#define FOLD2(Instruction, Operation)                                          \
+    do {                                                                       \
+        if (auto instr = Instruction::Cast(i)) {                               \
             if (auto lhs = isConst(instr->arg<0>().val())) {                   \
                 if (auto rhs = isConst(instr->arg<1>().val())) {               \
                     Operation(lhs->c, rhs->c);                                 \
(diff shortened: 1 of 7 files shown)


rir/src/compiler/opt/constantfold.cpp
```diff
@@ -13,13 +14,14 @@ namespace {
...
-    if (auto cst = LdConst::Cast(instr->followCastsAndForce())) {
+    instr = instr->followCastsAndForce();
+    if (auto cst = LdConst::Cast(instr)) {
         return cst;
     }
     return nullptr;
 }
 
-#define FOLD(Instruction, Operation)                                           \
+#define FOLD2_SIMPLE(Instruction, Operation)                                   \
```
charig: FOLD_BINARY_NATIVE

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -35,9 +37,18 @@ static LdConst* isConst(Value* instr) {
         }                                                                      \
     } while (false)
 
-#define FOLD2(__i__, Instruction, Operation)                                   \
+#define FOLD1(Instruction, Operation)                                          \
```
charig: FOLD_UNARY_CUSTOM

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -35,9 +37,18 @@ static LdConst* isConst(Value* instr) {
...
     do {                                                                       \
-        if (auto instr = Instruction::Cast(__i__)) {                           \
+        if (auto instr = Instruction::Cast(i)) {                               \
+            if (auto arg = isConst(instr->arg<0>().val())) {                   \
+                Operation(arg->c);                                             \
+            }                                                                  \
+        }                                                                      \
+    } while (false)
+
+#define FOLD2(Instruction, Operation)                                          \
```
charig: FOLD_BINARY_CUSTOM

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -65,69 +76,85 @@ void Constantfold::apply(RirCompiler& cmp, Closure* function,
...
+            FOLD2_SIMPLE(IDiv, symbol::Idiv);
+            FOLD2_SIMPLE(Lt, symbol::Lt);
+            FOLD2_SIMPLE(Gt, symbol::Gt);
+            FOLD2_SIMPLE(Lte, symbol::Le);
+            FOLD2_SIMPLE(Gte, symbol::Ge);
+            FOLD2_SIMPLE(Eq, symbol::Eq);
+            FOLD2_SIMPLE(Neq, symbol::Ne);
+            FOLD2_SIMPLE(Pow, symbol::Pow);
+
+            auto convertsToLglWithoutWarning = [](SEXP arg) {
```
charig: The name is already too long. So logical(I prefer boolean) instead would not hurt here.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -324,6 +324,18 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
...
+                   TYPEOF(monomorphic) == BUILTINSXP) {
+            // TODO implement support for call_builtin_ with names
+            Value* expected = insert(new LdConst(monomorphic));
+            Value* given = callee;
+            if (ldfun)
+                given = insert(new LdVar(ldfun->varName, ldfun->env()));
+            Value* t = insert(new Identical(given, expected));
+            auto cp = insert.addCheckpoint(srcCode, pos, stack);
+            insert(new Assume(t, cp));
+            pop();
```
charig, line 337: Why there is not a framestate here?
Btw, this code is extremely similar to the lambda in compileClosure(), would be good to factorize.
In a first sight seems that the else if could be include inside the above if managing some conditions (I mean would be more clear if we separate the differences inside one big if for monomorphics).

charig (approved):
I reviewed everything but for the backend changes(pir_2_rir) I only skimmed through them. Maybe later @JanJecmen wants to give a second opinion on that.

rir/src/compiler/opt/constantfold.cpp
o-: in R parlance logical is a logical vector, whereas boolean a *scalar* logical vector.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: framestate on calls is used only for inlining. builtins cannot be inlined...

PR #288, opened 2018-12-14
identify safe builtins

o-:
I eyeballed a couple of builtin functions to see if they access
environments. Unfortunately the vast majority does so in the case
of object arguments. So I made two classes, one of them are always
safe. The others just for non-objects.


o-:
i need this now

rir/src/compiler/util/safe_builtins_list.cpp
```diff
@@ -5,17 +5,277 @@
 namespace rir {
 namespace pir {
 
-bool SafeBuiltinsList::contains(SEXP builtin) {
+bool SafeBuiltinsList::always(int builtin) {
     static int safeBuiltins[] = {
         findBuiltin("vector"),
-        findBuiltin("c"),
```
JanJecmen, line 11: Where did `c` go?

o-: thanks that was an accident. It should be in the semi-safe list

PR #289, opened 2018-12-14
avoid deopts when encountering lazy loading stubs

o-:
our guards would fail constantly, because we where comparing the guarded
functions against their lazyloading stub. The reason is that ldvar would
return the stub, without forcing it. Even though the promise is already
long forced. The solution here is a bit hacky, but for now let's just
make identical check if either argument is an already evaluated promise.

Additionally this commit introduces the PIR_DEBUG_DEOPTS env variable
that allows us to print diagnostic information when deopts happen.

### DIFF
--- a/rir/src/compiler/transform/bb.cpp
+++ b/rir/src/compiler/transform/bb.cpp
@@ -1,6 +1,8 @@
 #include "bb.h"
 #include "../pir/pir_impl.h"
 #include "../util/visitor.h"
+#include "R/Funtab.h"
+#include "R/r.h"
 
 #include <unordered_map>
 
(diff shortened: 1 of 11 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -1899,6 +1899,14 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
         INSTRUCTION(identical_) {
             SEXP rhs = ostack_pop(ctx);
             SEXP lhs = ostack_pop(ctx);
+            // This instruction does not force, but we should still compare
+            // the actual promise value if it is already forces.
```
mhyee: Typo: forces -> forced

rir/src/compiler/transform/bb.cpp
```diff
@@ -110,8 +112,25 @@ Value* BBTransform::forInline(BB* inlinee, BB* splice) {
 }
 
 BB* BBTransform::lowerExpect(Code* code, BB* src, BB::Instrs::iterator position,
-                             Value* condition, bool expected, BB* deoptBlock) {
+                             Value* condition, bool expected, BB* deoptBlock,
+                             const std::string& debugMessage) {
     auto split = BBTransform::split(code->nextBBId++, src, position + 1, code);
+
+    static SEXP print = Rf_findFun(Rf_install("cat"), R_GlobalEnv);
```
mhyee, line 119: Shouldn't debugging stuff be behind some macro so they're gone in release mode?

mhyee (commented):
I have a nitpicky concern about the hacky solution. My problem is that `identical_` no longer does what it says. It'll be clever and try to look inside a promise.

For example, it means a non-promise value will be "identical" to a promise, if the non-promise value is the result of forcing the promise. I guess this is what we want to work around the lazy loading stubs, but it could be surprising in other cases. (I guess we're also saved by implementing `identical_` to compare `SEXP`s with C pointer equality.)

I think I'd be happier if we had a separate instruction for this, some sort of "promise equivalence," I guess. Or if there's a better alternative.

But this isn't a deal breaker, and I'm fine with merging in this PR and leaving a TODO/issue for a better solution.

o-:
> For example, it means a non-promise value will be "identical" to a promise, if the non-promise value is the result of forcing the promise. I guess this is what we want to work around the lazy loading stubs, but it could be surprising in other cases. (I guess we're also saved by implementing identical_ to compare SEXPs with C pointer equality.)

this `identical_` instruction is only used in guards and has nothing to do with equality in the source language. even more so, we currently just use it to compare functions.

I agree that now it's kind of a semantically silly instruction. it's pointer equality, modulo evaluated promises.

but on the other hand it is a safe approximation of pointer equality after executing ldfun, but without forcing anything -- which is what we actually want for the function guards...

rir/src/compiler/transform/bb.cpp
o-: but arguably you would debug perf issues in release mode, so I think this is appropriate to keep. plus there is certainly no overhead from having this.

o-:
so I renamed the instruction to `identical_noforce_`. that should hopefully make it more transparent that the instruction does something funky regarding promises.

PR #290, opened 2018-12-17
Refactoring

charig:
This refactoring just changes the `pir2rir` compiler so that it does not have anymore the responsibility of installing the resulting optimized function in the corresponding place. It now just returns the function and it is the responsibility of the client to decide what to do. It also removes the `origin` and `next` fields from RIR functions which was created for managing different optimizations levels of the same function but it is actually not used. For managing the different optimization levels we resort to the different slots of the dispatch table.


o-:
awesome

PR #291, opened 2018-12-19
eagerly force arguments to builtin functions

o-:
see individual commit msgs.

We missed this in the past, so we would actually pass unevaluated promises. (in short this means that speculating on builtins was mostly broken :) )

but, for that reason I had to move the guard before the evaluation of the promises, because of course if we force an argument, then find out that we are actually not calling a builtin after all, then the forcing would have been wrong.

### DIFF
--- a/rir/src/compiler/opt/elide_env.cpp
+++ b/rir/src/compiler/opt/elide_env.cpp
@@ -45,8 +45,10 @@ void ElideEnv::apply(RirCompiler&, Closure* function, LogStream&) const {
                             if (v != i->env())
                                 args.push_back(v);
                         });
-                        bb->replace(
-                            ip, new CallSafeBuiltin(b->blt, args, b->srcIdx));
+                        auto safe =
+                            new CallSafeBuiltin(b->blt, args, b->srcIdx);
+                        b->replaceUsesWith(safe);
+                        bb->replace(ip, safe);
                         envIsNeeded = false;
                     }
                 }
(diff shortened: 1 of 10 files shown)


rir/src/compiler/opt/elide_env.cpp
```diff
@@ -45,8 +45,10 @@ void ElideEnv::apply(RirCompiler&, Closure* function, LogStream&) const {
                             if (v != i->env())
                                 args.push_back(v);
                         });
-                        bb->replace(
-                            ip, new CallSafeBuiltin(b->blt, args, b->srcIdx));
+                        auto safe =
+                            new CallSafeBuiltin(b->blt, args, b->srcIdx);
+                        b->replaceUsesWith(safe);
+                        bb->replace(ip, safe);
```
charig, line 51: Just wondering...Is there any case in which we replace and do not want to replace uses?

charig (approved):
Changes seems ok modulo build is failing. Most were already approved in another PR.
The changes in the processing of calls in `rir_2_pir` will clash with #292.

rir/src/compiler/opt/elide_env.cpp
o-: hmm, probably zero...

o-: ok, I tried to change it. but there are some places (e.g. in lower expect) where the concerns (replacing uses vs. replacing instructions) are separated between different parts of the code. so it's not a simple change and thus I'm not gonna add it here.

o-:
>  The changes in the processing of calls in rir_2_pir will clash with #292.

yeah, that's because I had to reorder the guard. but it should subsume your refactoring, as this change also pulls the guard creation out and avoids the duplicated code

PR #292, opened 2018-12-19
More precise call profiling

charig:
## PR Contents 
Currently, the profiler adds a new CLOSXP to the observed call targets whenever it founds a new one.
But, this is unprecise. The reason is that the CLOSXP contains not only the formals and the dispatch table, but also the environment. And currently, we are not assuming anything about the environment for our optimizations. To be concrete consider this code snippet: 
```
test <- function() {
   a <- function() {}
   a()
}
```

If I run `test()` twice the profiler will contain two different targets for `a()` because each time we call `test()` it creates a different closure. This closure will probably point to the same formals and the same DT but the enclosing environment may be different. However, the code actually called does not depend on the environment.  

Beyond some cleaning, this PR mainly changes the profiler so that when a target is being added, in case there already exist in the chain a CLOSXP with the same dt and the same formals it does not consider it as new and so does not insert it. 

## Performance
The main drawback of the PR is that now, at runtime, for every call, when profiling is enabled, we need to execute two pointer equalities instead of one for deciding if we add the target to the observed targets chain. But, 1) this may not be so significant 2) at this particular moment we are focusing on the performance of the optimized call and this should help on that end 3) I have the impression that the DT somehow subsumes the formals, at least, for what regards to this PR. On that case, it would suffice to only compare for the DT (BODY). 

## Other Considerations
* What should happen if we execute the same function but from a different slot (different assumptions)
* The call target profiling/speculation at this point is very basic. So by speculating on targets we can only convert a `call` into a `static call` and then eventually inline. But we still have to always do the "lookup" at runtime by running the ldFun. If it is here only for inlining, ok, but maybe we should get the real examples in which our static analysis could not do the same. On the other hand, what would be really interesting is to devise a way of using profiling information to help removing the `ldfun` and avoid doing the "lookup" at run time.

### DIFF
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
@@ -131,6 +131,16 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
 namespace rir {
 namespace pir {
 
+Checkpoint* Rir2Pir::addCheckpoint(rir::Code* srcCode, Opcode* pos,
+                                   const RirStack& stack,
+                                   Builder& insert) const {
+    // Checkpoints in promises are badly supported (cannot inline promises
+    // anymore) and checkpoints in eagerly inlined promises are wrong. So for
+    // now we do not emit them in promises!
+    assert(!inPromise());
+    return insert.emitCheckpoint(srcCode, pos, stack);
+}
+
 bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
                         rir::Code* srcCode, RirStack& stack, Builder& insert,
                         CallTargetFeedback& callTargetFeedback) const {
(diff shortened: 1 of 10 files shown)


rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -281,61 +277,61 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
...
-                                               bc.callExtra().callArgumentNames,
-                                               args)
-                    : RList(FORMALS(monomorphic)).length() == args.size();
-
-            if (!correctOrder) {
+        if (monomorphic) {
+            auto target = callTargetFeedback.at(callee).getTarget(srcCode, 0);
+
+            auto insertSpecPrologue = [&]() {
+                Value* expected = insert(new LdConst(target));
```
o-: but this can't work. this guard will just immediately fail. you are still comparing with the closure here, which, as you note in this pr, changes on every invocation.

charig:
Regression test of why this PR is invalid in PR#295

PR #294, opened 2018-12-20
fix elide_env issue

o-:
In general we should make sure that our framestates capture
the forced versions of a value, if we are going to force it.

additionally the speculation on `a[b,c]` is currently wrong.
we need to look at it more carefully and decide what to do.


o-:
another bugfix, tbr

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -148,6 +148,12 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
     auto top = [&stack]() { return stack.at(0); };
     auto set = [&stack](unsigned i, Value* v) { stack.at(i) = v; };
 
+    auto force = [&](unsigned i) {
```
charig: forceIfLazy?

PR #295, opened 2019-01-09
Add regression test 

charig:
This test summarizes the lesson learnt by failed PR #292. Code was provided by @o-.

### DIFF
--- a/rir/tests/pir_regression.R
+++ b/rir/tests/pir_regression.R
@@ -81,3 +81,19 @@ hc3 = .Call("rir_invocation_count", h)
 ic3 = .Call("rir_invocation_count", i)
 stopifnot(hc3 == hc2+1)
 stopifnot(ic3 == ic2+1)
+
+# When subsequently calling the g inner function we must ensure
+# that val is properly bind. This means that we must activate a
+# the differnt SEXP everytime because val is bind to its enclosing
+# environment (the environemnt of the current activation of f).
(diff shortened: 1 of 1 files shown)


rir/tests/pir_regression.R
```diff
@@ -81,3 +81,19 @@ hc3 = .Call("rir_invocation_count", h)
...
 stopifnot(ic3 == ic2+1)
+
+# When subsequently calling the g inner function we must ensure
+# that val is properly bind. This means that we must activate a
+# the differnt SEXP everytime because val is bind to its enclosing
+# environment (the environemnt of the current activation of f).
+# This tests ensures that if an optimization tries to optimize 
+# this polymorphicness, the semantics are preserved 
+f <- function(val) {
+    g <- function() val
```
o-, line 92: this `g` is statically inlined. I don't think it would fail on your branch, does it?

charig:
```
> f <- function(val) {
+     g <- function() val
+     g()
+ }
> h <- rir.compile(function(x) f(x))
> stopifnot(h(1) == 1)
> stopifnot(h(2) == 2)
> h <- pir.compile(h)
> stopifnot(h(3) == 3)
Error: h(3) == 3 is not TRUE
```

o-: ah i see, that's another issue. we speculate before we inline. this sometimes leads to speculation where we would otherwise statically optimize.

PR #296, opened 2019-01-14
Stack allocator

JanJecmen:
just check if travis build passes...


JanJecmen:
Should travis pass, this is almost tbr... Some things left to do:

- **How to actually allocate?** for testing, I just flipped a coin on stack vs. local...

- ~better analyze which env is set at the end of a bb, so that we don't have to load it every time at the beginning of a successor bb~ implementted by #323 
- ~better handle when an env is needed after it's set, eg. if we have a setenv at the beginning and all instructions only use the one, we don't need to keep it around (and for eg. scheduled deopt we can emit getenv)~ after some thought, this is wrong.. deopt might need multiple envs
- how come mkenv is explicit and ldfunctionenv is not?
- how to pick lastInstr at the beginning of a merge block?
- ~optimize the common case of args being last use and matching the stack contents at tos, then no picks are needed, only fill in the values / locals and put those to correct offsets~ for now, just remove patterns such as double swap or n times pick(n), done
- ~jump through blocks now aren't empty, but have just one nop and no stack compensation~ done
- the alloc verifier is broken now

o-:
hey @JanJecmen , btw from your comments I find only
> the alloc verifier is broken now

critical. The rest I would say can be done later

o-:
btw. I just merged this branch (locally) with #323 and put only envs in registers. it rocks :)

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -262,6 +261,7 @@ class StaticAnalysis {
                         exitpoint = state;
                         reachedExit = true;
                     }
+                    changed[id] = false;
```
o-, line 311: good catch

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -336,7 +336,320 @@ class StaticAnalysis {
...
+
+    void logExit(const AbstractState& state) {
+        if (DEBUG_LEVEL >= AnalysisDebugLevel::Exit) {
+            log << "===== Exit state is\n";
+            log(state);
+        }
+    }
+
+    template <PositioningStyle POS>
+    AbstractState at(Instruction* i) const {
```
o-, line 526: are you using this api? if not, remove, because it is out of sync with the forward version.

rir/src/compiler/analysis/stack.cpp
```diff
@@ -0,0 +1,314 @@
...
+    // Check that the stacks have the same contents, or the differences all
+    // relate to phi inputs
+    for (size_t i = 0; i < data.size(); ++i) {
+        if (data[i] != other.data[i]) {
+            if (data[i]->isInstruction() && other.data[i]->isInstruction()) {
+                auto my = Instruction::Cast(data[i])->hasSingleUse();
+                auto their = Instruction::Cast(other.data[i])->hasSingleUse();
+                assert(my && Phi::Cast(my));
+                assert(their && Phi::Cast(their));
+                assert(my == their && "Stack mismatch.");
```
o-: why can we be sure that the stacks match?

o- (commented):
Hey, to help me review, can you give me a high-level overview?

What I got so far:

* The `StackValueAnalysis` is basically a liveness analysis. (if so, let's call it such)
* The StackAnalysis then computes the order of things on the stack.

Are there more analysis needed?

As a sidenote: your choice of storing a complete copy of the set of live variables for each instruction is a bit wasteful. The traditional choice is to have a liveness interval for each variable (I am ok with a follow up PR for fixing this though, or maybe I completely misunderstood what the StackValueAnalysis does :) )

rir/src/compiler/analysis/generic_static_analysis.h
JanJecmen: It's dumb to keep the two separate in the first place, it was supposed to be a quick temporary fix but it stuck and now is the second time I have to update :)

I guess they would be better unified, similar to the visitors..

JanJecmen:
Thanks for reviewing.

Yes, you are right, it is pretty much a liveness analysis. I realized it recently but never renamed it.
Also I realized that we already have the liveness intervals, I will try to reuse that.

Other than that, first we find out what is live at any point, then go forward and simulate the stack usage while maintaining that only one copy of every value is there and as soon as anything isn't live anymore, it gets removed.

I don't have a good answer why the stacks have to match when merging.. I noticed that they do, and added the assert to deal with it later, and it never failed in the end.
Can you think of an example where they wouldn't? Maybe some optimization could  introduce it in the future?

rir/src/compiler/analysis/generic_static_analysis.h
JanJecmen: Couldn't find a nice way to unify them... the backward is mostly copy paste of the forward :(
And what's more, I now use the liveness that we already had, so the backward analysis is not used anywhere anymore.. Remove?

JanJecmen:
- Factored out the liveness from pir2rir to a separate file, to reuse it in the stack analysis
- Got rid of the old stack values (~liveness) analysis
- Updated the backward analysis
- Fixed a bug with buffering the argument loads (could add source reference to a wrong bytecode) (kind of hacky now)
- Found and fixed (?) a bug with the analysis cache: instructions could be queued multiple times and then looking for them would dereference the map's `end()`

not sure it got better or worse since the last iteration :)

o-:
> I don't have a good answer why the stacks have to match when merging.. I noticed that they do, and added the assert to deal with it later, and it never failed in the end.

I think the copy insertion is what safes you.

I tried to break it with `if (a) { x <- 1; y <- 2} else {y <- 3; x <- 4}; x-y}`

And indeed at first it looks problematic:

```
BB2
  real$'  %2.0  = LdConst          [1] 1 
  real$'  %2.3  = LdConst          [1] 2 
  goto BB3
BB1
  real$'  %1.0  = LdConst          [1] 3 
  real$'  %1.3  = LdConst          [1] 4 
  goto BB3
BB3
  real$'  %3.0  = Phi              %1.0:BB1, %2.3:BB2
  real$'  %3.1  = Phi              %1.3:BB1, %2.0:BB2
  val?'   %3.2  = Sub              %3.1, %3.0, elided
```

But then after cssa lowering we have:

```
BB2
  real$'  %2.0  = LdConst          [1] 1 
  real$'  %2.3  = LdConst          [1] 2 
  real$'  %2.5  = %2.3
  real$'  %2.6  = %2.0
  goto BB3
BB1
  real$'  %1.0  = LdConst          [1] 3 
  real$'  %1.3  = LdConst          [1] 4 
  real$'  %1.5  = %1.0
  real$'  %1.6  = %1.3
  goto BB3
BB3
  real$'  %3.0  = Phi              %1.5:BB1, %2.5:BB2
  real$'  %3.1  = %3.0
  real$'  %3.2  = Phi              %1.6:BB1, %2.6:BB2
  real$'  %3.3  = %3.2
  val?'   %3.4  = Sub              %3.3, %3.1, elided
```

Magically the inputs are equalized :) The reason is that we insert the copies for each phi, one after the other.

PR #301, opened 2019-01-17
Nextntext

o-:
sorry, the history of this branch is beyond repair.

The major changes are:

* split pir::Closure into Closure and ClosureVersion. The closure is then a list of different versions optimized under different OptimizationContexts
* a first iteration of contextual dispatching: a call instruction has some givenAssumptions, and each version has some requiredAssumptions. Statically the dispatching is conservative (in Instruction::dispatch). Dynamically the dispatching adds runtime information (in interp.cpp::dispatch).
* all of our call instructions used in pir are now dispatching!
* A new properties field, that allows us to remember properties we figured out about ClosureVersions. For example if ForceDominance finds out that a function behaves like an eager function, we set the eager property.
* An eagerCall optimization, that looks for calls to eager functions and eagerly evaluates arguments for them.
* Earlier switch to optimized code: add adapter code that allows us to call directly from R interpreter to PIR-optimized code (instead of requiring unoptimized middle-function). This made much more code run in optimized version and thus revealed some small bugs.
* Support for recursive calls in the analysis framework.
* Small improvements in scope analysis, like better estimation which environment a ldfun could potentially taint.
* Several refactorings and cleanups


rir/src/runtime/Function.h
```diff
@@ -74,6 +74,7 @@ struct Function : public RirRuntimeObject<Function, FUNCTION_MAGIC> {
 
     unsigned deopt : 1;
     unsigned markOpt : 1;
+    unsigned unoptimizable : 1;
```
JanJecmen, line 78: `spare` should be 29 now

o-: hah, thanks, let's just remove that

rir/src/compiler/analysis/abstract_value.h
```diff
@@ -15,14 +15,17 @@ namespace pir {
 struct ValOrig {
     Value* val;
     Instruction* origin;
-    ValOrig(Value* v, Instruction* o) : val(v), origin(o) {}
+    unsigned recursionLevel;
```
charig, line 18: I don't clearly see why for the values we need the instruction

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -0,0 +1,117 @@
...
+            bool noMissing = true;
+            call->eachCallArg([&](Value* v) {
+                if (auto mk = MkArg::Cast(v)) {
+                    if (mk->eagerArg() == Missing::instance())
+                        args.insert(mk);
+                } else {
+                    noMissing = false;
+                }
+            });
+            if (!noMissing)
```
charig, line 45: It may be easier to change noMissing for missing and avoid the double negation that is a little confusing

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -0,0 +1,117 @@
...
+                continue;
+            todo.insert(args.begin(), args.end());
+
+            if (version->assumptions().includes(Assumption::EagerArgs_))
```
charig, line 49: I'd do this check after call->dispatch()

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -0,0 +1,117 @@
...
+                        }
+                    });
+                });
+            call->hint = newVersion;
+        }
+    });
+    if (todo.empty())
+        return;
+
+    Visitor::run(code, [&](BB* bb) {
```
charig, line 77: Some comments on this one. 
* Since we are using this lambda pattern and then we do not have a name for the main code, I would add a comment to a mehotd like this.
* I do not get the first condition. It inlines the promise if the analysis finds it is a missing? Why the inlining happens here and not in the pass that actually inline promises?
* The second condition sth similiar. If the closure do not use reflection you already set the value of the argument if it is eager. Ok, but why here?

charig (approved):

PR #303, opened 2019-01-18
Lazy context's promargs

charig:
Do not create argslists (needed by gnu-r contexts) when we may not need them. For instance on eager calls. Instead, we put in the corresponding field a wrapper with all the necessary data to lazily create them in case later gnu-r triggers and needs them (for instance for S3 dispatching). 

I am not sure how much it pays off. We are creating the `c wrappers` instead of the `R SEXP`. There we should be faster. But then, the change needs some tests here and there (on every call) to check if the field is actually an argslists or a wrapper. 

If it finally pays off, a second iteration could go through those.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -305,12 +305,9 @@ SEXP createLegacyArgsList(const CallContext& call, bool eagerCallee,
     return result;
 }
 
-SEXP createLegacyLazyArgsList(const CallContext& call, Context* ctx) {
-    if (call.hasStackArgs()) {
-        return createLegacyArgsListFromStackValues(call, false, ctx);
-    } else {
-        return createLegacyArgsList(call, false, ctx);
-    }
+SEXP argsLazyCreation(void* rirDataWrapper) {
+    ArgsLazyData* argsLazy = ArgsLazyData::unpack(rirDataWrapper);
+    return argsLazy->createArgsLists();
 }
 
 SEXP createLegacyArgsList(const CallContext& call, Context* ctx) {
@@ -623,8 +620,6 @@ SEXP rirCall(const CallContext& call, Context* ctx) {
 
     Function* fun = dispatch(call, table);
     fun->registerInvocation();
-    bool needsEnv = fun->signature().envCreation ==
-                    FunctionSignature::Environment::CallerProvided;
 
     if (fun->signature().optimization ==
             FunctionSignature::OptimizationLevel::Baseline &&
@@ -635,29 +630,32 @@ SEXP rirCall(const CallContext& call, Context* ctx) {
             name = lhs;
         ctx->closureOptimizer(call.callee, name);
         fun = dispatch(call, table);
-        needsEnv = fun->signature().envCreation ==
-                   FunctionSignature::Environment::CallerProvided;
     }
 
+    bool needsEnv = fun->signature().envCreation ==
+                    FunctionSignature::Environment::CallerProvided;
     SEXP env = R_NilValue;
-
     SEXP result = nullptr;
     if (needsEnv) {
-        auto arglist = createLegacyLazyArgsList(call, ctx);
+        auto arglist = createLegacyArgsList(call, ctx);
         PROTECT(arglist);
         env = closureArgumentAdaptor(call, arglist, R_NilValue);
         PROTECT(env);
         result = rirCallTrampoline(call, fun, env, arglist, ctx);
         UNPROTECT(2);
     } else {
-        auto arglist = createLegacyLazyArgsList(call, ctx);
-        PROTECT(arglist);
-        // TODO: find a way to lazily create the argslist,
-        // only when needed. Afaik, the argslist is only accessed ever in
-        // do_usemethod, therefore it should be possible to have our own
-        // do_usemethod, that lazily creates the argslist when needed.
-        result = rirCallTrampoline(call, fun, arglist, ctx);
-        UNPROTECT(1);
+        if (call.hasStackArgs()) {
+            // Instead of a SEXP with the argslist we create an
+            // structure with the information needed to recreate
+            // the list lazily if the gnu-r interpreter needs it
+            ArgsLazyData lazyArgs = ArgsLazyData(&call, ctx);
+            result = rirCallTrampoline(call, fun, (SEXP)&lazyArgs, ctx);
+        } else {
+            auto arglist = createLegacyArgsList(call, ctx);
+            PROTECT(arglist);
+            result = rirCallTrampoline(call, fun, arglist, ctx);
+            UNPROTECT(1);
+        }
     }
 
     assert(result);
--- a/rir/src/runtime/RirDataWrapper.h
+++ b/rir/src/runtime/RirDataWrapper.h
@@ -1,0 +1,48 @@
+#ifndef RIR_DATA_WRAPPER_H
+#define RIR_DATA_WRAPPER_H
+
+/** RirDataWrappers are wrappers (usually over SEXP) of relevant
+ * RIR level information that must be shared to the gnu-r
+ * interpreter. For instance for the interpreter to lazily
+ * create stuff depending on RIR information
+ */
+
+#include "R/r.h"
+#include <cassert>
+#include <cstdint>
+
+namespace rir {
+
+struct data_header {
+    // magic number to differentiate RIR objects
+    // needs to be in sync with gnur context.c
+    uint32_t magic;
+};
+
+template <typename BASE, uint32_t MAGIC>
+struct RirDataWrapper {
+    data_header info;
+
+    static BASE* check(SEXP s) {
+        BASE* b = (BASE*)s;
+        return b->info.magic == MAGIC ? b : nullptr;
+    }
+
+    static BASE* unpack(void* s) {
+        BASE* b = (BASE*)s;
+        assert(
+            b->info.magic == MAGIC &&
+            "Trying to unpack the wrong type of embedded RIR runtime object.");
+        return b;
+    }
+
+  protected:
+    RirDataWrapper(uint32_t count) : info{MAGIC} {
+        uint8_t* start = (uint8_t*)this + sizeof(uint32_t);
+        memset(start, 0, count * sizeof(void*));
+    }
+};
+
+} // namespace rir
+
+#endif
--- a/rir/src/interpreter/interp.h
+++ b/rir/src/interpreter/interp.h
@@ -23,10 +23,10 @@ SEXP rirExpr(SEXP f);
 
 SEXP rirEval_f(SEXP f, SEXP env);
 
+SEXP argsLazyCreation(void* rirDataWrapper);
+
 #ifdef __cplusplus
 }
 #endif
 
 #endif // RIR_INTERPRETER_C_H
-
-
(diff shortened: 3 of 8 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -305,9 +305,19 @@ SEXP createLegacyArgsList(const CallContext& call, bool eagerCallee,
...
 }
 
-SEXP createLegacyLazyArgsList(const CallContext& call, Context* ctx) {
+SEXP argsLazyCreation(void* rirDataWrapper) {
+    ArgsLazyData* argsLazy = ArgsLazyData::unpack(rirDataWrapper);
+    return argsLazy->createArgsLists();
+}
+
+SEXP createLegacyLazyArgsList(const CallContext& call, Context* ctx,
+                              bool needsEnv) {
```
o-: this is confusing to read. please remove needsEnv and make the decision to call createLegacyLazyArgsList vs. createLegacyArgsList at the caller's side

rir/src/interpreter/interp.cpp
```diff
@@ -305,9 +305,19 @@ SEXP createLegacyArgsList(const CallContext& call, bool eagerCallee,
...
     if (call.hasStackArgs()) {
-        return createLegacyArgsListFromStackValues(call, false, ctx);
+        if (needsEnv)
+            return createLegacyArgsListFromStackValues(call, false, ctx);
+        else
+            // double case to avoid alignment warning
+            return (SEXP)(void*)(new ArgsLazyData(&call, ctx));
```
o-: we can't put C++ objects into R datastructures. This can crash R in a myriad of ways. Only SEXP from the R GC are allowed.

o-: See https://github.com/reactorlabs/rir/blob/master/rir/src/utils/FunctionWriter.h#L49 and the RirRuntimeObject of how this can be done https://github.com/reactorlabs/rir/blob/master/rir/src/runtime/RirRuntimeObject.h#L38. The idea is that we have a layered object layout. From R's pov it is a SEXP of type EXTERNALSXP. The payload is then some subclass of RirRuntimeObject. The RirRuntimeObject has a secondary header that tells the R GC if there are pointers inside it, that need to be followed (but there are none in your case...). Other than that the R GC does not look inside the payload. RIR uses placement new to put the RIR object into the payload part. The RirRuntimeObject provides `container` and `unpack` functions to get from SEXP to rir object and back.

o- (changes requested):
Hey, some comments:

1. no C++ obj on the R heap (see comments inline)
2. the gnur patch is really messy (lots of whiteline changes). Please when making changes to gnuR, they should not touch a single line that does not need patching. We need to forward port those patches to new R versions and it is super painful to deal with completely avoidable merge conflicts there.
3. I put a breakpoint in interp.cpp at `SEXP argsLazyCreation(void* rirDataWrapper)` and it never triggers in our testsuite. This is completely untested code.
4. It does not seem to work. Here is a repro:
```
f <- function(a,b,c) nargs()
g <- function() {
  f() 
  f(1)
  f(1,2)
  f(1,2,3)
}
h <- function() g() 
stopifnot(h()==3)
stopifnot(h()==3)
stopifnot(h()==3)
stopifnot(h()==3)
stopifnot(h()==3)
stopifnot(h()==3)
stopifnot(h()==3)
```

rir/src/interpreter/interp.cpp
charig: Oh, this was totally on purpose. I know about `RirRuntimeObjects`. I created `ArgsLazyData` based on them: https://github.com/charig/rir/blob/lazyArgs/rir/src/runtime/ArgsLazyData.h.

The main difference is that they are not wrapped inside a `SEXP`. My point here is that the main goal of this PR is to avoid creating an unnecessary `SEXP` on every function call for the arguments. If we avoid that but need to create another SEXP for wrapping the data that is needed to lazily create the SEXP, and need to add some checks here and there to test what kind of argsList we really created, then what do we gain?
I took some hours to read the gnu-r code and it seemed that the RCNTXT are managed in ad-hoc manners (even by the gc), i.e, https://github.com/reactorlabs/gnur/blob/lazyPromargs/src/main/memory.c#L1693.

I resort on the gnu-r test suite that triggers some GCs for testing. Anyway, if this is not safe it would be very easy for me to make `ArgsLazyData` to subclassify `RirRuntimeObjects` instead of `RirDataWrapper`. But then I guess the PR lose much of its value.

charig:
> Hey, some comments:
> 
> 1. no C++ obj on the R heap (see comments inline)
> 2. the gnur patch is really messy (lots of whiteline changes). Please when making changes to gnuR, they should not touch a single line that does not need patching. We need to forward port those patches to new R versions and it is super painful to deal with completely avoidable merge conflicts there.
> 3. I put a breakpoint in interp.cpp at `SEXP argsLazyCreation(void* rirDataWrapper)` and it never triggers in our testsuite. This is completely untested code.
> 4. It does not seem to work. Here is a repro:
> 
> ```
> f <- function(a,b,c) nargs()
> g <- function() {
>   f() 
>   f(1)
>   f(1,2)
>   f(1,2,3)
> }
> h <- function() g() 
> stopifnot(h()==3)
> stopifnot(h()==3)
> stopifnot(h()==3)
> stopifnot(h()==3)
> stopifnot(h()==3)
> stopifnot(h()==3)
> stopifnot(h()==3)
> ```

True, there was not pir tests covering the case of a call with arguments on the stack that does not need environments. I resort to gnu-r test suite for the PR sorry. The test you propose covers this case but I checked and it was not failing because of the changes of this PR. PR #304 solves the issue and integrating #304 with this PR everything works. So, I resort on integrating #304 for the test addition.

rir/src/runtime/RirDataWrapper.h
```diff
@@ -0,0 +1,51 @@
...
+#include "R/r.h"
+#include <cassert>
+#include <cstdint>
+
+namespace rir {
+
+struct data_header {
+    // number of pointers for the required data.
+    // Maybe this could be another template parameter?
+    uint32_t data_count;
```
o-: unused variable

rir/src/interpreter/interp.cpp
o-: Ok I see. Though, it would still help, since it reduces the allocations from the whole argslist to just one cell.

Fair enough, I am conviced that we can get away with not allocating an R object, but just an internal marker thingy.

Some additional comments though:

* [this](https://github.com/reactorlabs/rir/pull/303/files#diff-48bddc5b82508439eaadd23792d2851bR664) is wrong, since it might be replaced by an actual SEXP during the call. Why not use stack allocated memory in the first place? In doCall just do:

```
ArgsLazyData lazyArgRecreateData(call, ctx);
arglist = &lazyArgRecreateData;
```

* [this](https://github.com/reactorlabs/rir/pull/303/files#diff-48bddc5b82508439eaadd23792d2851bR315) is not a good idea. Imho the warning is real. You should probably remove the pragma pack stuff instead.

* [this](https://github.com/reactorlabs/gnur/blob/lazyPromargs/src/main/context.c#L121) looks brittle. Could you at least put the marker at the beginning of the struct? And then on the rir side you could add a static assert. Something like:

```
RirDataWrapper(uint32_t count) : info{count, MAGIC} {
    //needs to be in sync with gnur context.c
   static_assert(&this->magic == &((utint32_t*)this));
}
```

o-:
meh, there is some bad interaction with #304 

1. the argslist should be created without the missing args.
2. the argslist should be created with user supplied missing args. (ie. you do `foo(1,,)` thats `1` and two missing args)

when we call `function(a,b,c)` with `f()`, or `f(,,)` then that is observably different. In one case the argslist is empty, in the second case the argslist has 3 missing args.

That's why #304 creates the argslist *before* stuffing the missing args. But now in this PR, the argslist will be created *after* stuffing the args.

when we [modify](https://github.com/reactorlabs/rir/pull/304/files#diff-48bddc5b82508439eaadd23792d2851bR600) nargs, we should probably remember the old value and when we recreate the argslist, we should not put those additional R_MissingArgs on the argslist.

Do you want me to add the support to #304, or will you add it here?

rir/src/interpreter/interp.cpp
charig: > Ok I see. Though, it would still help, since it reduces the allocations from the whole argslist to just one cell.
> 
> Fair enough, I am conviced that we can get away with not allocating an R object, but just an internal marker thingy.
> 
> Some additional comments though:
> 
> * [this](https://github.com/reactorlabs/rir/pull/303/files#diff-48bddc5b82508439eaadd23792d2851bR664) is wrong, since it might be replaced by an actual SEXP during the call. Why not use stack allocated memory in the first place? In doCall just do:
> 
> ```
> ArgsLazyData lazyArgRecreateData(call, ctx);
> arglist = &lazyArgRecreateData;
> ```

It is not. argsList always points to the `ArgsLazy` wrapper. The field that is actually replaced in place in case we need the promargs is the RCNTXT->promargs. I can not do what you suggest in `doCall` because at that point I do not know what kind of argsList I would need. To avoid manual memory and leverage on the stack `rirCall` needs a major refactoring to abstract the argList creation for anycase. 
  

> * [this](https://github.com/reactorlabs/rir/pull/303/files#diff-48bddc5b82508439eaadd23792d2851bR315) is not a good idea. Imho the warning is real. You should probably remove the pragma pack stuff instead.
> * [this](https://github.com/reactorlabs/gnur/blob/lazyPromargs/src/main/context.c#L121) looks brittle. Could you at least put the marker at the beginning of the struct? And then on the rir side you could add a static assert. Something like:
> 
> ```
> RirDataWrapper(uint32_t count) : info{count, MAGIC} {
>     //needs to be in sync with gnur context.c
>    static_assert(&this->magic == &((utint32_t*)this));
> }
> ```

charig:
> meh, there is some bad interaction with #304
> 
> 1. the argslist should be created without the missing args.
> 2. the argslist should be created with user supplied missing args. (ie. you do `foo(1,,)` thats `1` and two missing args)
> 
> when we call `function(a,b,c)` with `f()`, or `f(,,)` then that is observably different. In one case the argslist is empty, in the second case the argslist has 3 missing args.
> 
> That's why #304 creates the argslist _before_ stuffing the missing args. But now in this PR, the argslist will be created _after_ stuffing the args.
> 
> when we [modify](https://github.com/reactorlabs/rir/pull/304/files#diff-48bddc5b82508439eaadd23792d2851bR600) nargs, we should probably remember the old value and when we recreate the argslist, we should not put those additional R_MissingArgs on the argslist.
> 
> Do you want me to add the support to #304, or will you add it here?

From what I understand it is completely orthogonal to this PR? Why isn't any test failing in 304?

rir/src/interpreter/interp.cpp
o-: > It is not. argsList always points to the ArgsLazy wrapper.

my bad, sorry, you are right

> I can not do what you suggest in doCall

sure, just put `ArgsLazyData lazyArgRecreateData(call, ctx);` below line 655 and only use it in one branch

o-: (btw. the problem with new/free is also, that we will leak when a non-local return jumps over the delete)

o-:
> From what I understand it is completely orthogonal to this PR?

no. this sentence:

> That's why #304 creates the argslist before stuffing the missing args. But now in this PR, the argslist will be created after stuffing the args.

Yes sorry I was looking at your first commit only

o-:
included in #305

PR #304, opened 2019-01-21
fix call.nargs if dispatch pushes missing args

o-:
during PIR dispatch we explicitly materialize R_MissingArg on the
stack, if the callee does not provide enough arguments. But this
needs to be reflected in the call.nargs field. Because, for example
we need to remove the missing args again, or we need to inform
the callee that the missing args are there.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -60,7 +60,7 @@ struct CallContext {
                       callerEnv, ctx) {}
 
     const Code* caller;
-    const size_t nargs;
+    size_t nargs;
     const R_bcstack_t* stackArgs;
     const Immediate* implicitArgs;
     const Immediate* names;
@@ -549,7 +549,8 @@ SEXP closureArgumentAdaptor(const CallContext& call, SEXP arglist,
     return newrho;
 };
 
-bool matches(const CallContext& call, const FunctionSignature& signature) {
+RIR_INLINE bool matches(const CallContext& call,
+                        const FunctionSignature& signature) {
     // TODO: look at the arguments of the function signature and not just at the
     // global assumptions list. This only becomes relevant as soon as we want to
     // optimize based on argument types.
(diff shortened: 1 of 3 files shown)


o-:
@charig while looking at your PR I found this additional bug, that might interfere with what you are up to

rir/src/interpreter/interp.cpp
```diff
@@ -590,12 +590,13 @@ bool matches(const CallContext& call, const FunctionSignature& signature) {
         // explicitly mentioned on the stack
         for (size_t i = 0; i < signature.nargs() - call.nargs; ++i)
             ostack_push(ctx, R_MissingArg);
+        call.nargs += signature.nargs() - call.nargs;
```
charig: Why not just `call.nargs = signature.nargs()`

rir/src/interpreter/interp.cpp
```diff
@@ -1514,11 +1515,11 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
             CallContext call(c, ostack_at(ctx, n), n, ast,
                              ostack_cell_at(ctx, n - 1), getenv(), ctx);
             res = doCall(call, ctx);
-            ostack_popn(ctx, n + 1);
+            ostack_popn(ctx, call.nargs + 1);
```
charig, line 1525: Why don't we use `call.nargs` or `n` in both cases?

charig (approved):

rir/src/interpreter/interp.cpp
o-: n is the supplied number of args. call.nargs is the actual number (after fixed up missings). so before the call we have n on the stack, after the call we have call.nargs on the stack.

charig: `n` is read from the bytecode. It is the number of args the function expects. And call.nargs is finally patched so that it contains the actual number of arguments too (previously it contained only the actual syntactic arguments passed). So If I understand this correctly they are the same and it is confusing to pop for `call.nargs` but assert on `n`. If I am wrong in the premise just ignore the comment and merge.

o-:
```
n = how many args did the user supply
call.nargs = now many args are effectively passed to the function
fun->signature().nargs = how many args does the function want
```

* if we stuffed the passed arguments with missing, then `n <= call.nargs == fun->signature().nargs`.
* if we did not stuff passed arguments with missing, because they are not passed on the stack, then `n == call.nargs <- fun->signature().nargs`.

o-:
included in #305

PR #306, opened 2019-01-23
 make pir typesystem differ between values and promise wrapped values

o-:
based on #301 

A R value can be:

1. an unevaluated promise
2. an evaluated promise
3. a value

In the past we did not differ between 2 and 3 which can lead to
confusing situations and easily leads to a situation where we
put the wrong kind of value in the wrong place.

This commit adds a new type flag `promiseWrapped`.

Therefore:

1. maybeLazy && maybePromiseWrapped
2. !maybeLazy && maybePromiseWrapped
3. !maybeLazy && !maybePromiseWrapped

To transition 2 => 3 a Force instruction is always needed, because
many internal instruction expect either one or the other.

From 1 => 2 it is enough to prove that a value is already evaluated
and then remove the maybeLazy flag.

Evaluated promises are marked with the type flag `~` in the textual format.

### DIFF
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -138,7 +138,8 @@ AbstractResult ScopeAnalysis::apply(ScopeAnalysisState& state,
         auto force = Force::Cast(i);
         auto arg = force->arg<0>().val();
         if (!arg->type.maybeLazy()) {
-            effect.max(state.returnValues[i].merge(ValOrig(arg, i, depth)));
+            if (!arg->type.maybePromiseWrapped())
+                effect.max(state.returnValues[i].merge(ValOrig(arg, i, depth)));
             handled = true;
         }
 
@@ -146,7 +147,9 @@ AbstractResult ScopeAnalysis::apply(ScopeAnalysisState& state,
             lookup(state, arg->followCastsAndForce(),
                    [&](const AbstractPirValue& analysisRes) {
                        if (!analysisRes.type.maybeLazy()) {
-                           effect.max(state.returnValues[i].merge(analysisRes));
+                           if (!analysisRes.type.maybePromiseWrapped())
+                               effect.max(
+                                   state.returnValues[i].merge(analysisRes));
                            handled = true;
                        } else if (analysisRes.isSingleValue()) {
                            arg = analysisRes.singleValue().val;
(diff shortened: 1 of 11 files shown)


rir/src/compiler/analysis/scope.cpp
```diff
@@ -138,15 +138,18 @@ AbstractResult ScopeAnalysis::apply(ScopeAnalysisState& state,
         auto force = Force::Cast(i);
         auto arg = force->arg<0>().val();
         if (!arg->type.maybeLazy()) {
-            effect.max(state.returnValues[i].merge(ValOrig(arg, i, depth)));
+            if (!arg->type.maybePromiseWrapped())
```
charig, line 141: Only `!maybePromiseWrapped` is enough

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -911,9 +908,15 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
 
             case Tag::CastType: {
                 auto cast = CastType::Cast(instr);
```
charig, line 910: What are casts for then?

charig (approved):
I do not see in this PR any place that suggests the new type is needed.
Maybe a reference to a PR using it or a small comment?

o-:
> I do not see in this PR any place that suggests the new type is needed.

The most useful case right now are eager functions. Basically [this](https://github.com/reactorlabs/rir/pull/306/files#diff-d9e8b8c3c395dfef187a13fed00361e6R95) is a bugfix. Because before this PR, the callee assumes that the argument is a value (without promise) and the caller passes an evaluated promise.

This works as long as we pass the argument to builtins or function that check if the arg is a promise (some do), but fails as soon we pass it to something else. For example `function(a) a==1` would fail, since it would compare the promise with value 1 against the value 1.

o-:
> What are casts for then?

Currently we use the cast mainly [here](https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/opt/inline.cpp#L171) to cast from `prom` to `val?`.

Physically a `prom` and a `val?` are the same thing. But semantically a `prom` is a promise-as-a-value. For example a `call` instruction takes `prom` as arguments. The idea is that you want to consume the promise. On the other hand `val?` is a promised-value. Basically you are referring to the value inside the promise, but it might not yet be computed. For example you can have an `int?` which is an integer, that might have to be computed first.

Now, a call instruction takes as argument `prom`. On the other hand a ldarg instruction produces a `val?`. Basically calling a function converts a promise-as-the-value to a promised-value.

When we inline this conversion disappears, so to restore sanity we cast from one type to the other.

rir/src/compiler/analysis/scope.cpp
o-: the second statement is executed for both

PR #309, opened 2019-01-24
Default args

o-:
based on #306

implements missing args as assumptions: only support those cases where from the contextual assumptions we know exactly which arguments are passed and which ones are missing. Using that information we then add the missing arguments in the pir functions by explicitly creating a promise of the default args.

Creating default args (the core of this PR) is implemented in rir_2_pir_compiler.cpp:112

Then there are a number of issues I had to resolve to get it working:

* I added the `missing` instruction to pir, that is used to check if an argument is missing. Warning: default args are only evaluated if the argument is immediately missing. But the missing instruction also returns true if the argument is transitively missing.

* To know if an argument was missing, before it was set to a default argument, R sets the MISSING flag on the binding. For example `function (a=1) c(missing(a), a)` when called with no args, returns `TRUE, 1`. Thus it is important to preserve that missing flag, when setting `a` to the default argument `1`. I implemented this with a new rir instruction `starg`, which (in contrast to `stvar`) does not reset the missing flag.

* Kind of the dual to the above: Pir needs to be able to set args as missing, eg. if we compile a function with 2 formal args, but under the assumption that one arg is passed. If we would compile `MkEnv(a=missing)` to `mkenv; stvar "a", missing` as we used to, then the missing flag is not set. Therefore I changed the rir `mkenv` instruction to create the whole env (like the pir one) and accordingly set the missing flags on the bindings. As a nice sideeffect this make the rir version of mkenv a lot faster.

* The `maybeMissing` semantics of the pir type was completely broken. I removed that flag. Instead there is now an `RType::missing`, which represents the type of `R_MissingValue`.

* Dead store elimination did not consider some instructions that would read the environment.

* The order of the entries in the dispatch table was wrong, thus we would recompile already compiled versions.

* We conflated `R_MissingArg` (ie. an argument was omited when calling a function) and `R_UnboundValue` (ie. the placeholder value when a promise is not evaluated).

* Non-instruction values as phi inputs broke the register allocator.

* The inliner did not check the promises of the inlinee for the blacklist of reflective context accessing function calls.

* Call implicit is bad for rir, because it does not allow us to supply missing args on the stack. I removed it, pir does not produce it anymore.

* Pir just deleted the `visible`/`invisible` instructions from rir. We need to preserve them. (TODO: remove redundant ones)


rir/src/runtime/Assumptions.h
```diff
@@ -31,21 +30,93 @@ enum class Assumption {
...
 std::ostream& operator<<(std::ostream& out, Assumption a);
 
 } // namespace rir
 
+namespace std {
+template <>
+struct hash<rir::Assumptions> {
+    std::size_t operator()(const rir::Assumptions& v) const {
+        using std::hash;
+        return hash<unsigned long long>()(v.flags.to_i()) ^
```
mhyee: Minor, but I don't know if this is the best way to combine the hash of two ints? Feels like there might be a bunch of collisions.

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -95,7 +95,6 @@ class TheCleanup {
                 // promise deletion would be broken, if we were to use
                 // CallImplicit in the middle-end, since this instruction can
```
mhyee: This comment is now outdated.

mhyee: If `CallImplicit` is gone from PIR, why do we still have `callImplicit` for RIR? What's the difference?

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -103,6 +113,45 @@ void Rir2PirCompiler::compileClosure(Closure* closure,
...
+    for (unsigned i = closure->nargs() - assumptions.numMissing();
+         i < closure->nargs(); ++i) {
+        if (closure->formals().hasDefaultArgs()) {
+            auto arg = closure->formals().defaultArgs()[i];
+            if (arg != R_MissingArg) {
+                Value* res = nullptr;
+                if (TYPEOF(arg) != EXTERNALSXP) {
+                    // A bit of a hack to compile default args, which somehow
+                    // are not compiled.
+                    // TODO: why are they sometimes not compiled??
```
mhyee, line 130: Are they _sometimes_ compiled or _never_ compiled? (Sometimes would be weirder than never.)

rir/tests/rir_default.R
```diff
@@ -27,3 +27,22 @@ rir.compile(function() {
     stopifnot(f(2) == c(2,2))
     stopifnot(f(,1) == c(1,1))
 })()
+
+
+f <- function(a=1,b=2,c=3) c(a,b,c,missing(a), missing(b), missing(c), nargs())
+g <- function(a) c(missing(a), nargs())
+h <- function(a) g(a)
```
mhyee, line 34: Is this `h` function being tested?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -158,12 +158,14 @@ Instruction* Instruction::hasSingleUse() {
 void Instruction::eraseAndRemove() { bb()->remove(this); }
 
 void Instruction::replaceUsesIn(Value* replace, BB* target) {
-    if (!replace->type.isA(type)) {
+    if (replace->type.isRType() != type.isRType() ||
+        (replace->type.maybePromiseWrapped() && !type.maybePromiseWrapped())) {
```
mhyee, line 162: Is it

`replace->type.maybePromiseWrapped() && !type.maybePromiseWrapped()`

or

`replace->type.maybePromiseWrapped() != type.maybePromiseWrapped()`

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    /* skip first stack frame (points here) */
+    for (int i = 1; i < size && messages != NULL; ++i) {
+        std::string msg = messages[i];
+        auto namestart = msg.find("(");
+        auto nameend = msg.find(")");
+        if (namestart == std::string::npos || nameend == std::string::npos) {
+            std::cerr << msg << "\n";
+        } else {
+            auto name = msg.substr(namestart + 1, nameend - namestart - 2);
```
mhyee: Shouldn't the length be `nameend - namestart - 1`?

The length of the substring, including `(` and `)` is `nameend - namestart + 1`.

```
(abcde)
0123456

namestart = 0
nameend   = 6

start index is 1 = namestart + 1
length is      5 = nameend - namestart - 1
```

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <unistd.h>
+
+#undef assert
+
+void printCBacktrace() {
+    std::cerr << "\nC Backtrace\n\n";
+    void* array[128];
+
+    // get void*'s for all entries on the stack
+    int size = backtrace(array, 128);
```
mhyee, line 22: So I wasn't familiar with `backtrace` or `execinfo.h`, so a comment pointing to the man pages would be helpful.

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    char** messages = backtrace_symbols(array, size);
+
+    /* skip first stack frame (points here) */
+    for (int i = 1; i < size && messages != NULL; ++i) {
+        std::string msg = messages[i];
```
mhyee, line 30: It would also help to have an example of the backtrace text, to better see what the string parsing is doing.

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+    auto cleanupToS = [&](SEXP v, size_t l = 32) {
+        rir::CaptureOut c;
+        Rf_PrintValue(v);
+        auto r = c();
+        r.pop_back();
+        std::string::size_type pos = 0;
+        while ((pos = r.find("\n", pos)) != std::string::npos) {
+            r[pos] = ' ';
+        }
+        if (r.length() > l)
```
mhyee: The `l` looks too much like a `1` so I was confused.

~~inb4 get a better font~~

mhyee: For that matter, the other one-character names are too terse. v -> value, l -> len, c -> captured, r -> result

`c` was funny because I saw `c()` and thought of the R function...

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+            auto realname = abi::__cxa_demangle(name.c_str(), 0, 0, &status);
+            std::cerr << realname << "+" << offset << "\n";
+            free(realname);
+        }
+    }
+
+    free(messages);
+}
+
+void printRBacktrace() {
```
mhyee, line 56: An explanation of what's going on here would be helpful.

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+                    std::cerr << CHAR(PRINTNAME(a.tag())) << "=";
+                auto arg = *a;
+
+                if (TYPEOF(arg) == PROMSXP && PRVALUE(arg) != R_UnboundValue) {
+                    arg = PRVALUE(arg);
+                }
+                if (TYPEOF(arg) == PROMSXP) {
+                    std::cerr << arg;
+                } else {
+                    std::cerr << "'" << cleanupToS(arg) << "'";
```
mhyee: What is `cleanupToS` doing (and what does it even stand for? my first thought was "top of stack") and why does the promise arg skip this step?

rir/src/common.cpp
```diff
@@ -0,0 +1,121 @@
...
+                }
+                if (a + 1 != args.end())
+                    std::cerr << ", ";
+            }
+        }
+
+        std::cerr << "\n  ENV: ";
+        auto args = rir::RList(FRAME(ctx->cloenv));
```
mhyee, line 102: This chunk of code is basically a duplicate of the one above. But YAGNI / rule-of-three, so I'm fine with just leaving this alone and punting on the refactoring.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -292,9 +292,6 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
             if (argi == DOTS_ARG_IDX) {
                 log.warn("Cannot compile call with ... arguments");
                 return false;
-            } else if (argi == MISSING_ARG_IDX) {
```
mhyee, line 262: Need to update the TODO above.

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -71,12 +71,33 @@ class TheScopeResolution {
                     continue;
                 }
 
+                // Constand fold "missing" if we can
```
mhyee: Type: Constand -> constant

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -71,12 +71,33 @@ class TheScopeResolution {
...
-                        missing->replaceUsesAndSwapWith(
-                            new LdConst(R_FalseValue), ip);
+                        // Missing still returns TRUE, if the argument was
+                        // initially missing, but then overwritten by a default
+                        // argument. Currently our analysis cannot really
+                        // distinguish those cases. Therefore, if the current
+                        // value of the variable is guaranteed to not be a
+                        // missing value, we additionally need verify that the
+                        // initial argument (the argument to the mkenv) was also
+                        // guaranteed to not a missing value.
```
mhyee: Nitpick: guaranteed to not **be** a missing value.

rir/src/compiler/opt/cleanup.cpp
o-: > If CallImplicit is gone from PIR, why do we still have callImplicit for RIR? What's the difference?

The baseline compiler produces only call implicit. We would have to measure how bad it is to remove it there. The problem is that the alternative is to have a bunch for `promise_` instructions to set up the arguments, and that might be significantly slower for code that stays in baseline...

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
o-: sometimes...

rir/src/compiler/pir/instruction.cpp
o-: the former. I want to assert that we are not replacing a promise with a non-promise

PR #310, opened 2019-01-24
Bench support

charig:


JanJecmen:
I love that `+140 −191,552` summary ;-D

o-:
awesome. one last comment: the runtime was now 47 minutes, which is super close to the max runtime of 50'. can we have even less iterations?

mhyee:
It looks like the number of iterations is 5? How low can we go?

If the Travis log is accurate, the benchmarks are taking 30 minutes to run. Maybe most of the time is being taken up by setup?

PR #311, opened 2019-01-24
New opcode specialized for Subassign2d

Jakobeha:
I added a new opcode for assigning an item in a 2D array.

Before, instructions like `a[b,c] <- d` and `a[[b,c]] <- d`, when compiled, would call `<-` and `[`/`[[` directly. Now, they both use this opcode.

The opcode should make these instructions faster in certain cases. For example, in `a[[b,c]] <- d`, when `a` is just an array and `b` and `c` are indices in the array, this opcode will assign to the C array address directly, similar to how `a[[b,c]]` would access the array address directly.

I implemented it based off of `Subassign1_1D`, `Subassign2_1D`, `Extract1_2D`, and `Extract2_2D`.


mhyee:
>  fixed bug when duplicating shared matrix - why are shared matrices duplicated?

R has copy-on-write semantics. In the example below, I create a matrix and pass it to a function `f` that assigns into the matrix. Inside `f`, the matrix is modified, but we don't want to modify the original matrix `m`. One solution is to always copy, but that's expensive if you never modify the matrix. So R copies only if the matrix might be modified. (Yes, there are some cases where R copies when it doesn't actually need to.)

```
> m <- matrix(c(2, 4, 3, 1, 5, 7), nrow=3, ncol=2)
> m
     [,1] [,2]
[1,]    2    1
[2,]    4    5
[3,]    3    7
> f <- function(m) m[[1,2]] <- 42
> f(m)
> m
     [,1] [,2]
[1,]    2    1
[2,]    4    5
[3,]    3    7
```

In the example below, I use the superassign operator, because I want to update the matrix in the parent environment.

```
> g <- function(m) m[[1,2]] <<- 42
> g(m)
> m
     [,1] [,2]
[1,]    2   42
[2,]    4    5
[3,]    3    7
```

For more information, see http://r.cs.purdue.edu/pub/ecoop12.pdf, especially the parts about "copy-on-write" or "copying."

mhyee:
Great, all the tests pass now! I'll review this tomorrow. @JanJecmen could you also take a look (unless you're busy)?

rir/tests/rir_assign.R
```diff
@@ -110,3 +110,77 @@ f13()
...
+#   }
+# 
+#   library(mgcv)
+#   set.seed(1)
+#   n <- 400
+#   dat <- gamSim(1, n=n)
+#   dat$y <- rzip(dat$f / 4 - 1)
+#   b <- gam(y~s(x0)+s(x1)+s(x2)+s(x3), family=ziP(), data=dat)
+# })
+# f19()
```
JanJecmen: Why is `f19` commented out? Is it broken? It looks like something a bit more involved, so either remove it or put it in some `rir_regression` file?

rir/src/compiler/pir/instruction_list.h
```diff
@@ -29,7 +29,9 @@
     V(Extract2_2D)                                                             \
```
JanJecmen, line 29: It's not in this pr, but why are the `Extract*_1D` under binops and not rw instructions?

rir/src/compiler/pir/instruction.h
```diff
@@ -790,6 +790,36 @@ class FLIE(Subassign2_1D, 4, Effect::None, EnvAccess::Leak) {
...
+  public:
+    Subassign1_2D(Value* val, Value* vec, Value* idx1, Value* idx2, Value* env,
+                  unsigned srcIdx)
+        : FixedLenInstructionWithEnvSlot(PirType::val(),
+                                         {{PirType::val(), PirType::val(),
+                                           PirType::val(), PirType::val()}},
+                                         {{val, vec, idx1, idx2}}, env,
+                                         srcIdx) {}
+    Value* rhs() { return arg(0).val(); }
+    Value* lhsValue() { return arg(1).val(); }
```
JanJecmen: I would rather say `rhsValue` and `lhs`, because the left side is the matrix you are subassigning, right?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -525,22 +525,40 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
...
         Value* idx = pop();
         Value* vec = pop();
         Value* val = pop();
         push(insert(new Subassign2_1D(val, vec, idx, env, srcIdx)));
         break;
     }
 
+    case Opcode::subassign1_2_: {
+        Value* idx1 = pop();
+        Value* idx2 = pop();
```
JanJecmen, line 545: Aren't these swapped? And for the other opcode as well
```
Value* idx2 = pop();
Value* idx1 = pop();
```

JanJecmen: Or, seeing travis is green, how come this order is ok? :)

rir/src/interpreter/interp.cpp
```diff
@@ -2503,7 +2467,82 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
...
+
+            SEXP args = CONS_NR(vec, CONS_NR(idx, CONS_NR(val, R_NilValue)));
+            SET_TAG(CDDR(args), symbol::value);
+            PROTECT(args);
+
+            res = nullptr;
+            SEXP call = getSrcForCall(c, pc - 1, ctx);
+            RCNTXT assignContext;
+            Rf_begincontext(&assignContext, CTXT_RETURN, call, *env,
+                            ENCLOS(*env), args, R_SubassignSym);
```
JanJecmen: Why use `R_SubassignSym` and not the one we have (just to be consistent)?

rir/src/interpreter/interp.cpp
```diff
@@ -2596,6 +2633,113 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
...
+                // Fast case only if
+                // 1. index is numerical and scalar
+                // 2. matrix is real and shape of value fits into real
+                //      or matrix is int and shape of value is int
+                //      or matrix is generic
+                // 3. value fits into one cell of the matrix
+                if ((idx1T == INTSXP || idx1T == REALSXP) &&
+                    (XLENGTH(idx1) == 1) && // 1
+                    (idx2T == INTSXP || idx2T == REALSXP) &&
+                    (XLENGTH(idx2) == 1) && // 2
```
JanJecmen: Just being pedantic, but this is still `// 1` ;)
And there is no `// 4`

rir/src/ir/Compiler.cpp
```diff
@@ -376,40 +376,56 @@ bool compileSpecialCall(Context& ctx, SEXP ast, SEXP fun, SEXP args_) {
...
-                        return true;
+                auto idx2 = is2d ? (g.begin() + 3) : idx;
+                if ((fun2 == symbol::DoubleBracket ||
+                     // TODO: Figure out why 2d bracket assign fails sometimes
+                     // (specifically, f19 in tests/pir-assign)
+                     (!is2d && fun2 == symbol::Bracket)) &&
+                    *idx != R_DotsSymbol && *idx != R_MissingArg &&
+                    !idx.hasTag() &&
+                    (!is2d || (*idx2 != R_DotsSymbol && *idx2 != R_MissingArg &&
+                               !idx2.hasTag()))) {
```
JanJecmen: This is very confusing.. but then again, so is the whole compiler

rir/tests/rir_assign.R
Jakobeha: It's commented out because it takes too long on Travis, I don't know how to make it smaller

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
Jakobeha: They should be same order as `extract*_2_` so maybe they are swapped. Travis is probably green because PIR doesn't do anything with `Subassign*_2_` (which is also why `rhs`/`lhsValue` is OK, I'll fix both)

rir/src/interpreter/interp.cpp
Jakobeha: Same as `R_ExtractSym` - maybe I should change that one too?

rir/src/ir/Compiler.cpp
Jakobeha: Checks that the indices aren't `...`, missing, or labeled (e.g. `a[foo = bar,...]` would fail). Doesn't check `idx2` unless it's a 2d subassign, because otherwise `idx2` isn't used

rir/src/simple_instruction_list.h
```diff
@@ -1,6 +1,12 @@
 #ifndef SIMPLE_INSTRUCTION_LIST_H
 #define SIMPLE_INSTRUCTION_LIST_H
 
+// Steps to define a simple instruction:
```
mhyee: Thanks for adding this comment. Could you also explain what a "simple instruction" is? (I suspect that most new instructions we add will not be simple instructions.)

Don't you also need to update something in `rir_2_pir` and `pir_2_rir`?

rir/tests/rir_assign.R
mhyee: Yeah, this was a regression, but it's too slow on Travis _and_ the pre-push hook. I'd probably just delete this test entirely.

How slow are these other tests? Does it make sense to make the matrices smaller?

rir/src/ir/BC_noarg_list.h
```diff
@@ -6,64 +6,66 @@
...
+    V(NESTED, isfun, isfun)                                                    \
+    V(NESTED, invisible, invisible)                                            \
+    V(NESTED, visible, visible)                                                \
+    V(NESTED, extract1_1, extract1_1)                                          \
+    V(NESTED, extract1_2, extract1_2)                                          \
+    V(NESTED, extract2_1, extract2_1)                                          \
+    V(NESTED, extract2_2, extract2_2)                                          \
+    V(NESTED, swap, swap)                                                      \
+    V(NESTED, isobj, isobj)                                                    \
+    V(NESTED, return_, return )
```
mhyee, line 68: Nitpick: added space.

rir/src/compiler/pir/instruction_list.h
mhyee: Blame points to @o- who added `Extract*_*D` in [this commit](https://github.com/reactorlabs/rir/commit/ad3b5d95ca15d5e701af361417b074786c17ee06#diff-5389190c5ecd1008632e75e114d5b977) and then moved `Extract*_2D` to rw instructions in [this commit](https://github.com/reactorlabs/rir/commit/457b583ce566412d6d1c6c143881dd9fc531ed6f#diff-5389190c5ecd1008632e75e114d5b977).

Should we move `Extract*_1D` over as well? Thoughts @o-?

rir/src/ir/Compiler.cpp
mhyee: At this point, I'd be tempted to just have separate cases for 1D and 2D, instead of trying to handle both at the same time. Even if it means duplicating the code a little.

Either way, I would prefer if this condition could be less confusing, but I'm not sure how to achieve that...

rir/src/compiler/translations/pir_2_rir.cpp
```diff
@@ -896,11 +896,6 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 cs << BC::is(is->sexpTag);
                 break;
             }
-            case Tag::Subassign2_1D: {
```
mhyee, line 899: Good catch!

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
mhyee: Yeah, compared to `interp.cpp` it looks like these are swapped.

I don't know if there's an easy way to add some sort of PIR test that doesn't involve an optimization pass.

rir/src/compiler/pir/instruction.h
mhyee: Does this also mean `Subassign*_1D` above should be changed?

rir/tests/rir_regression.R
```diff
@@ -58,3 +58,4 @@ function (x)
 for (i in 1:5000) f(c(1,2,3))
 
 (rir.compile(function(a) {a;a}))(1)
+
```
mhyee: Nitpick: unnecessary added line.

rir/tests/rir_assign.R
Jakobeha: Other tests are fine, the matrices are only around 5x5

rir/src/compiler/pir/instruction_list.h
Jakobeha: I moved both

rir/src/compiler/pir/instruction.h
Jakobeha: I changed both

rir/src/simple_instruction_list.h
Jakobeha: I don't think so, since I added 2 new simple instructions, I just needed to change `interp.cpp`

mhyee: Ah, my mistake. `rir_2_pir` and `pir_2_rir` use the `SIMPLE_INSTRUCTIONS` macro to handle those cases.

Maybe the comment explaining "simple instructions" could say something like "the cases for rir_2_pir and pir_2_rir are trivial and handled by this macro, so you don't need to update anything there."

rir/src/ir/BC_noarg_list.h
Jakobeha: Clang-format wants this, so I kept it

rir/src/compiler/pir/instruction_list.h
o-: ehm, yeah, well I guess they are not RW, but just R, that's maybe why?
Maybe check where the lists are used and just use your best judgement...

rir/src/ir/Compiler.cpp
```diff
@@ -122,13 +122,18 @@ class Context {
     void pushPromiseContext(SEXP ast) { code.push(new PromiseContext(ast, fun, code.empty() ? nullptr : code.top())); }
 
     Code* pop() {
-        auto res = cs().finalize(0);
+        Code* res = cs().finalize(0);
```
mhyee, line 125: Just curious, but why were all the `auto`s changed? (You can leave it the way it is now. I think seeing the actual type does help, unless it's a really long and ugly type.)

rir/src/ir/Compiler.cpp
```diff
@@ -122,13 +122,18 @@ class Context {
...
         delete code.top();
         code.pop();
         return res;
     }
 };
 
+// `true` if an argument isn't missing, labeled, or `...`.
+bool isRegularArg(RListIter& arg) {
```
mhyee: Since this is only used by `compileSpecialCall`, I would probably make it a lambda function, and define it inside `compileSpecialCall`, on line 145.

Something like:
```
auto isRegularArg = [](RListIter& arg) {
 return ...
};
```

mhyee:
Looks like everything's passing. So once you change the function to a lambda, I'll merge it in.

Jakobeha:
Ok, I applied the fix

mhyee:
Thanks! Looking forward to how it affects our benchmarks.

PR #313, opened 2019-01-29
fix travis checks

o-:


o-:
@charig 
if you don't mind I would like to change this back. because otherwise I don't see which test is actually failing

charig (approved):
Don't mind.
Thanks for the fix :).
I'd add `if [[ !"$BENCHS" -eq 1 ]]` the test commands. 
Now they may not be needed but the tests on the bench instance do not add any information at all.

PR #315, opened 2019-02-04
add a flag to measure how much time is spent in compiler

o-:


rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -124,29 +126,79 @@ void Rir2PirCompiler::compileClosure(Closure* closure,
...
+std::chrono::time_point<std::chrono::high_resolution_clock> endTime;
+CompilerPerf PERF;
+
 void Rir2PirCompiler::optimizeModule() {
     logger.flush();
     size_t passnr = 0;
     for (auto& translation : translations) {
         module->eachPirClosure([&](Closure* c) {
             c->eachVersion([&](ClosureVersion* v) {
+                if (MEASURE_COMPILER_PERF)
```
charig: True that we can use it with a raw pir_debug, but I do not see why we would want to profile the logger, so, I would measure the time without including the log times

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -124,29 +126,79 @@ void Rir2PirCompiler::compileClosure(Closure* closure,
...
+                    PERF.addTime(translation->getName(), passDuration.count());
+                }
 
 #ifdef ENABLE_SLOWASSERT
                 assert(Verify::apply(v));
 #endif
             });
         });
     }
+    if (MEASURE_COMPILER_PERF)
```
charig, line 165: Same here, I do not see a reason for profiling the verifier...

charig (approved):
Nice addition. Maybe I would add the CompilerPerf class to another file just to avoid bloating the code of the main behavior orthogonal concerns

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
o-: the verifier was actually major source of time spent at some point...

o-:
> Maybe I would add the CompilerPerf class to another file

ok, did that

PR #316, opened 2019-02-04
Interpreter performance

o-:
some small and stupid things that popped up on callgrind. Also fix gcc 8 warnings. Based on #309, this pr only adds the last commit.

### DIFF
--- a/rir/src/runtime/ArgsLazyData.h
+++ b/rir/src/runtime/ArgsLazyData.h
@@ -26,6 +26,9 @@ typedef std::function<SEXP(void*)> LazyFunction;
 
 struct ArgsLazyData : public RirDataWrapper<ArgsLazyData, LAZY_ARGS_MAGIC> {
     ArgsLazyData() = delete;
+    ArgsLazyData(const ArgsLazyData&) = delete;
+    ArgsLazyData& operator=(const ArgsLazyData&) = delete;
+
     ArgsLazyData(const CallContext* callCtx, Context* cmpCtx)
         : RirDataWrapper(2), callContext(callCtx), compilationContext(cmpCtx){};
 
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
@@ -16,11 +16,8 @@ namespace pir {
 
 // Currently PIR optimized functions cannot handle too many arguments or
 // mis-ordered arguments. The caller needs to take care.
-const Assumptions::Flags Rir2PirCompiler::minimalAssumptions =
-    Assumptions::Flags(
-        {Assumption::CorrectOrderOfArguments, Assumption::NotTooManyArguments});
-const Assumptions Rir2PirCompiler::defaultAssumptions = Assumptions(
-    {Assumption::CorrectOrderOfArguments, Assumption::NotTooManyArguments}, 0);
+constexpr Assumptions::Flags Rir2PirCompiler::minimalAssumptions;
+constexpr Assumptions Rir2PirCompiler::defaultAssumptions;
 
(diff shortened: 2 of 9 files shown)


rir/src/runtime/ArgsLazyData.h
```diff
@@ -26,6 +26,9 @@ typedef std::function<SEXP(void*)> LazyFunction;
 
 struct ArgsLazyData : public RirDataWrapper<ArgsLazyData, LAZY_ARGS_MAGIC> {
     ArgsLazyData() = delete;
+    ArgsLazyData(const ArgsLazyData&) = delete;
+    const ArgsLazyData operator=(const ArgsLazyData&) = delete;
```
mhyee: The build failure is from cppcheck:

```
'ArgsLazyData::operator=' should return 'ArgsLazyData &'.
```

mhyee (commented):
I looked at the last commit and it looks good. Just one cppcheck issue.

One of the builds also timed out, but when I restarted it was fine. (And the restarted build only took 14 min, well below the timeout.)

mhyee:
Also, what sort of stuff did you see with callgrind? Because every time I use it (even with callgrind_annotate and kcachegrind) I find the output overwhelming.

o-:
> Also, what sort of stuff did you see with callgrind?

* There was a lot of time spent in malloc/delete. The main culprit was that we allocated a vector for the synthesized frames on every call, even if we don't need them. Just having an std::vecotor on the stack is apparently fairly slow
* Also a lot of time was spent in CallContext constructor. I realized that this one does `addDynamicAssumptions` for every call, even if we would then later end up in `callBuiltin` or `callLegacy` where we don't even do any dispatching, thus don't need any assumptions. So I moved that out of the constructor and only do it where needed.
* Then I saw that `Assumption::isEager` shows up on the profile. Since I wanted the assumptions class to be a for-free wrapper around a bitset that was bad. So I made everything `RIR_INLINE`.
* But then I looked at the assembly code for isEager and realized that it produces terrible code:
1. static function variables are terrible, they result in a global variable lookup and a branch for every execution. I was not aware and thought that the compiler does some initializer tricks, but apparently every call to the function will trigger a check if it's initialized.
2. I just wanted to give a name to a couple of flags. the distilled snippet is:
```
static Assumption ListOfAssumptionsIMightWantToCheck = {A1, A2, A3};
assumptions.includes(ListOfAssumptionsIMightWantToCheck[theNumber]);
```
and I foolishly assumed that the compiler would be clever enough to basically turn this into:
```
switch(theNumber)
  case A1: return bittest(assumptions, 33)
  case A1: return bittest(assumptions, 43)
  ...
```
But the code was really horrible and looked like the compiler had no idea what the contents of the list of assumptions was. So I turned this into a static class member and made it a constexpr. After that the code looked much better.

Also make sure that your benchmark spends like 100 times more time outside of the compiler (assuming you don't want to profile the compiler), otherwise it's of course very noisy. I'd try to start with a benchmark where perf shows that like 80% is spent in `rirEvalCode`.

Then, basically I am looking for odd things.
* malloc/operator delete, that was a red flag since we should not use the C++ heap much outside of the compiler.
* isEager is a tiny getter, it should definetly not show up on a profile
* A constructor for the stupid data struct CallContext, which is just aggregated list of args we pass around (basically to avoid having to pass 10 arguments to all the different call kinds) should definetly not show up.

On the benchmark I played with (lots of recursive calls) this is straight up 15% of speedup. So I believe there are still low-hanging fruits in the interpreter.

BUT as a word of caution: don't spend too much time here. I think it's a good thing to do from time to time to find the worst stupidities. But the real speedups have to come from either the compiler, or feature completing rir.

rir/src/runtime/ArgsLazyData.h
o-: I can never remember. why doesn't this language have sane defaults :(((

PR #317, opened 2019-02-05
 Identify if only a subset of arguments is eager 

o-:
Try to analyze the order of argument forcing. Then use that
property to pass those arguments in an eager fashion.

### DIFF
--- a/rir/src/compiler/opt/eager_calls.cpp
+++ b/rir/src/compiler/opt/eager_calls.cpp
@@ -28,20 +28,39 @@ void EagerCalls::apply(RirCompiler& cmp, ClosureVersion* closure,
             ClosureVersion* version = call->tryDispatch();
 
             if (!version ||
-                !version->properties.includes(
-                    ClosureVersion::Property::IsEager) ||
                 call->nCallArgs() != cls->nargs())
                 continue;
 
+            bool allEager =
+                version->properties.includes(ClosureVersion::Property::IsEager);
+            if (!allEager && version->properties.argumentForceOrder.empty())
+                continue;
+
+            auto isEager = [&](size_t i) {
+                // TODO: implement the case where we have a non-standard
+                // argument evaluation order
+                if (allEager)
+                    return true;
+                for (auto a : version->properties.argumentForceOrder) {
+                    if (a == i)
+                        return true;
+                    if (a > i)
+                        return false;
+                }
+                return false;
+            };
+
             std::unordered_set<MkArg*> args;
             bool noMissing = true;
+            size_t i = 0;
             call->eachCallArg([&](Value* v) {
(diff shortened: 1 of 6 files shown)


rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -136,15 +133,63 @@ struct ForcedBy {
...
+        return ambiguousForceOrder;
+    }
+
+    bool eagerLikeFunction(ClosureVersion* fun) const {
+        if (!forceArgsBeforeEffect)
+            return false;
+        if (ambiguousForceOrder || argumentForceOrder.size() < fun->nargs())
+            return false;
+        for (size_t i = 0; i < fun->nargs(); ++i)
+            if (argumentForceOrder[i] != i)
```
mhyee, line 182: I'm trying to figure out what is going on here, and it seems like `eagerLikeFunction` requires _all_ arguments to be forced, in left-to-right order?

So if I had a function

```
f <- function(a1, a2, a3, a4) ...
```

the arguments need to be forced in the order `a1`, `a2`, `a3`, and `a4`?

And if they're force in some other order, e.g., `a2`, `a1`, `a3`, `a4`, that's no good?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -136,15 +133,63 @@ struct ForcedBy {
...
+                }
+            }
+        }
+
         return res;
     }
 
+    bool maybeForced(size_t i) const {
+        for (auto f : argumentForceOrder) {
+            if (f == i)
```
mhyee, line 172: This is to check if the given argument id `i` has already been forced?

Might be a bit heavy to bring in `<algorithm>` and `std::find`, but I think that would be slightly more clear. I definitely went "huh?" when I saw `if (f == i)`. Either way, a comment would help.

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -28,20 +28,39 @@ void EagerCalls::apply(RirCompiler& cmp, ClosureVersion* closure,
...
+
+            auto isEager = [&](size_t i) {
+                // TODO: implement the case where we have a non-standard
+                // argument evaluation order
+                if (allEager)
+                    return true;
+                for (auto a : version->properties.argumentForceOrder) {
+                    if (a == i)
+                        return true;
+                    if (a > i)
```
mhyee, line 46: These index comparisons were also throwing me off, until I looked at `force_dominance.cpp`.

rir/src/compiler/opt/force_dominance.cpp
o-: yes

PR #319, opened 2019-02-11
fix bugs from some_eager_args branch

o-:
A couple of issues in the branch prevented this from properly working.
First of all inferAvailableAssumptions was broken and would only change
argument 0. Then eager_call optimization might have cleared already existing
eagerness assumptions.

To simplify the overall approach I removed the eagerArgs and nonObjArgs
flags and now only rely on the eagerArg0 kind of flags, which specify
exactly one argument position. The combination of having eagerArgs,
saying all args are eager, and eagerArg0, saying just arg0 is eager, at
the same time is just too error prone.

### DIFF
--- a/rir/src/runtime/Assumptions.cpp
+++ b/rir/src/runtime/Assumptions.cpp
@@ -16,14 +16,14 @@ std::ostream& operator<<(std::ostream& out, Assumption a) {
     case Assumption::Arg3IsEager_:
         out << "Eager3";
         break;
-    case Assumption::EagerArgs_:
-        out << "Eagers";
+    case Assumption::Arg4IsEager_:
+        out << "Eager4";
+        break;
+    case Assumption::Arg0IsEager_:
+        out << "Eager0";
         break;
     case Assumption::CorrectOrderOfArguments:
         out << "CorrOrd";
-        break;
-    case Assumption::NonObjectArgs_:
-        out << "!Objs";
         break;
     case Assumption::Arg1IsNonObj_:
         out << "!Obj1";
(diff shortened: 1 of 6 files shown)


rir/src/runtime/Assumptions.cpp
```diff
@@ -55,7 +58,7 @@ std::ostream& operator<<(std::ostream& out, const Assumptions& a) {
     return out;
 }
 
-constexpr std::array<Assumption, 3> Assumptions::ObjAssumptions;
-constexpr std::array<Assumption, 3> Assumptions::EagerAssumptions;
+constexpr std::array<Assumption, 5> Assumptions::ObjAssumptions;
+constexpr std::array<Assumption, 4> Assumptions::EagerAssumptions;
```
JanJecmen, line 62: Why do we now have 5 and 4, when before it was 3 and 3?

o-: just wanted to see who comments first on it :)

o-: before, there was one flag for all are eager + specific flags for 1-3 when the fomer was not set (same for obj). but that was just to brittle. so i removed the all are eager flag and added some more specific flags.

o-: the 4/5 thing i really just put there to annoy people who like symmetry, sorry :)

JanJecmen: LOL ok ;-D

charig: :)...But is there any reason why we track 5 arguments for eagerness and only for 4 for "objectness"?

PR #320, opened 2019-02-12
Speculate builtins

o-:
in eagerCall optimization pass also speculate on builtins, that we did not speculate so far.

If we have some builtin call, but we can't statically prove that the call does go to a builtin, then we would like to speculate.
Because speculating on a builtin allows us to eagerly evaluate all arguments.
So far we had to decide on whether to make this speculation in the first pass of the compiler.
This PR presents the first case where we are able to do this speculation late in the pipeline.
To be able to support this case, we need to find out if there is a deoptimization point available.
The deoptimization point can come from anywhere before the speculation, but after the last effect.
It is always better to use an earlier deoptimization point, because that leads to less deoptimization points being used, thus more that can be removed, thus less restrictions for the optimizer.
I added a simple `AvailableCheckpoints` analysis, that answers that question.
The api is `AvailableCheckpoint checkpoint(code); Checkpoint* res = checkpoint.at(instruction); if (res) ...`.
In the same PR is also improve the StaticAnalysis framework a bit, to speed up random access queries, like in this one.
(For the existing optimizations it was enough to be able to quickly iterate over the analysis result at every instruction in sequence.)

### DIFF
--- a/rir/src/compiler/analysis/available_checkpoints.h
+++ b/rir/src/compiler/analysis/available_checkpoints.h
@@ -1,0 +1,41 @@
+#ifndef PIR_AVAILABLE_CHECKPOINTS_H
+#define PIR_AVAILABLE_CHECKPOINTS_H
+
+#include "../util/cfg.h"
+#include "abstract_value.h"
+#include "generic_static_analysis.h"
+
+namespace rir {
+namespace pir {
+
+class AvailableCheckpoints : public StaticAnalysis<AbstractUnique<Checkpoint>> {
+  public:
+    ClosureVersion* code;
+    AvailableCheckpoints(ClosureVersion* cls, LogStream& log)
+        : StaticAnalysis("AvailableCheckpoints", cls, cls, log), code(cls) {}
+
+    AbstractResult apply(AbstractUnique<Checkpoint>& state,
+                         Instruction* i) const override {
+        if (state.get()) {
+            if (i->hasObservableEffect()) {
+                state.clear();
+                return AbstractResult::Updated;
+            }
+        } else {
+            if (auto cp = Checkpoint::Cast(i)) {
+                state.set(cp);
+                return AbstractResult::Updated;
+            }
+        }
+        return AbstractResult::None;
+    };
+
+    Checkpoint* at(Instruction* i) {
+        return StaticAnalysis::at<PositioningStyle::BeforeInstruction>(i).get();
+    }
+};
+
+} // namespace pir
+} // namespace rir
+
+#endif
--- a/rir/src/compiler/analysis/generic_static_analysis.h
+++ b/rir/src/compiler/analysis/generic_static_analysis.h
@@ -66,6 +66,20 @@ class StaticAnalysis {
 
     virtual AbstractResult apply(AbstractState&, Instruction*) const = 0;
 
+    constexpr static size_t MAX_CACHE_SIZE = 128 / sizeof(AbstractState);
+
+    std::unordered_map<Instruction*, AbstractState> cache;
+    std::deque<Instruction*> cacheQueue;
+    void addToCache(Instruction* i, const AbstractState& state) const {
+        if (cacheQueue.size() > MAX_CACHE_SIZE) {
+            auto oldest = cacheQueue.front();
+            const_cast<StaticAnalysis*>(this)->cacheQueue.pop_front();
+            const_cast<StaticAnalysis*>(this)->cache.erase(cache.find(oldest));
+        }
+        const_cast<StaticAnalysis*>(this)->cache.emplace(i, state);
+        const_cast<StaticAnalysis*>(this)->cacheQueue.push_back(i);
+    }
+
   protected:
     AbstractState exitpoint;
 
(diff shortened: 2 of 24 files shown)


rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -325,15 +326,22 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
...
         Assume* assumption = nullptr;
         // Insert a guard if we want to speculate
-        auto ldfun = LdFun::Cast(callee);
         if (monomorphicBuiltin || monomorphicClosure) {
             Value* expected = insert(new LdConst(monomorphic));
             Value* given = callee;
             // This change here potentially allows the delay_instr pass
             // to move the ldfun into the deopt branch
-            if (ldfun)
+            if (ldfun && ldfun->varName != symbol::c)
```
charig, line 354: What is the issue `symbol::c`?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -1018,6 +1034,10 @@ Value* Rir2Pir::tryTranslate(rir::Code* srcCode, Builder& insert) const {
                 assert(false);
                 return nullptr;
             }
+
+            if (!inPromise() && !insert.getCurrentBB()->isEmpty() &&
+                insert.getCurrentBB()->last()->hasEffect())
```
charig, line 1049: So we add a checkpoint after each effectful instruction now? What happen with calls? There is already a framestate added? We resort to pir later removing the redundancy?

rir/src/compiler/pir_tests.cpp
```diff
@@ -215,6 +215,8 @@ bool canRemoveEnvironmentIfTypeFeedback(const std::string& input) {
     bool t = verify(&m);
     m.eachPirClosureVersion(
         [&t](pir::ClosureVersion* f) { t = t && envOfAddElided(f); });
+    if (!t)
```
charig, line 218: debugging stuff?

rir/src/compiler/pir/instruction.h
```diff
@@ -279,6 +280,9 @@ class InstructionImplementation : public Instruction {
     static constexpr bool mayLeakEnv_ = ENV >= EnvAccess::Leak;
 
     bool hasEffect() const override { return EFFECT > Effect::None; }
+    bool hasObservableEffect() const override {
+        return EFFECT > Effect::Order && hasEffect();
```
charig, line 284: Why hasEffect()?

rir/src/compiler/analysis/available_checkpoints.h
```diff
@@ -0,0 +1,41 @@
+#ifndef PIR_AVAILABLE_CHECKPOINTS_H
```
charig, line 1: It may be a little overkilling an analysis that just looks for the last cp.

rir/src/compiler/analysis/generic_static_analysis.h
```diff
@@ -158,30 +172,55 @@ class StaticAnalysis {
 
     template <PositioningStyle POS>
     AbstractState at(Instruction* i) const {
+        if (!done)
```
charig, line 175: Where are we calling at before the analysis has finished?

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -14,91 +17,213 @@ namespace rir {
...
+        // skip ldfun
+        ip++;
+
+        auto expected = new LdConst(builtin);
+        ip = bb->insert(ip, expected);
+        ip++;
+        Instruction* given = ldfun;
+
+        // Function c overwritten by non-function c happens too often. The more
+        // conservative ldvar is not a good fit here.
```
charig: The comment?

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -14,91 +17,213 @@ namespace rir {
...
+                            // forcing arguments.
+                            if (checkpoint.at(ldfun)) {
+                                replaced.emplace(ldfun, ldfun->hint);
+                                replaceCallWithCallBuiltin(bb, ip, call,
+                                                           replaced.at(ldfun));
+                            }
+                        }
+                    } else {
+                        // Only speculate if we don't have a static guess
+                        if (!ldfun->guessedBinding()) {
```
charig, line 104: Maybe we can add profiling also for standard calls?

charig (approved):

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: > So we add a checkpoint after each effectful instruction now?

yes. It is hard to predict where checkpoints are needed. For example a case where this pr really helps, is when we have a builtin in a promise. Now often it is possible that after inlining the promise, we can use the checkpoint from the outer function to speculate on the builtin.

> What happen with calls? There is already a framestate added?

In theory we could use this framestate. Though it is the framestate *during* the call. E.e. it does not capture the result of the call. Reusing it would be a hack. We would have exactly the problem we talked about, where the framestate is inserted before the call, is referenced by the call, but logically after the call. Very confusing. Plus, to use the framestate for inlining we would need to somehow remove the call result after the fact.

> We resort to pir later removing the redundancy?

Yes. Unfortunately I don't see a way. Often many of them can be removed after some optimizations. But it's hard to predict which.

rir/src/compiler/pir/instruction.h
o-: some subclasses override `hasEffect`. Since there is an ordering in the effects, if an instruction has no effect at all, it also doesn't have an observable effect....

rir/src/compiler/analysis/available_checkpoints.h
o-: why? Do you have a suggestion how else to do it? I mean, even if you attach cp to instructions, we would still need to recompute using such an analyis. To statically know which is the last cp since the last effect is inherently a dataflow problem...

rir/src/compiler/analysis/generic_static_analysis.h
o-: in this PR the last checkpoint analysis is used without calling it first. This computes it lazily, so when we do not want to speculate, the analysis is not computed.

rir/src/compiler/pir_tests.cpp
o-: but useful, if the test fails, you also see the code of the the failing test...

rir/src/compiler/opt/eager_calls.cpp
o-: sorry, I am not sure if I understand this comment. What do you mean by standard call. We should have profiling for all calls. But I guess sometimes we don't get feedback. Not 100% sure why.

rir/src/compiler/analysis/available_checkpoints.h
charig: I was just refering to use a "fixpoint" static analysis. A visitor should be enough, shouln't it?

o-: no, it's a fixpoint...

think about:
```
  cp1 = checkpoint
loop:
  is cp1 valid here?
  effect
  goto loop
```

rir/src/compiler/opt/eager_calls.cpp
charig: Oh yes, but then this seems to open the possibility to speculate even if we already observed during the warmup multiple targets.  Oth, I do not fully get the `Rf_findVar` here in a pir pass.

o-: > Oh yes, but then this seems to open the possibility to speculate even if we already observed during the warmup multiple targets

yes, actually your comment made me realize that and I fixed this by adding a dummy value in this case.

> Oth, I do not fully get the Rf_findVar here in a pir pass.

well, if we already statically see that the function is overwritten, then we don't need to speculate, since we know it will fail...

PR #321, opened 2019-02-12
[wip] Typed stack

Jakobeha:
Makes the RIR interpreter use R's typed stack and store simple scalars unboxed when possible. Currently, uses unboxed scalars in arithmetic, extract/assign, and sequence operations. Still boxes when loading/storing variables, or calling functions - perhaps we can solve this later.


o-:
A general comment to your approach:

If you convert a typed stack cell to a sexp, you should probably do it inline. Because otherwise there is a danger, that you create copies.

for example if you convert the same number to a sexp twice, and one of them gets changed, then this will lead to a wrong result, since the copies are now out of sync.
so, if you need to convert to a sexp, then you should do it inline (ie. set the sexp to the same stack location and just set the tag to 0).

Also, the interpreter sometimes keeps things on the stack on purpose. Because as long as they are on the stack, they are visible to the GC. For example we do stuff like:

```
SEXP a = stack_at(1)
SEXP a = stack_at(0)
SEXP res = add(a,b)      // <-- here a and b are protected, since they are still on the stack
stack_pop(2)
```

that's another reason why I would convert inline. For example, you could have an api like:

```
static SEXP stack_at_as_sexp(...) {
  if (cell.tag == INT) {
    cell.u.sexp = Rf_AllocVector(cell.u.int);
    cell.tag = SEXP
  }
  ...
  return cell.u.sexp
}
```

o-:
And another thing: if you happen to somehow get rid of the underscore names and use camelCase instead, that would not hurt. I know the existing code uses underscore, but maybe we can get rid of at least some of them.

o-:
@Jakobeha  currently this massively slows down execution. haven't looked into why.

PR #322, opened 2019-02-13
protect constants in pir

o-:
Constants were not protected until they are lowered to rir. This
commit ensures that they already end up in the constants pool, when
the pir instruction gets created.

### DIFF
--- a/rir/src/compiler/pir/instruction.cpp
+++ b/rir/src/compiler/pir/instruction.cpp
@@ -223,11 +223,18 @@ bool Instruction::envOnlyForObj() {
     return false;
 }
 
+LdConst::LdConst(SEXP c, PirType t)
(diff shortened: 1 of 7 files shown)


o-:
bugfix, tbr

rir/src/compiler/pir/instruction.cpp
```diff
@@ -223,11 +223,18 @@ bool Instruction::envOnlyForObj() {
...
+    : FixedLenInstruction(PirType(c)), idx(Pool::insert(c)) {}
+
+SEXP LdConst::c() const { return Pool::get(idx); }
+
 void LdConst::printArgs(std::ostream& out, bool tty) const {
     std::string val;
     {
         CaptureOut rec;
-        Rf_PrintValue(c);
+        Rf_PrintValue(Pool::get(idx));
```
charig, line 237: c()

PR #323, opened 2019-02-13
Avoid env reload

o-:


o-:
@JanJecmen this PR is only for the last commit

rir/src/compiler/analysis/last_env.h
```diff
@@ -0,0 +1,33 @@
...
+namespace pir {
+
+class LastEnv : public StaticAnalysis<AbstractUnique<Value>> {
+  public:
+    LastEnv(ClosureVersion* cls, Code* code, LogStream& log)
+        : StaticAnalysis("Last Env", cls, code, log) {}
+
+    AbstractResult apply(AbstractUnique<Value>& state,
+                         Instruction* i) const override {
+        if (i->hasEnv() && !MkEnv::Cast(i) && i->env() != state.get()) {
```
JanJecmen, line 17: I am still confused by `MkEnv` vs. `LdFunctionEnv`... This looks like exactly the same condition as the p2r compiler `explicitEnvValue`. Why is one explicit and the other not?
Or, shouldn't this test also check the other one?

JanJecmen (approved):
lgtm!

rir/src/compiler/analysis/last_env.h
o-: This is about the arguments. For almost all instruction an environment argument means, that the env should be set using `set_env_`. Except `mk_env_`, this one expects the parent env as a normal argument on the stack and not as the value of `*env`.

PR #325, opened 2019-02-13
Visibility analysis

o-:


o-:
@JanJecmen again, only last commit. I'll give you this one too, since it is so similar to the redundant env elision

rir/src/compiler/analysis/visibility.h
```diff
@@ -0,0 +1,79 @@
...
+
+class CurrentVisibility {
+  public:
+    enum State { Visible, Invisible, Unknown };
+
+    State state = State::Unknown;
+
+    CurrentVisibility() {}
+    CurrentVisibility(const CurrentVisibility&) = default;
+    CurrentVisibility& operator=(const CurrentVisibility&) = default;
```
JanJecmen: I think you don't need to type any of these explicitly

rir/src/compiler/analysis/visibility.h
```diff
@@ -0,0 +1,79 @@
...
+#include "generic_static_analysis.h"
+
+namespace rir {
+namespace pir {
+
+class CurrentVisibility {
+  public:
+    enum State { Visible, Invisible, Unknown };
+
+    State state = State::Unknown;
```
JanJecmen: Just `Unknown` to be consistent?

JanJecmen (approved):
:+1:

PR #326, opened 2019-02-14
Improve debugging

o-:
* Add a flag to see phases instead of every pass.
* Make RIR_DEBUG_PASS_FILTER a regex
* add a RIR_DEBUG_FUNCTION_FILTER
* improve the debugging documentation

### DIFF
--- a/rir/src/compiler/opt/pass_definitions.h
+++ b/rir/src/compiler/opt/pass_definitions.h
@@ -113,6 +113,14 @@ class PASS(OptimizeAssumptions);
 
 class PASS(EagerCalls);
 
+class PhaseMarker : public PirTranslator {
+  public:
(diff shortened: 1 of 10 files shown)


rir/src/compiler/opt/pass_definitions.h
```diff
@@ -113,6 +113,14 @@ class PASS(OptimizeAssumptions);
 
 class PASS(EagerCalls);
 
+class PhaseMarker : public PirTranslator {
+  public:
+    PhaseMarker(const std::string& name) : PirTranslator(name) {}
```
mhyee: `explicit`

mhyee (approved):

charig:
Thanks for the iteration over the documentation!

PR #327, opened 2019-02-14
we can't use exceptions due to longjmp, might as well disable them

o-:


mhyee:
What version of gcc do you have? On the server, I'm using 7.3.0, and after I remove the "ignore strict overflow" pragma, I don't see any errors.

With gcc 8.2.0, I get a different warning:

```
rir/src/ir/BC_inc.h: In static member function ‘static rir::BC::ImmediateArguments rir::BC::decodeImmediateArguments(rir::Opcode, rir::Opcode*)’:
rir/src/ir/BC_inc.h:616:47: warning: ‘void* memcpy(void*, const void*, size_t)’ copying an object of non-trivial type ‘struct rir::BC::StaticCallFixedArgs’ from an array of ‘enum class rir::Opcode’ [-Wclass-memaccess]
                    sizeof(StaticCallFixedArgs));
```

o-:
@mhyee also 7.3.0, but only in release build. and yes I have also seen the 8.2.0 warning, which is legit, but we are taking precautions (so we'll have to silence that one too).

mhyee:
Ah, I thought running `cmake -GNinja . ` would set release mode by default, but I had to explicitly write `-DCMAKE_BUILD_TYPE=release`.

Once you silence the mem access warning, I think this PR is good to merge.

mhyee (approved):

mhyee:
For what it's worth, I don't see either warning with clang 7.0.1.

mhyee:
Actually, I just realized that `.ycm_extra_conf.py` is missing `-fno-rtti -fno-exceptions`.

o-:
> Actually, I just realized that .ycm_extra_conf.py is missing -fno-rtti -fno-exceptions.

good catch!

o-:
> Once you silence the mem access warning, I think this PR is good to merge.

this warning is actually not introduced by this commit, but already older. And now I cannot reproduce it anymore on 8.2.1 so maybe it was fixed on gcc's end.

UPDATE: ok, now i was able to repro and silenced it.

o-:
I hate those stupid warning flags across compilers. Travis says:

> /home/travis/build/reactorlabs/rir/rir/src/ir/BC_inc.h:582:32: error: unknown option after ‘#pragma GCC diagnostic’ kind [-Werror=pragmas]
> #pragma GCC diagnostic ignored "-Wclass-memaccess"
>                                ^~~~~~~~~~~~~~~~~~~

I am just gonna revert this part and stick to the version that works on travis and apparently on most of our machines.

mhyee:
Sorry, but I lost track: what's the current status here? Will you be opening (or re-opening) a PR for (some) of these changes, or is this dead?

mhyee:
Oh, I see it already got merged in 811a1f8f6cc98ac9c6783a6160057405687d8f7d.

PR #328, opened 2019-02-14
Speculatively avoid creating environments thar are there only for force instructions

charig:
Instead, we create a stub, that, ideally, should we much faster to create because it is not a SEXP.
In case the force really accesses its parent environment reflectively, we must create the environment on the fly and deoptimize.

Although the environment is actually stored as the `sysparent` field of `RCNTXT` when forcing promises, it also becomes the `cloenv` of the current `RCNTXT`. Thus we must take care of both fields during a reflective access and also for the garbage collection.


rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -14,61 +14,97 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
...
+
+                    if (!environment->stub &&
+                        environment->usesAreOnly(function->entry, forces)) {
+                        environment = MkEnv::Cast(force->env());
+
+                        if (!stubFor.count(environment)) {
+                            auto stub = (MkEnv*)environment->clone();
+                            stub->stub = true;
+                            stubFor.emplace(environment, stub);
+                            environment->bb()->insert(
```
o-: if `environment()->bb() == bb` then this will potentially trash the iterator `it` (because you are inserting into the container you are iterating over).

rir/src/compiler/opt/delay_env.cpp
```diff
@@ -20,7 +20,7 @@ void DelayEnv::apply(RirCompiler&, ClosureVersion* function, LogStream&) const {
             while (it != bb->begin()) {
                 it--;
                 auto t = MkEnv::Cast(*it);
-                if (t && done.find(t) == done.end()) {
+                if (t && !t->stub && done.find(t) == done.end()) {
```
o-: why not delay the stub?

rir/src/compiler/pir/instruction.h
```diff
@@ -895,10 +896,21 @@ class FLI(Is, 1, Effect::None, EnvAccess::None) {
     void printArgs(std::ostream& out, bool tty) const override;
 };
 
-class FLI(IsObject, 1, Effect::None, EnvAccess::None) {
+class FLI(TypeTest, 1, Effect::None, EnvAccess::None) {
```
o-: I don't particularly like that you combined those two instructions.

It feels like those two instructions are somewhat different. `isObject` is testing the R type. `isStub` is testing the internal data representation of a value, that has always type environment.

So in the case of testing for the stub, we are not actually doing a type test. This can also be seen by the fact that  environment and environment stub have the same type in pir. Otoh if the static pir type says `!maybeObject` then we can constantfold isObject.

rir/src/interpreter/interp.cpp
```diff
@@ -230,6 +230,36 @@ RIR_INLINE void __listAppend(SEXP* front, SEXP* last, SEXP value, SEXP name) {
...
+        arglist = CONS_NR(val, arglist);
+        SET_TAG(arglist, name);
+        SET_MISSING(arglist, val == R_MissingArg ? 2 : 0);
+    }
+    SEXP environment = Rf_NewEnvironment(R_NilValue, arglist, parent);
+
+    // TODO: Find a proper way of going through the function stack and
+    // replacing the stub with the new environment
+    Locals local(localsBase, 5);
+    for (auto i = 0; i < 5; i++) {
```
o-: why 5? I think you have to go through the whole frame, ie. from current stack top until localsBase. You could have arbitrary many values on the stack.

rir/src/interpreter/interp.cpp
```diff
@@ -1307,6 +1344,25 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
...
+            size_t n = readImmediate();
+            advanceImmediate();
+            // Do we need to preserve parent and the arg vals?
+            SEXP parent = ostack_pop(ctx);
+            assert(TYPEOF(parent) == ENVSXP &&
+                   "Non-environment used as environment parent.");
+            advanceImmediateN(n);
+            auto args = new std::vector<SEXP>(n);
+            for (size_t i = 0; i < n; ++i) 
+                args->push_back(ostack_pop(ctx));
```
o-, line 1328: I am not convinced that this is sound. In most cases it will be, because we keep a copy of function arguments when we do a call. But for inlined `mk_envs_` I believe that you will hide the arguments from the GC and this will result in the gc collecting them.

rir/src/interpreter/interp.cpp
```diff
@@ -1326,7 +1382,8 @@ SEXP evalRirCode(Code* c, Context* ctx, SEXP* env, const CallContext* callCtxt,
             // environment
             memset(&bindingCache, 0, sizeof(bindingCache));
             SEXP e = ostack_pop(ctx);
-            assert(TYPEOF(e) == ENVSXP && "Expected an environment on TOS.");
+            assert((TYPEOF(e) == ENVSXP || LazyEnvironment::check(e)) &&
+                   "Expected an environment on TOS.");
```
o-, line 1357: materialize if lazy env?

o- (commented):
Love it :)))

some comments:

* Could we shrink the changes to GNU-R, if we would have an api that works for both the environment and the argslist? Somthing like `isLazyWrapper` and `materialize` should work for both. And then we can disambiguate on the rir side. If we start adding more lazy datastructures, we probably don't want to export a function to GNU R for each of them.
* can you constantfold the `isStub` test? For example if the stub is not used in any force instruction (due to optimizations) you can replace isStub with false in the constanfold pass. (only if you care, can also be added later...)
* I really think that the env stub needs to be traceable by the GC. see inline comment

o-:
modulo the comments, it lgtm

rir/src/interpreter/interp.cpp
charig: No, that would materialize lot of environments that are actually not used.

rir/src/interpreter/interp.cpp
charig: No, It is not. We must find a way to trace those objects. My point is that in our current code this problem does not raise. So maybe we can think on a good design and implement it later. Maybe we should change the `RirWrappers` to `externalSPX`. The problem is twofold. On one hand we do not want to introduce an overhead to the gc. On the other we do not want to lose to much time with the wrapper creation, otherwise the optimization starts to lose sense.

rir/src/compiler/pir/instruction.h
charig: You have a point for sure that I did not think about. Oth, they are very similar instructions at the pir level. I do not see a strong conceptual difference at the pir level and adding instructions for every type of test that we eventually add I think may be nonsense. Pir does not matter what kind of test. Just runs a "predicate" (`isTest`) on a value. Maybe I could change the name from `TypeTest` to `IsTest`.  For the very few cases we need to distinguish (like the constant folding you mention) there is the testFor enum. Then, when translating back to pir I am honoring the difference. 

Summarizing, I admit you have a point. Still, rolling back may imply some time and I am certainly not sure there is a problem here. Neither that this would not be useful in the short term. That said, honestly, if you are strongly against I would gladly rollback.

rir/src/interpreter/RirDataWrapper.h
```diff
@@ -13,32 +13,40 @@
...
         return b;
     }
 
   protected:
-    explicit RirDataWrapper(uint32_t count) : info{MAGIC} {
-        uint8_t* start = (uint8_t*)this + sizeof(uint32_t);
+    explicit RirDataWrapper(uint32_t count, uint32_t gc)
+        : info{RIR_DATA_WRAPPER_MAGIC, MAGIC, gc} {
+        uint8_t* start = (uint8_t*)this + sizeof(data_header);
         memset(start, 0, count * sizeof(void*));
```
o-: I know not technically part of this PR. Of course `sizeof(void*)` is the same as `sizeof(SEXP)` on reallife architectures, but cppcheck complains here. Anyway, I would not put the memset here. Seems to me that this is subclass responsability

rir/src/interpreter/RirDataWrapper.h
```diff
@@ -13,32 +13,40 @@
...
+
+    // magic number to differentiate different RIR data wrappers
+    const uint32_t wrapper;
+
+    // #SEXPs stored in an vector* at the begging of the struct to be preserved
+    const uint32_t gc;
 };
 
 template <typename BASE, uint32_t MAGIC>
 struct RirDataWrapper {
```
o-, line 37: Resposibility of this class is poorly defined. For example the GC support is not implemented here, but in the subclass, so why even bother with storing the number of gc'd objects here?. why two different magic numbers, one seems enough.

rir/src/interpreter/RirDataWrapper.h
```diff
@@ -13,32 +13,40 @@
...
     data_header info;
 
-    static BASE* check(SEXP s) {
+    static BASE* check(void* s) {
         BASE* b = (BASE*)s;
-        return b->info.magic == MAGIC ? b : nullptr;
+        return b->info.wrapper == MAGIC ? b : nullptr;
     }
 
     static BASE* unpack(void* s) {
```
o-: this is not technically unpacking anything. it is a cast.

rir/src/interpreter/RirDataWrapper.h
o-: Imho this class has various problematic unsafe things it does (e.g. casting `void*` to `BASE*` without any restrictions on `BASE`). And it is unclear what it is doing. It should be shrunken down to just the safe casting bits.
For example this would  should be an ok version:
```
template<typename BASE, uint32_t MAGIC>
struct TaggedStruct {
    const uint32_t matic = MAGIC;
    static BASE* Cast(SEXP ptr) {
        static_assert(sizeof(*ptr) >= sizeof(TaggedStruct));
        TaggedStruct* tagged = reinterpret_cast<TaggedStruct*>(ptr);
        if (tagged->magic == MAGIC)
            return reinterpret_cast<BASE*>(tagged);
        return nullptr;
    }
}
```

rir/src/compiler/pir/instruction.h
```diff
@@ -904,10 +906,21 @@ class FLI(Is, 1, Effect::None, EnvAccess::None) {
...
   public:
-    explicit IsObject(Value* v)
+    enum type { Object, EnvironmentStub };
+    type testFor;
+    explicit TypeTest(Value* v)
         : FixedLenInstruction(NativeType::test, {{PirType::val()}}, {{v}}) {}
+    TypeTest* object() {
```
o-: please add testFor to the constructor instead of `object()` and `environmentStub()`. This way of creating objects seems brittle. what if you call both of them, none?

(I know I write similar api's, but only  immutable data style. e.g. in pirType, I always make a copy and return the modified copy)

rir/src/compiler/transform/bb.cpp
```diff
@@ -167,5 +167,16 @@ BB* BBTransform::lowerExpect(Code* code, BB* src, BB::Instrs::iterator position,
     return split;
 }
 
+void BBTransform::insertAssume(BB* src, Value* condition, Value* cp,
```
o-: maybe `assume` and `assumeNot` instead of the boolean?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -203,6 +203,9 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
     case Opcode::ldvar_:
         v = insert(new LdVar(bc.immediateConst(), env));
         push(insert(new Force(v, env)));
+        if (!inPromise()) {
+            addCheckpoint(srcCode, nextPos, stack, insert);
```
o-: should be covered already by the general checkpoint insertion after the bc compilation

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -90,7 +90,23 @@ class TheCleanup {
...
+                                                        Tag::TypeTest};
+                    if (env->stub && env->usesAreOnly(function->entry, tags)) {
+                        env->replaceUsesWith(Env::elided());
+                        removed = true;
+                        next = bb->remove(ip);
+                    }
+                } else if (auto test = TypeTest::Cast(i)) {
+                    if (test->testFor == TypeTest::EnvironmentStub &&
+                        test->arg(0).val() == Env::elided()) {
+                        i->replaceUsesWith((Value*)True::instance());
```
o-: no `(Value*)`

rir/src/compiler/opt/elide_env.cpp
```diff
@@ -69,6 +69,11 @@ void ElideEnv::apply(RirCompiler&, ClosureVersion* function, LogStream&) const {
                     if (!force->input()->type.maybeLazy())
                         force->elideEnv();
                 }
+            } else {
+                if (auto test = TypeTest::Cast(i)) {
+                    if (test->testFor == TypeTest::EnvironmentStub)
+                        envNeeded.insert((Value*)test->arg(0).val());
```
o-: remove `(Value*)`

rir/src/compiler/pir/instruction.h
o-: One thing I noticed reviewing this PR is that you have a lot of `isTest->arg<0>().val()` and then even casting it, whereas it would be a lot better to have `isTest->env()`. you decide

(that was an anwer to shoudl IsEnvStub be it's own instruction. not sure why gh renders it here)

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -131,11 +131,21 @@ void Constantfold::apply(RirCompiler& cmp, ClosureVersion* function,
...
-                if (!isObj->arg<0>().val()->type.maybeObj()) {
+            if (auto isTest = TypeTest::Cast(i)) {
+                if (isTest->testFor == TypeTest::Object &&
+                    !isTest->arg<0>().val()->type.maybeObj()) {
                     i->replaceUsesWith(False::instance());
                     next = bb->remove(ip);
                 }
+
+                if (isTest->testFor == TypeTest::EnvironmentStub) {
+                    auto environment = MkEnv::Cast(isTest->arg<0>().val());
```
o-, line 142: I don't see any invariant in the isTest instruction, that guarantees that this argument is a `MkEnv`.

Either add invariants to the stubtest, or test here, ie. `if (auto environment = MkEnv::Cast(...)) {`

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -16,42 +16,96 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
...
+                auto nextBB = bb->trueBranch();
+                while (nextBB->isEmpty())
+                    nextBB = nextBB->trueBranch();
+                return nextInstr(*(nextBB->begin()), nextBB, false);
+            }
+            return i;
+        };
+
     Visitor::run(function->entry, [&](BB* bb) {
+        std::unordered_map<Value*, Value*> stubFor;
```
o-: dead

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -16,42 +16,96 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
 
     AvailableCheckpoints checkpoint(function, log);
 
+    auto nonObjectArgs = [&](Instruction* i) {
+        auto answer = true;
+        i->eachArg([&](Value* arg) {
+            if (arg->type.maybeObj() && arg->typeFeedback.maybeObj())
+                answer = true;
```
o-: `= false`

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -16,42 +16,96 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
...
+                i = *ip;
+            }
+
+            if (i->branches()) {
+                auto cp = Checkpoint::Cast(i);
+                assert(cp);
+                auto nextBB = bb->trueBranch();
+                while (nextBB->isEmpty())
+                    nextBB = nextBB->trueBranch();
+                return nextInstr(*(nextBB->begin()), nextBB, false);
```
o-: how can you know that the bb is not empty?

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -16,42 +16,96 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
...
+    auto nonObjectArgs = [&](Instruction* i) {
+        auto answer = true;
+        i->eachArg([&](Value* arg) {
+            if (arg->type.maybeObj() && arg->typeFeedback.maybeObj())
+                answer = true;
+        });
+        return answer;
+    };
+
+    std::function<Instruction*(Instruction*, BB*, bool)> nextInstr =
```
o-: this seems to be a lot more than nextInstruction. I am not sure I understand

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -16,42 +16,96 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
...
+                    });
+                    next = ip + 1;
+                    i->type.setNotObject();
+                }
+
+                // Speculatively elide envs on forces that only require them in
+                // case they access promises reflectively
+                if (auto force = Force::Cast(i)) {
+                    auto nextInstruction = nextInstr(force, bb, true);
+                    if (checkpoint.at(nextInstruction)) {
```
o-: ad-hoc searching for deopt point, that seems to ignore many edge-cases. For example, how about:
```
  force
  branch a b
a:
  checkpoint
  ...
b:
  return
```

this would just put the assume in the a branch and ignore the b branch, not?

o-: Another example:
```
    force
    effect
    checkpoint
```
I think it would skip effect, not?

o-: I see the problem here. `AvailableCheckpoints` was written for eager checks, but has no `nextCheckpoint` api, required for lazy checks. This needs some thinking. Do we need a to use a backwards analysis to get the nextCheckpoint?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -187,6 +187,32 @@ void Instruction::replaceUsesAndSwapWith(
     bb()->replace(it, replace);
 }
 
+bool Instruction::usesAreOnly(BB* target, std::unordered_set<Tag> tags) {
```
o-, line 190: nice

rir/src/compiler/pir/instruction.cpp
```diff
@@ -187,6 +187,32 @@ void Instruction::replaceUsesAndSwapWith(
...
+    bool answer = true;
+    Visitor::run(target, [&](Instruction* i) {
+        i->eachArg([&](InstrArg& arg) {
+            if (arg.val() == this && tags.find(i->tag) == tags.end()) {
```
o-: `!tags.count(i->tag)`

rir/src/interpreter/interp.cpp
```diff
@@ -2816,6 +2900,9 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP* env,
             c->registerInvocation();
             assert(c->code() <= pc && pc < c->endCode());
             SEXP e = ostack_pop(ctx);
+            if (LazyEnvironment::check(e)) {
+                e = (LazyEnvironment::unpack(e))->create();
+            }
```
o-, line 2900: why don't we need to update the stack here?

rir/src/interpreter/interp.cpp
```diff
@@ -3057,6 +3144,9 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP* env,
             synthesizeFrames->pop_front();
             SEXP res = ostack_pop(ctx);
             SEXP e = ostack_pop(ctx);
+            if (LazyEnvironment::check(e)) {
+                e = (LazyEnvironment::unpack(e))->create();
+            }
```
o-, line 3144: why don't we need to update the stack here?

o- (commented):
nice addition thanks.

Some additional comments for the gnur changes:

https://github.com/reactorlabs/gnur/compare/768cadbdea70663a0f452b65fe0ded5bcfa6079d...78441f80217cb7010b0966d39e970bef505953ec#diff-d564936611d9b4c8e598252d86c13559R574
Ehm, that is brittle and will break for sure. Is it really needed, did you check that `externalKeepalive() == 0` is too slow as a check?

rir/src/interpreter/RirDataWrapper.h
charig: The memset is not needed at all. I inspired in `RirDataObject` for creating the wrapper, and they have the memset at init time. I thought it was there just for "safety" reasons and that is why I put it also here. And that is also the reason why it is not a subclass responsibility.

o-: it's not the same situation.

`RirDataObject` actually does manage the GC area, as in it owns the memory and it manages access through getters/setters. So it is superclass responsability to initialize.

`RirDataWrapper` does not own the memory. It initializes memory owned by the subclass, which might even be undefined.

rir/src/interpreter/RirDataWrapper.h
charig: It is a kind of cryptic struct strongly connected on how gnu-r uses it. So maybe some comments are worth. But it is not true that the gc is a subclass responsibility. Every `RirDataWrapper` needs to store how many objects that need to be traced by the gc it stores. Then each subclass only implements a function `gcData` that return a pointer to an array of the corresponding size so that gnu-r can walk through all the data. 

Then, two magic numbers is just a matter of is to simplify gnu-r checks. So, gnu-r treats all wrappers that same way, they only need to know the gc objects to trace. I do not want to have tests full of or for all the eventual wrappers in the gnu-r gc. Then, at materialization time we need to disambiguate into different structures. And they are in a inner struct and not as just struct const fields because again, I inspired in `RirRuntimeObjects`

rir/src/compiler/transform/bb.cpp
charig: you mean two different functions?

rir/src/compiler/opt/elide_env_spec.cpp
charig: Because the last two lines loop when the nextBB is empty?

rir/src/compiler/opt/elide_env_spec.cpp
charig: Exactly, this was the first attempt to loop for a forward checkpoint when we have not had the analysis. After the analysis the problem remained there. 

For the examples you wrote, I do not know, my point is that the invariant is that after a force, for which we could speculate, we must have a checkpoint without having effects in between. Then, moving the checkpoint to the branch may happen and that breaks this code true. But why would we do that? 

Summary, yes, a backward analysis would be the proper solution and would simplify the code of this motion by removing the nextInstr function. But I am not sure how to do that soundly in case we can move checkpoints to branches.

rir/src/interpreter/interp.cpp
charig: We need and we are doing it.

rir/src/interpreter/interp.cpp
charig: We need and we are doing it.

rir/src/compiler/opt/elide_env_spec.cpp
o-: > But why would we do that?

The pir passes have to be independently correct. relying on the fact that other passes (currently) don't generate a certain kind of graph is not what we want. if we guarantee a certain invariant on the code, it has to be checked in the verifier!

how to do it soundly: exactly like the availableassumptions, but with a backwards analysis (you can plug and play forward/backward analysis into a static analysis implementation)

PR #329, opened 2019-02-15
fastrack common builtin calls from pir optimized code

o-:
calling from pir optimized code into builtins is rather slow. because pir collects args on the stack, but the builtins take an arguments list.

In this commit I just picked a couple of low-hanging fruits. I looked at the builtin implementations in gnur (ie. check names.c, then go to the relevant `do_...` implementation of the builtin). For a couple of builtins that occur often in our benchmark suite I implemented the subset of types and arguments, that can be easily implemented in a couple of lines..

This commit also features a DEBUG_SLOWCASES define that can be enabled. This will print a statistics of the most used slow builtin calls when the process terminates and can help to identify the next builtin to be implemented.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -517,7 +517,7 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
     }
 
     switch (call.callee->u.primsxp.offset) {
-    case 90: { // c
+    case 90: { // "c"
         if (nargs == 0)
             return R_NilValue;
 
@@ -548,11 +548,13 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
             auto len = XLENGTH(args[i]);
             for (long j = 0; j < len; ++j) {
                 assert(pos < total);
+                // We handle LGL and INT in the same case here. That is fine,
+                // because they are essentially the same type.
+                SLOWASSERT(NA_INTEGER == NA_LOGICAL);
                 if (type == REALSXP) {
                     if (TYPEOF(args[i]) == REALSXP) {
                         REAL(res)[pos++] = REAL(args[i])[j];
                     } else {
-                        SLOWASSERT(NA_INTEGER == NA_LOGICAL);
                         if (INTEGER(args[i])[j] == NA_INTEGER) {
                             REAL(res)[pos++] = NA_REAL;
                         } else {
@@ -567,7 +569,7 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
         return res;
     }
 
-    case 109: { //"vector"
+    case 109: { // "vector"
         if (nargs != 2)
             return nullptr;
         if (TYPEOF(args[0]) != STRSXP)
@@ -625,6 +627,7 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
     }
 
     case 412: { // "list"
+        // "lists" at the R level are VECSXP's in the implementation
         auto res = Rf_allocVector(VECSXP, nargs);
         for (size_t i = 0; i < nargs; ++i)
             SET_VECTOR_ELT(res, i, args[i]);
@@ -719,7 +722,7 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
         return isArray(args[0]) ? R_TrueValue : R_FalseValue;
     }
 
-    case 389: { // "is.atomic" (389) nargs : 1 arg0 : double a 0
+    case 389: { // "is.atomic" (389)
         if (nargs != 1)
             return nullptr;
         switch (TYPEOF(args[0])) {
@@ -739,7 +742,7 @@ SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
         assert(false);
     }
 
-    case 384: { // "is.object" (384) nargs : 1 arg0 : double a 0
+    case 384: { // "is.object" (384)
         if (nargs != 1)
             return nullptr;
         return OBJECT(args[0]) ? R_TrueValue : R_FalseValue;
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
     return res;
 }
 
+SEXP tryFastSpecialCall(const CallContext& call, Context* ctx) {
+    SLOWASSERT(call.hasStackArgs() && !call.hasNames());
+    return nullptr;
+}
+
+SEXP tryFastBuiltinCall(const CallContext& call, Context* ctx) {
+    SLOWASSERT(call.hasStackArgs() && !call.hasNames());
+
+    static constexpr size_t MAXARGS = 32;
+    std::array<SEXP, MAXARGS> args;
+    auto nargs = call.suppliedArgs;
+
+    if (nargs > MAXARGS)
+        return nullptr;
+
+    for (size_t i = 0; i < call.suppliedArgs; ++i) {
+        auto arg = call.stackArg(i);
+        if (TYPEOF(arg) == PROMSXP)
+            arg = PRVALUE(arg);
+        if (arg == R_UnboundValue || arg == R_MissingArg ||
+            ATTRIB(arg) != R_NilValue)
+            return nullptr;
+        args[i] = arg;
+    }
+
+    switch (call.callee->u.primsxp.offset) {
+    case 90: { // c
+        if (nargs == 0)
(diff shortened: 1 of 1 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+        if (TYPEOF(arg) == PROMSXP)
+            arg = PRVALUE(arg);
+        if (arg == R_UnboundValue || arg == R_MissingArg ||
+            ATTRIB(arg) != R_NilValue)
+            return nullptr;
+        args[i] = arg;
+    }
+
+    switch (call.callee->u.primsxp.offset) {
+    case 90: { // c
```
Jakobeha: "c" should be in quotes for consistency

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+        auto res = Rf_allocVector(type, total);
+        for (size_t i = 0; i < nargs; ++i) {
+            auto len = XLENGTH(args[i]);
+            for (long j = 0; j < len; ++j) {
+                assert(pos < total);
+                if (type == REALSXP) {
+                    if (TYPEOF(args[i]) == REALSXP) {
+                        REAL(res)[pos++] = REAL(args[i])[j];
+                    } else {
+                        SLOWASSERT(NA_INTEGER == NA_LOGICAL);
```
Jakobeha: Maybe comment why this is important (here you're handling the integer and logical cases, and both are handled the same way).

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+                        if (INTEGER(args[i])[j] == NA_INTEGER) {
+                            REAL(res)[pos++] = NA_REAL;
+                        } else {
+                            REAL(res)[pos++] = INTEGER(args[i])[j];
+                        }
+                    }
+                } else {
+                    INTEGER(res)[pos++] = INTEGER(args[i])[j];
```
Jakobeha, line 565: Would rely on the assumption that `NA_INTEGER == NA_LOGICAL` (again, maybe comment you're handling both integers and logicals here)

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+                }
+            }
+        }
+        return res;
+    }
+
+    case 109: { //"vector"
```
Jakobeha: Missing a space - `// "vector"`

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+        case VECSXP:
+            return allocVector(type, length);
+        case LISTSXP:
+            if (length > INT_MAX)
+                return nullptr;
+            return allocList((int)length);
+        default:
+            return nullptr;
+        }
+        assert(false);
```
Jakobeha, line 626: Is this `assert` necessary?

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+    }
+
+    case 412: { // "list"
+        auto res = Rf_allocVector(VECSXP, nargs);
```
Jakobeha, line 631: Not sure why this creates a vector, are lists internally represented as vectors? If so, you should probably mention this in a comment.

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+        if (XLENGTH(arg) != 1)
+            return nullptr;
+
+        switch (TYPEOF(arg)) {
+        case INTSXP:
+            return INTEGER(arg)[0] == NA_INTEGER ? R_TrueValue : R_FalseValue;
+        case LGLSXP:
+            return LOGICAL(arg)[0] == NA_LOGICAL ? R_TrueValue : R_FalseValue;
+        case REALSXP:
+            return ISNAN(REAL(arg)[0]) ? R_TrueValue : R_FalseValue;
```
Jakobeha, line 661: (This might be a mistake I made), why not `REAL(arg)[0] == NA_REAL`?

rir/src/interpreter/interp.cpp
```diff
@@ -491,6 +491,264 @@ RIR_INLINE SEXP legacyCall(const CallContext& call, Context* ctx) {
...
+        default:
+            return R_FalseValue;
+        }
+        assert(false);
+    }
+
+    case 384: { // "is.object" (384) nargs : 1 arg0 : double a 0
+        if (nargs != 1)
+            return nullptr;
+        return OBJECT(args[0]) ? R_TrueValue : R_FalseValue;
```
Jakobeha, line 748: What about `isObject`?

Jakobeha (commented):
I think the code for fast builtin calls should go in a different file. Already, `interp.cpp` is huge and  does a lot of different things.

Perhaps also move the code for the logic of the more complex calls into a separate function (e.g. a function for `c`, `is.vector`, etc.). For example, we might want to concat a lot of values in the future, and then we could just reuse the function for `c`. But this is more of a personal style choice, these functions could just be unnecessary abstraction.

rir/src/interpreter/interp.cpp
```diff
@@ -785,14 +1043,83 @@ RIR_INLINE SEXP rirCall(CallContext& call, Context* ctx) {
...
+                message << "arg0 lazy";
+            else if (arg == R_MissingArg)
+                message << "arg0 missing";
+            else
+                message << " arg0 : " << type2char(TYPEOF(arg)) << " a "
+                        << (ATTRIB(arg) != R_NilValue);
+        }
+        if (!counter.count(message.str()))
+            counter[message.str()] = 0;
+        counter[message.str()]++;
```
JanJecmen, line 1075: I think I noticed this pattern somewhere else already..
The check if the key is present is not necessary. `operator[]` will insert a new default constructed value if the key is missing :)

o-:
> I think the code for fast builtin calls should go in a different file. Already, `interp.cpp` is huge and does a lot of different things.

This would be a big refactoring, with uncertain performance impacts. For example I would need to lift CallContext out of interp.h, which I am not sure is a good idea. I would like to keep it in the file and then maybe extract it in a separate PR, if that's ok.

rir/src/interpreter/interp.cpp
o-: initial value of primitive types is undefined. You are not guaranteed to get `0`.

rir/src/interpreter/interp.cpp
o-: yeah, if you forget to return or break above, it will fall through. Those are really annoying bugs, that burned me in the past. putting an asert there, is just for me to remember that I wanted to handle all the cases above, should I ever refactor that part.

rir/src/interpreter/interp.cpp
o-: what is called a "list" at the R level is a VECSXP internally. The CONS linked lists are called "pairlist" at the R level.

rir/src/interpreter/interp.cpp
o-: no, because there are many `NAN` values in the double type. Only one of them is the `NA_REAL`. I am not sure where the `NA_REAL` is used in particular. But at least this function returns true for all na's.

rir/src/interpreter/interp.cpp
o-: This is what the builtin does... straight copy from the R code. Don't ask me why it's not using isObject :)

o-:
@Jakobeha I merged this, because the benchmarking machine only checks versions merged to master. And since we have so many PRs in flight, I decided that it's probably a good idea to make good use of the benchmark machine, by merging the "most ready" PR. I'll create a new PR with the fast builtins extracted.

rir/src/interpreter/interp.cpp
JanJecmen: Not true :) Map apparently does something called value initialization
https://stackoverflow.com/a/50371768/9977179

o-: f* me. C++

o-: always something unexpected to discover, thanks :)

PR #333, opened 2019-02-19
make visitor reproducible

o-:
hopefully reduces benchmark noise


mhyee:
My only worry is that we start depending on the visitor order, but that's probably really, really unlikely.

mhyee (approved):

PR #332, opened 2019-02-19
Speeeeed cleanup

o-:
cleanup of #329 

deals with the open review comments.

additionally this commit finally moves the interpreter into the rir namespace...

PR #334, opened 2019-02-19
 implement a couple more fast builtins

o-:
also fix the implementation of vector.

based on #332

### DIFF
--- a/rir/src/interpreter/builtins.cpp
+++ b/rir/src/interpreter/builtins.cpp
@@ -3,6 +3,42 @@
 
 namespace rir {
 
+R_xlen_t asVecSize(SEXP x) {
+    if (isVectorAtomic(x) && LENGTH(x) >= 1) {
+        switch (TYPEOF(x)) {
+        case INTSXP: {
+            int res = INTEGER(x)[0];
+            if (res == NA_INTEGER)
+                break;
+            return (R_xlen_t)res;
+        }
+        case REALSXP: {
+            double d = REAL(x)[0];
+            if (ISNAN(d))
+                break;
+            if (!R_FINITE(d))
+                break;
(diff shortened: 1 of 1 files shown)


rir/src/interpreter/builtins.cpp
```diff
@@ -3,6 +3,42 @@
 
 namespace rir {
 
+R_xlen_t asVecSize(SEXP x) {
+    if (isVectorAtomic(x) && LENGTH(x) >= 1) {
```
JanJecmen, line 7: I am confused by this, shouldn't it be `LENGTH(x) == 1` and otherwise give error?

JanJecmen: never mind, looked at the gnur implementation.. still weird though

JanJecmen: Ok I think it is broken after all. This passes as a fast path if length(x) > 1, but gnur gives an error in that case

> f <- pir.compile(rir.compile(function(x) vector("character", x)))
> 
> f(4)
[1] "" "" "" ""
> f(4:10)
[1] "" "" "" ""

rir/src/interpreter/builtins.cpp
```diff
@@ -254,11 +281,107 @@ SEXP tryFastBuiltinCall(const CallContext& call, InterpreterInstance* ctx) {
...
+        if (nargs != 2)
+            return nullptr;
+
+        auto a = args[0];
+        auto b = args[1];
+
+        if (XLENGTH(a) != 1 || XLENGTH(b) != 1)
+            return nullptr;
+
+        auto combination = (TYPEOF(args[0]) << 8) + TYPEOF(args[1]);
```
JanJecmen, line 358: whoa, what a hack ;D

PR #336, opened 2019-02-21
Increase max compilation unit size and fix a number of bugs in the co…

o-:
…mpiler

Due to large methods being excluded from compilation a number of
bugs managed to sneak in. They got revealed when I tried to
increase the limits. This commit fixes:

* the RIR push_ bytecode changed the visibility flag. However pir
LdConst of course is effectless. This can lead to missed invisible
settings. This is solved by making the visible explicit in rir already.
I fixed this, because it spams the test logs with, with NULL's from
dangling else branches in `cleanEx`, which lost their invisible.
* The above fix kills some optimizations. Thus i improved the visibility
analysis a bit. But also, We don't really care about
visible that much. We just want to get it right most of the times. Thus
this commit ignores visibility in some strategic cases. For example
promises which are pure except for visibility change, we still inline
(the subsequent call will change visibility anyway). Checkpoints also
ignore visibility. Worst case we deopt with the wrong visibility
setting.
* The scope analysis would sometimes generate pretty terrible phis, that
caused the lowering in the allocator to just diverge. Excluded those
kind of phis in the verifier and fixed the scope analysis to not produce
them.
* cleanup would happily merge the block with a phi input and the block
with the actual phi. this leads to input+phi ending up in the same
block.
* banned a couple more reflective functions from being inlined

### DIFF
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -228,27 +228,27 @@ class TheVerifier {
                                   << " predecessors but "
                                   << phi->nargs() - directInputs
                                   << " more arguments than immediate preds.\n";
-                        std::cerr << "\
-This Means we created this kind of graph, where a3 is in BB3 instead of\n\
-BB0. This will be impossible to convert back to CSSA.\n\
-  .-----.                             \n\
-  | BB1 |         .-----.             \n\
-  |-----|----.----| BB0 |             \n\
-  | a1= |    |    '-----'             \n\
-  '-----'    v                        \n\
-          .-----.                     \n\
-          | BB3 |                     \n\
-          |-----|                     \n\
-          | a3= |                     \n\
-          '-----'                     \n\
-             |                        \n\
-             v                        \n\
-  .---------------------.             \n\
-  |         BB4         |             \n\
-  |---------------------|             \n\
-  | phi(BB1:a1, BB3:a3) |             \n\
-  '---------------------'             \n\
-";
+                        std::cerr << "This Means we created this kind of "
+                                     "graph, where a3 is in BB3 instead of\n"
+                                     "BB0. This will be impossible to convert "
+                                     "back to CSSA.\n"
+                                     "  .-----.                             \n"
+                                     "  | BB1 |         .-----.             \n"
+                                     "  |-----|----.----| BB0 |             \n"
+                                     "  | a1= |    |    '-----'             \n"
+                                     "  '-----'    v                        \n"
+                                     "          .-----.                     \n"
+                                     "          | BB3 |                     \n"
+                                     "          |-----|                     \n"
+                                     "          | a3= |                     \n"
+                                     "          '-----'                     \n"
+                                     "             |                        \n"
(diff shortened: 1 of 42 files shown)


rir/src/compiler/analysis/verifier.cpp
```diff
@@ -169,32 +169,109 @@ class TheVerifier {
...
+                        });
+                        std::cout << "}\n";
+
+                        std::cerr << "Error at instruction '";
+                        i->print(std::cerr);
+                        std::cerr << "'\nI have inputs from all my immediate "
+                                  << " predecessors but "
+                                  << phi->nargs() - directInputs
+                                  << " more arguments than immediate preds.\n";
+                        std::cerr << "\
```
mhyee: I think there's a stupid preprocessor trick you might be able to use to slightly clean this up. Consecutive string literals get concatenated by the preprocessor, so you could write:

```
"This means we created ... \n"
"  .-----.                             \n"
"  | BB1 |         .-----.             \n"
"  |-----|----.----| BB0 |             \n"
"  | a1= |    |    '-----'             \n"
"  '-----'    v                        \n"
```

rir/src/compiler/util/phi_placement.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // block.
+    while (true) {
+        auto& preds = cfg.immediatePredecessors(searchBlock);
+        assert(!preds.empty());
+
+        for (auto pre : preds)
+            if (!hasAllInputs(pre))
+                goto checkCandidate;
+
+        searchBlock = *preds.begin();
```
JanJecmen, line 32: How does this work? Do we know here that preds has only one block?

o-: this is the case where all predecessors have all inputs. that means the graph looks like:
```
  a ... b
   \  /
   /  \
   \  /
   searchBlock
```
so it does not matter which pred we follow. we want to move up, until at least one input branches off.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.h
```diff
@@ -11,7 +11,7 @@ namespace pir {
 
 class Rir2PirCompiler : public RirCompiler {
   public:
-    static constexpr size_t MAX_INPUT_SIZE = 1000;
+    static constexpr size_t MAX_INPUT_SIZE = 3000;
```
charig: Maybe add another env var for this...

rir/src/compiler/util/safe_builtins_list.cpp
```diff
@@ -3,6 +3,7 @@
 #include "R/Funtab.h"
 #include "R/Symbols.h"
 
+#include <iostream>
```
charig: why?

o-:
@charig maybe you want to review the rest of the commits too? I added an instruction to explicitly create contexts, so we can have RCNTXT for inlined functions. This solves many of our problems.

o-:
replaced by #345 to get the new CI tests

PR #337, opened 2019-02-21
fix vector builtin impl

JanJecmen:
Just a small bugfix:
`f <- pir.compile(rir.compile(function(x) vector("character", x)))`
would work even with `f(1:10)` but gnur gives  error in that case.


o-:
thanks

PR #338, opened 2019-02-22
[wip] Simple ranges

Jakobeha:
Optimizes "simple ranges". A simple range has the form `for (i in c:n) ...`, where `c` is a constant integer. Instead of allocating a vector, this PR rewrites this into a while loop, where i is initially set to `c` and it enumerates every integer up to `n`.

I tried to cover all cases - forwards simple ranges, backwards simple ranges, simple ranges where the end expression is a real or logical. Simple ranges where the end expression isn't a scalar (e.g. `1:"foo"`) generate errors and warnings almost the same as gnuR, but not exactly, although they should always fail iff gnuR would fail.


o-:
superseeded by #387 i guess

PR #343, opened 2019-02-23
Use gitlab ci, reduce testload on travis

o-:
the ubuntu image we use on the gitlab ci has a newer cppcheck version. need to fix some things for it

### DIFF
--- a/rir/src/ir/insns.h
+++ b/rir/src/ir/insns.h
@@ -2,8 +2,6 @@
 #error "DEF_INSTR must be defined before including insns.h"
 #endif
 
-#include "simple_instruction_list.h"
-
 // DEF_INSTR(name, imm, pop, push, pure)
 
 /**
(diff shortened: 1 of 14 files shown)


rir/src/ir/insns.h
```diff
@@ -2,8 +2,6 @@
 #error "DEF_INSTR must be defined before including insns.h"
 #endif
 
-#include "simple_instruction_list.h"
```
mhyee, line 5: Was this breaking cppcheck? Because removing it just from here means this file is no longer in sync with other files that include `simple_instruction_list.h`.

o-: yes, cppcheck could not even parse it. and honestly after 3 layers of macros i decided that the little code duplication (where every error will be caught by warnings, since we use complete switch statements) is worth the improved comprehensibility (also for the static analysis).

mhyee: In that case, I propose we completely remove this macro/include. (Sorry @Jakobeha.)

o-: I thought it's used in some other places where it is kind of handy, not? I am not necessarily against the macro. just here its a bit unfortunate because we are mixing the second order kind of macro-list with the include kind of macro-list.

mhyee: Hmm. I guess the point was to have a single place to add the simple instructions, and then it would update everywhere.

Now it updates everywhere except for here. So now that you mention it, it's probably good enough.

So no need to change anything else. Sorry!

PR #344, opened 2019-02-23
Fix edge-split check in verifier

mhyee:
The first commit just moves the check outside the `if`, because we want to check non-empty BBs as well.

The second commit hopefully makes the error message more readable. I found the original message unhelpful, and it got worse after the bugfix because it could say something like "BB 3 merges into 4 and branches into 4 at the same time." Now this is what it says:

```
BB5 is a merge node, but predecessor BB3 has two successors (BB4 and BB5):
 n      3
  \   /   \
    5      4
```


o-:
thanks

PR #345, opened 2019-02-25
Add context for inlined functions and fix a number of other bugs

o-:
Increases max compilation unit size, which uncovered all those bugs addressed in this PR.

See the individual commits.

Replaces #336

### DIFF
--- a/rir/src/compiler/analysis/query.cpp
+++ b/rir/src/compiler/analysis/query.cpp
@@ -36,7 +36,9 @@ bool Query::noEnvSpec(Code* c) {
 
 bool Query::pure(Code* c) {
     return Visitor::check(c->entry, [](Instruction* i) {
-        return !i->hasEffect() && !i->changesEnv();
+        // Yes visibility is a global effect. We try to preserve it. But geting
+        // it wrong is not a strong correctness issue.
+        return !i->hasEffectIgnoreVisibility() && !i->changesEnv();
     });
 }
 
--- a/rir/src/compiler/analysis/unnecessary_contexts.h
+++ b/rir/src/compiler/analysis/unnecessary_contexts.h
@@ -1,0 +1,59 @@
+#ifndef PIR_UNNECESSARY_CONTEXTS_H
+#define PIR_UNNECESSARY_CONTEXTS_H
+
+#include "../util/cfg.h"
+#include "abstract_value.h"
+#include "generic_static_analysis.h"
+
+namespace rir {
+namespace pir {
+
+struct UnnecessaryContextsState : public AbstractUnique<PushContext> {
+    bool needed = false;
+    AbstractResult merge(const UnnecessaryContextsState& other) {
+        AbstractResult res = AbstractUnique<PushContext>::merge(other);
+        if (get() && !needed && other.needed) {
+            needed = true;
+            res.update();
+        }
+        return res;
+    }
+};
+
+class UnnecessaryContexts : public StaticAnalysis<UnnecessaryContextsState> {
+  public:
+    ClosureVersion* code;
+    UnnecessaryContexts(ClosureVersion* cls, LogStream& log)
+        : StaticAnalysis("UnnecessaryContexts", cls, cls, log), code(cls) {}
+
+    AbstractResult apply(UnnecessaryContextsState& state,
+                         Instruction* i) const override {
+        if (auto p = PushContext::Cast(i)) {
+            state.set(p);
+            return AbstractResult::Updated;
+        } else if (CallInstruction::CastCall(i) || CallBuiltin::Cast(i) ||
+                   Checkpoint::Cast(i)) {
+            // Contexts are needed for non-local returns, reflection and
+            // deoptimization.
+            state.needed = true;
+            return AbstractResult::Updated;
+        } else if (auto p = PopContext::Cast(i)) {
+            if (state.get()) {
+                assert(state.get() == p->push());
+                state.clear();
+                return AbstractResult::Updated;
+            }
+        }
+        return AbstractResult::None;
+    };
+
+    bool canRemove(PopContext* i) {
+        auto res = StaticAnalysis::at<PositioningStyle::BeforeInstruction>(i);
+        return res.get() && res.get() == i->push() && !res.needed;
+    }
+};
+
+} // namespace pir
+} // namespace rir
+
+#endif
--- a/rir/src/compiler/analysis/unnecessary_contexts.h
+++ b/rir/src/compiler/analysis/unnecessary_contexts.h
@@ -31,10 +31,10 @@ class UnnecessaryContexts : public StaticAnalysis<UnnecessaryContextsState> {
         if (auto p = PushContext::Cast(i)) {
             state.set(p);
             return AbstractResult::Updated;
-        } else if (CallInstruction::CastCall(i) || CallBuiltin::Cast(i) ||
-                   Checkpoint::Cast(i)) {
-            // Contexts are needed for non-local returns, reflection and
-            // deoptimization.
+        } else if (CallInstruction::CastCall(i) || CallBuiltin::Cast(i)) {
+            // Contexts are needed for non-local returns and reflection. On
+            // deoptimization we can synthesize them, thus none needed for
+            // checkpoints.
             state.needed = true;
             return AbstractResult::Updated;
         } else if (auto p = PopContext::Cast(i)) {
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -176,32 +176,125 @@ class TheVerifier {
             assert(i == bb->last() &&
                    "Only last instruction of BB can have controlflow");
 
-        Phi* phi = Phi::Cast(i);
-        i->eachArg([&](const InstrArg& a) -> void {
-            auto v = a.val();
-            auto t = a.type();
-            if (auto iv = Instruction::Cast(v)) {
-                if (phi) {
-                    if (slow &&
-                        !cfg(bb->owner).isPredecessor(iv->bb(), i->bb())) {
+        if (auto phi = Phi::Cast(i)) {
+            size_t directInputs = 0;
+            phi->eachArg([&](BB* input, Value* v) {
+                if (auto iv = Instruction::Cast(v)) {
+                    if (input == phi->bb()) {
                         std::cerr << "Error at instruction '";
                         i->print(std::cerr);
                         std::cerr << "': input '";
                         iv->printRef(std::cerr);
-                        std::cerr << "' does not come from a predecessor.\n";
+                        std::cerr << "' has the phi block as input block\n";
                         ok = false;
                     }
-                } else if ((iv->bb() == i->bb() &&
-                            bb->indexOf(iv) > bb->indexOf(i)) ||
-                           (iv->bb() != i->bb() && slow &&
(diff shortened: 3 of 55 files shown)


rir/src/compiler/analysis/query.cpp
```diff
@@ -17,7 +17,9 @@ bool Query::noEnv(Code* c) {
 
 bool Query::pure(Code* c) {
     return Visitor::check(c->entry, [](Instruction* i) {
-        return !i->hasEffect() && !i->changesEnv();
+        // Yes visibility is a global effect. We try to preserve it. But geting
```
charig, line 39: Great comment :). Still, I'd move it to the definition of `hasEffectIgnoreVisibility`. And put sth more like a sth like `TODO: devise a good story for supporting visibility without incuring in considerable overheads`

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -176,32 +176,125 @@ class TheVerifier {
...
 
-        Phi* phi = Phi::Cast(i);
+        if (auto phi = Phi::Cast(i)) {
+            size_t directInputs = 0;
+            phi->eachArg([&](BB* input, Value* v) {
+                if (auto iv = Instruction::Cast(v)) {
+                    if (input == phi->bb()) {
+                        std::cerr << "Error at instruction '";
+                        i->print(std::cerr);
+                        std::cerr << "': input '";
```
charig, line 188: I do not get this new block.
1. Couldn't this happen in a loop block?
2. The error message: `input has the phi block as input block`. What is the input block of an argument? The block where it is defined? I'd say input lives in the same block as the phi or sth like that.

rir/src/compiler/analysis/visibility.cpp
```diff
@@ -0,0 +1,65 @@
...
+    case Tag::CallBuiltin:
+    case Tag::CallSafeBuiltin:
+        int flag;
+        if (auto c = CallBuiltin::Cast(i)) {
+            flag = getFlag(c->builtinId);
+        } else {
+            flag = getFlag(CallSafeBuiltin::Cast(i)->builtinId);
+        }
+
+        if (flag < 2) {
```
charig: flag is only for visibility? If yes change to getVisibility. Otherwise, maybe for code readability instead of the `2` we can use a static mask defined somewhere for gathering the proper bits

rir/src/compiler/opt/inline.cpp
```diff
@@ -97,6 +97,9 @@ class TheInliner {
                             return false;
                         }
                     }
+                    if (CallInstruction::CastCall(i)) {
```
charig: So, we are finally not inlining any function that contains a call?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -879,24 +881,11 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 break;
             }
 
-            case Tag::CastType: {
```
charig, line 882: Good catch moving this to the verifier!

rir/src/interpreter/interp.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include <assert.h>
 #include <deque>
+#include <set>
```
charig, line 14: why?

rir/src/interpreter/interp.cpp
```diff
@@ -346,6 +346,46 @@ static void loopTrampoline(Code* c, InterpreterInstance* ctx, SEXP* env,
     Rf_endcontext(&cntxt);
 }
 
+const static SEXP inlineContextTrampolineDeoptMarker = (SEXP)0x7008;
```
charig: inside?

rir/src/compiler/analysis/unnecessary_contexts.h
```diff
@@ -0,0 +1,56 @@
...
+                state.clear();
+                return AbstractResult::Updated;
+            }
+        }
+        return AbstractResult::None;
+    };
+
+    bool canRemove(PopContext* i) {
+        auto res = StaticAnalysis::at<PositioningStyle::BeforeInstruction>(i);
+        return res.get() && res.get() == i->push() && !res.needed;
```
charig, line 52: the `res.get() seems a little confusing to me. Maybe an assert()?

rir/src/compiler/opt/context.cpp
```diff
@@ -0,0 +1,55 @@
...
+        return;
+
+    assert(toRemove.size() % 2 == 0);
+
+    Visitor::run(function->entry, [&](BB* bb) {
+        auto ip = bb->begin();
+        while (ip != bb->end()) {
+            auto next = ip + 1;
+            auto instr = *ip;
+
```
charig, line 42: e -> context?

Maybe an assert for even or 0 after the visitor?

rir/src/compiler/pir/instruction.h
```diff
@@ -690,11 +690,14 @@ class Branch
...
+class Return
+    : public FixedLenInstructionWithEnvSlot<Tag::Return, Return, 2,
+                                            Effect::Order, EnvAccess::Capture,
+                                            Controlflow::Exit> {
   public:
-    explicit Return(Value* ret)
-        : FixedLenInstruction(PirType::voyd(), {{PirType::val()}}, {{ret}}) {}
+    explicit Return(Value* ret, Value* env)
+        : FixedLenInstructionWithEnvSlot(PirType::voyd(), {{PirType::val()}},
+                                         {{ret}}, env) {}
```
charig: I do not remember anything in this commit related to the addition of the env here...

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -1018,6 +1020,22 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+            case Tag::PushContext: {
+                if (!pushContexts.count(instr))
+                    pushContexts[instr] = cs.mkLabel();
+                cs << BC::pushContext(pushContexts.at(instr));
+                break;
+            }
+
+            case Tag::PopContext: {
+                auto push = PopContext::Cast(instr)->push();
+                if (!pushContexts.count(push))
```
charig, line 1041: Is it possible that we still have not processed the push?

rir/src/interpreter/interp.cpp
```diff
@@ -346,6 +346,46 @@ static void loopTrampoline(Code* c, InterpreterInstance* ctx, SEXP* env,
...
+    RCNTXT cntxt;
+    // The first env should be the callee env, but that will be done by the
+    // callee. We store sysparent there, because our optimizer may actually
+    // delay instructions into the inlinee and might assume that we still have
+    // to outer env.
+    initClosureContext(ast, &cntxt, sysparent, sysparent,
+                       symbol::delayedArglist, op);
+
+    auto trampoline = [&]() {
+        if ((SETJMP(cntxt.cjmpbuf))) {
```
charig, line 394: I have no idea but happens inside this if but it is also in the `rirCallTrampoline_` so ok...

rir/src/compiler/opt/inline.cpp
```diff
@@ -257,7 +259,29 @@ class TheInliner {
                         }
                     });
 
-                    Value* inlineeRes = BBTransform::forInline(copy, split);
+                    auto inlineeRet = BBTransform::forInline(copy, split);
```
charig, line 262: inlinee
inilineReturn
inlineReturnBB?

rir/src/interpreter/interp.cpp
```diff
@@ -3052,11 +3124,15 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP* env,
...
 eval_done:
     if (synthesizeFrames) {
+        SEXP e = *env;
         while (!synthesizeFrames->empty()) {
             FrameInfo* f = synthesizeFrames->front();
             synthesizeFrames->pop_front();
             SEXP res = ostack_pop(ctx);
-            SEXP e = ostack_pop(ctx);
+            // Inlined functions may or may not have contexts
+            if (e == R_GlobalContext->cloenv)
```
charig: I do not clearly see why this condition couldn't hold while we did not create an context?

rir/src/interpreter/instance.h
```diff
@@ -172,14 +172,19 @@ class Locals final {
   private:
     R_bcstack_t* base;
     unsigned localsCount;
+    bool existingLocals;
```
charig, line 175: Do not fully get this bool? Does it have any relation with inlining? When are the locals already reserved on the stack?

rir/src/ir/BC.cpp
```diff
@@ -100,6 +100,7 @@ BC_NOARGS(V, _)
     case Opcode::br_:
     case Opcode::brtrue_:
     case Opcode::beginloop_:
+    case Opcode::push_context_:
```
charig, line 104: Why pop_contenxt_ is not considered in these switches?

charig (approved):
lgtm modulo ci

rir/src/compiler/analysis/verifier.cpp
o-: 1. only if we allowed critical edges

rir/src/compiler/analysis/visibility.cpp
o-: this is R terminology, would rather not invent new names for it. see names.c

rir/src/compiler/opt/inline.cpp
o-: we are. read the next line. it's just that we need to add push/pop context instructions

rir/src/interpreter/interp.cpp
o-: `grep std::set rir/src/interpreter/interp.cpp` ?

rir/src/compiler/analysis/unnecessary_contexts.h
o-: doing a null test by relying on `nullptr <==> false` is very idiomatic. assert? we are not aborting if it's null, just return false.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: probably not. but didn't want to unnecessarily enforce a particular order to dump the basic blocks...

rir/src/interpreter/interp.cpp
o-: stjmp-longjmp is used for non-local returns. You can jump from a nested function to this place. And you can also restart a function instead of just returning (used for continuations)

rir/src/interpreter/interp.cpp
o-: superseeded by changes

rir/src/interpreter/instance.h
o-: for inlinedContexts we recursively call the interpreter function, but we just continue executing as before. in particular we continue using the same locals and the stack. the additional interpreter execution is just to sneak a context trampoline in between, to bind the setjmps.

rir/src/ir/BC.cpp
o-: it in the simple instruction macro that creates the entries...

PR #347, opened 2019-02-26
[wip] for loop box reuse

JanJecmen:
Try to reuse the sexp used for `for` loop variable.

Adds a new bytecode (basically a combinded extract and stvar) just for the `for` loops.
Adds one slot to the stack that holds the box for the loop variable, in the first iteration new vector will be allocated, later that one will be reused.

Edit:
- Performance is dubious
- Looks like it inhibits some pir optimizations


o-:
@JanJecmen so what's the status? you say perf is dubious, do you still intend to merge? why don't you use the existing pir instructions and only implement it on the rir side? I mean you can just compile to the current version in rir_2_pir and lower to the more dense version in pir_2_rir. this way you have the fast rir version, but you don't make it impenetrable to pir

JanJecmen:
The issue is, if I make it penetrable for pir, then I will basically get the old version of the for loop after pir->rir
Anyway, I'd say we don't merge this now, I cannot see any performance difference on anything that is not `for (i in 1:x) {}`

PR #352, opened 2019-03-04
Profile env

charig:


JanJecmen:
I think clang fromat failed here pretty hard..

JanJecmen:
Won't this get rid of all the binop speculation in pir?

charig:
I disabled the clang formatting to avoid the mess.

And no, it should only disable the profiling when the env var says that.

rir/src/ir/Compiler.cpp
```diff
@@ -86,8 +86,13 @@ class CompilerContext {
     FunctionWriter& fun;
     Preserve& preserve;
 
+    bool profile;
+
     CompilerContext(FunctionWriter& fun, Preserve& preserve)
-        : fun(fun), preserve(preserve) {}
+        : fun(fun), preserve(preserve),
+          profile(
```
o-: Why would you want to read this variable on every invocation? can you make this a static member of the compiler, instead of an instance variable of the context?

JanJecmen:
> And no, it should only disable the profiling when the env var says that.

Oh, that's not what I meant. When it's disabled, we will generate different pir code, ie. there will be no speculation on types of binops (if I didn't miss anything)

In general it's quite wild as it is now:
if I explicitly do `pir.compile(rir.compile(f))` and if I do `f(); f(); f(); ...` then the pir will be different..

charig:
No, the profiling is only disabled if you explicitly run the code with `RIR_PROFILING=off`. By default it is enabled.

JanJecmen:
Sure, my point is, I thought this flag would be used to run the benchmarks, but then the results would not be credible, because we would measure different code.
But maybe I misunderstood..

charig:
It is to run benchmarks yes.
Personally I think it is the other way. When we compile from PIR back to RIR we remove all profiling. 
Therefore, we are having a "fake speedup" if profiling is significant. Our real speedups are when comparing to rir without all the addons that are there only for "helping" pir. 
On the other hand, it is also unfair to compare rir against `gnu-r` if rir profiles and gnu-r don't.
So, essentially the flag is there for 2 things: 1) Our RIR baseline for benchmarks should not include profiling 2) Measure once the profiling overhead which is exactly this: http://rflies.rir.o1o.ch/changes/?tre=10&rev=882dcee686a3fd009caa7c72ceb2b6c2fb5e931f&exe=3&env=1

JanJecmen:
Ok, fair.
My thinking was, pir needs the profiling, but you are right that baseline rir should not include it

PR #353, opened 2019-03-05
Flaky tests

o-:


rir/src/compiler/opt/inline.cpp
```diff
@@ -200,14 +200,11 @@ class TheInliner {
...
-                            if (auto mk = MkArg::Cast(a)) {
+                            if (MkArg::Cast(a)) {
                                 // We need to cast from a promise to a lazy
                                 // value
-                                auto cast = new CastType(
-                                    a, RType::prom,
-                                    mk->isEager()
-                                        ? mk->eagerArg()->type.forced()
-                                        : ld->type);
+                                auto cast =
```
o-, line 216: @mhyee that was the bug you where seeing.

when we inline we do the following:
```
   prom  %1 = mkarg(...)
   val^   %2 = Cast(val^, %1)
```

we add a cast to convert a *promise-as-a-value* to a *lazy-value*. The former is a promise in value position, ie. we mean the promise itself. The later is a lazy value, ie. we mean the value we get out of the promise. That cast is needed, because normally the call instruction does this implicit "conversion".

Now I tried to be clever and already propagate the underlying type, if the argument was eager. But that caused the compiler to think that the result of `Cast` is a value without promise wrapper, when in fact it has a promise wrapper, that needs to be removed.

Most of the time this does not cause any issues, since many operations just implicitly evaluate promises. But the `asbool_` instruction does not do so...

mhyee:
I stumbled upon this because I changed the following:

https://github.com/reactorlabs/rir/blob/7c13298568cf587f933e19fcade1ac8be70dbd85/rir/src/compiler/pir/instruction.h#L902L905
```cpp
    Is(uint32_t sexpTag, Value* v)
        : FixedLenInstruction(PirType(RType::logical).scalar(),
                              {{PirType::val()}}, {{v}}), sexpTag(sexpTag) {}
```

Should the type be `NativeType::test` instead of `PirType(RType::logical).scalar()`? When I made that change, that's when the `asbool_` thing showed up.

o-:
> Should the type be NativeType::test instead of PirType(RType::logical).scalar()?

if you look at the implementation of `is_` it only returns true/false, so test is more precise. but of course then the problem is that if you want to use the instruction in a situation where an rtype is requested, you have a problem... it seems to me that this instruction could be either or. certainly `PirType(RType::logical).scalar()` does not feel wrong (and I suspect that it would not break on top of this branch anymore). otoh, you can't directly use it as branch input anymore (it will add a useless asbool).

in some cases we use the same instruction but just change the type. the type of an instruction is mutable and this seems like one of the instructions where i'd say, make the type either or, dending on where you want to feed it to.

PR #356, opened 2019-03-06
improve visibility analysis

o-:
Visibility analysis keeps track of observed visibility changing
instructions. A visibility change is observed, only if its effect
lasts until a function exit point.

If we merge two states, then we conservatively assume that either
branches last visibility changing instruction could be observed.
As soon as we have a visibility changing instruction, all previous
visibility changes are not observed anymore.

### DIFF
--- a/rir/src/compiler/analysis/visibility.cpp
+++ b/rir/src/compiler/analysis/visibility.cpp
@@ -4,19 +4,18 @@
 namespace rir {
 namespace pir {
 
-AbstractResult VisibilityAnalysis::apply(CurrentVisibility& vis,
+AbstractResult VisibilityAnalysis::apply(LastVisibilityUpdate& vis,
                                          Instruction* i) const {
-    // Default: visible
-    if (!code->entry->isEmpty() && i == *code->entry->begin())
-        vis.state = CurrentVisibility::Visible;
-
+    AbstractResult res;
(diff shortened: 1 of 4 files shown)


rir/src/compiler/analysis/visibility.cpp
```diff
@@ -4,30 +4,26 @@
...
     case Tag::LdVarSuper:
     case Tag::Extract1_1D:
     case Tag::Extract2_1D:
     case Tag::Extract1_2D:
     case Tag::Extract2_2D:
-        if (vis.state != CurrentVisibility::Visible) {
-            vis.state = CurrentVisibility::Visible;
-            return AbstractResult::Updated;
-        }
+        isVisibilityChanging();
```
JanJecmen, line 26: Are these instructions the only ones that change visibility? Could you use `Instruction::mightChangeVisibility` here?

JanJecmen (approved):

rir/src/compiler/analysis/visibility.cpp
o-: unfortunately not. I can only be sure that an instruction is that last visible change to the visibility flag if I am certain that it changes visibility. the mightChange query doesn't help. i don't see an easy way to get does certainly change visibility at the instruction level...

PR #358, opened 2019-03-08
Stack shuffling

JanJecmen:
-- just to run the gitlab tests now --

- Use the cleaner api to access analysis results instead of the brittle `after(previous)`
- Fix the bug with mandelbrot that crashed the allocator
- TODO: refactor the instruction argument loading to get rid of dumb code (eg. `swap swap`..)

### DIFF
--- a/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
@@ -70,9 +70,8 @@ class SSAAllocator {
 
     void computeStackAllocation() {
 
-        // For now, just flip a coin on where to put it
         static auto toStack = [](Instruction* i) -> bool {
-            return Phi::Cast(i) || !MkEnv::Cast(i);
+            return i->numberOfUses() < 2;
         };
 
         std::unordered_set<Value*> phis;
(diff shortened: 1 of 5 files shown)


JanJecmen:
The failing test was the assert `Cannot find outermost function context` in `interp.cpp` during deoptimization.. I managed to get the fail also on master, so hopefully it's not something caused by this pr
Possibly fixed by fixing the protect bug in `mk_env_` but not really sure about that

edit: not the case :)

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -70,9 +70,8 @@ class SSAAllocator {
 
     void computeStackAllocation() {
 
-        // For now, just flip a coin on where to put it
         static auto toStack = [](Instruction* i) -> bool {
-            return Phi::Cast(i) || !MkEnv::Cast(i);
+            return i->numberOfUses() < 2;
```
o-: I am a bit worried that you also change the allocation heuristics in the same commit. I would like to see that it works now with the old heuristics, and then change them later.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -1232,88 +1232,84 @@ void Pir2Rir::lower(Code* code) {
...
+                                    if (Instruction::Cast(val)->bb() == bb) {
+                                        assert(!done);
+                                        update.emplace_back(val, pred);
+                                        done = true;
+                                    }
+                                });
+                                if (done)
                                     return false;
+                                // Check if some phi in this block has some arg
+                                // same
```
o-: does not parse

o-:
> The failing test was the assert Cannot find outermost function context in interp.cpp during deoptimization.

I don't succeed to reproduce either. Can you give me instructions on how to repro that one?

rir/tests/pir_regression8.R
```diff
@@ -0,0 +1,23 @@
+foo <- function() {
```
mhyee, line 1: There's a tab/space mixup here?

PR #359, opened 2019-03-08
stubs are guarded, they cannot get tainted by force

o-:
currently this has slightly negative perf. I did not check why. since that was the whole idea of the elided stubs it would be insteresting to find out why. we might still want to merge?

### DIFF
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -260,6 +260,12 @@ AbstractResult ScopeAnalysis::apply(ScopeAnalysisState& state,
     if (!handled) {
         if (i->hasEnv()) {
             bool envIsNeeded = i->hasEnv();
+
+            if (auto mk = MkEnv::Cast(i->env())) {
+                if (mk->stub)
+                    envIsNeeded = false;
+            }
+
             // Already exclude the case where an operation needs an env only for
             // object arguments, but we know that none of the args are objects.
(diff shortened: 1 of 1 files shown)


charig (approved):

rir/src/compiler/analysis/scope.cpp
```diff
@@ -260,6 +260,12 @@ AbstractResult ScopeAnalysis::apply(ScopeAnalysisState& state,
     if (!handled) {
         if (i->hasEnv()) {
             bool envIsNeeded = i->hasEnv();
+
```
charig, line 263: I am for this change. It is a fix actually. But there is one thing to consider. 
Without this change a deopt was actually not needed. So, we could create stub always in case they are really faster than standard environments. After this change, a deopt is needed in case an instruction creates an environment.

charig (approved):

rir/src/compiler/analysis/scope.cpp
o-: good point. do you see a way of checking that?

charig: At this point we are actually always deopting -> this change is sound I'd go for it.
What I had in mind was to explore compiling 2 versions and have like an "optimized" deopt that just jump to the version that expects a real environment but does not incurs in many materializations. Was a brainstorming idea but it may be unfeasible or a dead end.

PR #361, opened 2019-03-08
Fix some issues with scope resolution

charig:
* Although we can not run the safe version of a builtin we know that the safe builtins will not taint the environment
* Instructions should not have reflective effects if we dynamically know they not need an environment


o-:
superseeded by #365

PR #362, opened 2019-03-09
Loop invariant

charig:
Basic loop invariant code motion optimization. 
At this point it only hoists ldFuns and ldVars from loops whenever that is safe to do. 
Essentially, whenever we can prove the loop body will not change the environment.
All the loop detection is based on @mhyee's WIP on loop peeling. We should refactor afterwards to avoid repetition.  

NOTE:
It only works when there is already a pre header for the loop. 

TODOs:
* Create preheader when there is non.
* Integrate with hoist instruction.
* Add support to also hoist mk_envs.


rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    BB* outOfLoopPredecessor(const CFG& cfg) {
+        std::vector<BB*> diff;
+        std::set_difference(cfg.immediatePredecessors(header).begin(),
+                            cfg.immediatePredecessors(header).end(),
+                            body.begin(), body.end(),
+                            std::inserter(diff, diff.end()));
+
+        return diff.front();
```
mhyee: I don't think this is right? What if there are multiple (non-loop) nodes that precede the header?

I think we want the _loop preheader_, which is the single node that's the only way to enter the loop. It doesn't always exist, so sometimes you have to create it. But the point is that anything hoisted out of the loop should go there.

I'll see if I can add it to the loop API.

charig: Yes, definetely. There was an assert in the first version I do not know why I removed it. Ping me if you add the code for the preheader.

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -63,6 +66,35 @@ class LoopDetection {
             o->isInnermost_ = false;
         }
 
+        /*
+         * TODO: finds a preheader when there is one.
+         * We should create a preheader when there is non.
+         */
+        BB* preheader(const CFG& cfg) {
+            std::vector<BB*> outOfLoopPredecessor;
```
mhyee: `BBList`

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -63,6 +66,35 @@ class LoopDetection {
...
+            for (auto pred : cfg.immediatePredecessors(header())) {
```
mhyee: My personal habit is to write `const auto&` unless there's a reason not to. `&` is more important to avoid making unnecessary copies, but here you're just copying pointers, so it should be cheap.

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -63,6 +66,35 @@ class LoopDetection {
             o->isInnermost_ = false;
         }
 
+        /*
+         * TODO: finds a preheader when there is one.
+         * We should create a preheader when there is non.
+         */
+        BB* preheader(const CFG& cfg) {
```
mhyee, line 73: I think it makes more sense to find preheaders inside the `LoopDetection` constructor, where you already have the `cfg`. It'll also be easier when we actually create the preheader.

It might also be useful to cache the preheader, i.e. add a `preheader_` field to `Loop`.

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -63,6 +66,35 @@ class LoopDetection {
...
+                if (!body_.count(pred))
+                    outOfLoopPredecessor.push_back(pred);
+            }
+
+            return outOfLoopPredecessor.size() != 1
+                       ? nullptr
+                       : outOfLoopPredecessor.front();
+        }
+
+        bool holdsPropery(const InstrActionPredicate& action) {
```
mhyee: `holdsProperty`. Or maybe `check`, like the Visitor function?

rir/src/compiler/analysis/query.cpp
```diff
@@ -15,6 +15,17 @@ bool Query::noEnv(Code* c) {
                           [](Instruction* i) { return !MkEnv::Cast(i); });
 }
 
+bool Query::lookupOutOfLoopEnv(Code* c) {
```
mhyee: This seems to be unused now? In any case, the TODO below needs to be updated.

rir/src/compiler/util/cfg.h
```diff
@@ -23,6 +23,7 @@ class CFG {
     bool isMergeBlock(BB* a) const;
     const BBList& immediatePredecessors(BB* a) const;
     const BBList& exits() const { return exits_; }
+    size_t size() const { return predecessors_.size(); }
```
mhyee: I don't think this is needed any more.

rir/src/utils/configurations.cpp
```diff
@@ -109,10 +109,11 @@ void Configurations::defaultOptimizations() {
     // ==== Phase 4) Final round of default opts
     for (size_t i = 0; i < 3; ++i) {
         addDefaultOpt();
+
         optimizations.push_back(new pir::CleanupCheckpoints());
-    }
+        optimizations.push_back(new pir::LoopInvariant());
 
-    phasemarker("Phase 4: finished");
+        phasemarker("Phase 4: finished");
```
mhyee: `phasemarker` should be outside the for loop.

rir/src/utils/configurations.cpp
```diff
@@ -109,10 +109,11 @@ void Configurations::defaultOptimizations() {
     // ==== Phase 4) Final round of default opts
     for (size_t i = 0; i < 3; ++i) {
         addDefaultOpt();
+
         optimizations.push_back(new pir::CleanupCheckpoints());
-    }
+        optimizations.push_back(new pir::LoopInvariant());
```
mhyee: Is this where we want to do the loop invariant code motion? I feel like it should be done earlier...

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -125,6 +125,13 @@ class PASS(OptimizeContexts);
 
 class PASS(DeadStoreRemoval);
 
+/*
+ * At this point, loop code invariant mainly tries to hoist ldFun operations
+ * outside the loop in case it can prove that the loop body will not change
+ * the binding
+ */
+class PASS(LoopInvariant);
```
mhyee, line 133: Nitpick: `LoopInvariantCodeMotion` better describes what's happening (it's a verb and not a noun), even though it's a lot longer of a name.

EDIT: I guess we're being inconsistent with the naming convention of our passes. Most of them are verbs, but then there's things like `EagerCalls` and `LoadElision`.

rir/src/compiler/test/PirCheck.cpp
```diff
@@ -135,6 +135,15 @@ static bool testOneNot(ClosureVersion* f) {
     return numNots == 1;
 }
 
+static bool testLookupOutOfLoop(ClosureVersion* f) {
```
mhyee: I'm not sure what this is supposed to test for? From the name, it seems like it's making sure the lookup (`LdFun`?) has been hoisted out of the loop. But the implementation of this function just looks for any `LdFun`.

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "../util/cfg.h"
+#include "../util/safe_builtins_list.h"
+#include "../util/visitor.h"
+#include "R/r.h"
+#include "pass_definitions.h"
+
+namespace rir {
+namespace pir {
+
+typedef std::pair<BB*, BB*> BackEdge;
```
mhyee: Not needed anymore.

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,97 @@
...
+typedef std::function<bool(Instruction*)> InstrActionPredicate;
```
mhyee: Not needed anymore.

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,97 @@
+#include "../analysis/loop_detection.h"
+#include "../pir/pir_impl.h"
+#include "../transform/bb.h"
```
mhyee: Can any of these includes be cleaned up, since there was some merging and refactoring?

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (CallSafeBuiltin::Cast(i))
+        return true;
+
+    if (CallInstruction::CastCall(i))
+        return false;
+
+    return !(i->mayUseReflection());
+};
+
+bool isSafeToHoistLoads(LoopDetection::Loop& loop) {
```
mhyee: `const`?

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -0,0 +1,97 @@
...
+    return loop.holdsPropery(noEnvironmentTainting);
```
mhyee: You shouldn't need to make `noEnvironmentTainting` a lambda, to be able to pass it as an argument to `loop.holdsProperty`. On the other hand, it's only used here, so maybe you could just write it inline:

```c++
loop.holdsProperty([](Instruction *i) {
    // ...
});
```

Or you could do the same as below, and define `auto noEnvironmentTainting = ...` inside the `isSafeToHoistLoads` function. Just a matter of style.

rir/src/compiler/analysis/loop_detection.h
mhyee: Yeah, I'm looking at where you construct a `LoopDetection` class, and there you have to compute the CFG again... which `LoopDetection`'s constructor also has to do.

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -63,6 +66,35 @@ class LoopDetection {
...
+            auto answer = true;
```
mhyee, line 86: My personal habit is to write out types, unless they're too long or it's redundant.

mhyee (commented):
Mostly minor things. I think there are only two non-trivial comments:

* Moving `preheader` from `Loop` to the `LoopDetection` constructor.
* When to schedule the loop-invariant code motion pass. I haven't really thought about when it should run, i.e. what optimizations need to run before it, and if any optimizations afterwards can take advantage of it.

I made a comment about using `const auto&`, which I use as often as I can out of habit. It's really easy to shoot yourself in the foot with C++ make make a whole bunch of expensive copies, so I try to prevent that. I don't think that happens at all in this PR, so you can leave things as just `auto` if you don't feel like making the changes.

rir/src/compiler/analysis/loop_detection.h
charig: Yes, recomputing the CFG is a little ugly I realized it at the moment and agree.
But, I still think we should compute the preheader lazily. Otherwise, our loop will depend on the fact that the CFG up to its header remains immutable which I see as a pretty strong constraint.

rir/src/utils/configurations.cpp
charig: Yes, I put it late because we were discussing that the loop detection may be slow. I add it now near hoistInstruction with which this opt should be integrated.

rir/src/compiler/opt/loop_invariant.cpp
charig: I include `noEnvironmentTainting` inside `isSafeToHoistLoads` but I do not inline it because I think giving it a name makes `isSafeToHoistLoads` easier to read.

mhyee (approved):
Thanks! Please squash and merge (since there's a lot of wip commits).

o-:
hey, sorry, but [this](https://github.com/reactorlabs/rir/pull/362/files#diff-70ef74c2e47842295d0a7f76db1ab1a1R11) loop check is not sound at all.

1. ldfun can have side effects, you can't just hoist it
2. you use the safeOnObjects builtins list without proving that arguments are not objects
3. `mayHaveReflection` is not enough. there is a predicate `changesEnv` you need to check. because if anything changes an env inside the loop, then the binding can change.

here is a testcase for (1):
```
f <- function(j){
    j <- j
    while (j < 2) {
        c(j)
        j <- j + 1 
    }   
}
g <- function(i) {
  if (i) 
    f(0)
  else
    f(2)
}
g(TRUE)
g(FALSE)
g(TRUE)
delayedAssign("c", fail())
g(FALSE)
```
this breaks on master.

PR #365, opened 2019-03-12
refactor effect system

o-:
This adds the following two changes:

1. convert effect lattice (where stronger effect includes all previous
effects) with a set of effects.
3. make effects per instruction instance. So we can keep fine grained
taps on effects of single instructions (depending on their context).

This also fixes some mislabeled instructions. For example binops are
certainly not pure. When they have object args, they can doe whatever
and otherwise they can still print warnings and errors.

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -90,7 +90,6 @@ enum class EnvAccess : uint8_t {
 // This is a trivial lattice, any effect with higher order contains all the
 // lower order effects.
 enum class Effect : uint8_t {
-    None,
     // Instruction doesn't really have effects itself, but it should not be
     // hoisted over any other instruction with effect. Example: Assume
     Order,
@@ -106,8 +105,12 @@ enum class Effect : uint8_t {
     Force,
     // Instruction might use reflection
     Reflection,
-    Any,
+    LeakArg,
+
+    FIRST = Order,
+    LAST = LeakArg,
 };
+typedef EnumSet<Effect> Effects;
 
 // Controlflow of instruction.
 enum class Controlflow : uint8_t {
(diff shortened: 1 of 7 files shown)


rir/src/compiler/pir_tests.cpp
```diff
@@ -52,7 +52,7 @@ closuresByName compileRir2Pir(SEXP env, pir::Module* m) {
     pir::StreamLogger logger({pir::DebugOptions::DebugFlags() |
                                   // pir::DebugFlag::PrintIntoStdout |
                                   // pir::DebugFlag::PrintEarlyRir |
-                                  // pir::DebugFlag::PrintOptimizationPasses |
+                                  pir::DebugFlag::PrintOptimizationPasses |
```
charig: Debug

rir/src/compiler/pir/instruction.h
```diff
@@ -853,51 +875,55 @@ class FLIE(Subassign2_2D, 5, Effect::None, EnvAccess::Leak) {
...
-class FLIE(Extract2_2D, 4, Effect::None, EnvAccess::Leak) {
+class FLIE(Extract2_2D, 4, Effects::Any(), EnvAccess::Leak) {
   public:
     Extract2_2D(Value* vec, Value* idx1, Value* idx2, Value* env,
                 unsigned srcIdx)
         : FixedLenInstructionWithEnvSlot(
               PirType::val().scalar(),
               {{PirType::val(), PirType::val(), PirType::val()}},
               {{vec, idx1, idx2}}, env, srcIdx) {}
+    void updateType() override final { maskEffectsOnNonObjects(); }
```
charig, line 916: Isn't this more like a resetType?

rir/src/compiler/pir/instruction.h
```diff
@@ -125,14 +128,49 @@ class Instruction : public Value {
...
+    }
+    bool mayForcePromises() const { return effects.includes(Effect::Force); }
+    bool leaksArg(Value* val) const {
+        // TODO: for escape analysis we use leaksEnv || hasEffect as a very
+        // crude approximation whether this instruction leaks arguments. We
+        // should do better.
+        return leaksEnv() || effects.includes(Effect::LeakArg);
+    }
+
+    void maskEffectsOnNonObjects() {
```
charig: This is only for binops/unops. The name mask also confused me. I would put sth like resetBinopEffects and add the `(maybeObj) Effects::any()`

charig: Also why the close the if with `};`

rir/src/utils/EnumSet.h
```diff
@@ -28,7 +28,17 @@ class EnumSet {
     }
 
   public:
-    EnumSet() {}
+    typedef Store StoreType;
+    typedef Element ElementType;
```
charig: not used

rir/src/utils/EnumSet.h
```diff
@@ -92,6 +104,8 @@ class EnumSet {
 
     RIR_INLINE Store to_i() const { return set_; }
 
+    constexpr operator StoreType() const { return set_; }
```
charig: what is this?

charig (approved):

rir/src/compiler/pir/instruction.h
o-: why reset? in general this is the api that we call after some optimization passes. it means: given your new input state, can you give a more precise output type. this is not reseting the output type. it is updating it. So I don't understand in what sense that would be resetting (which means setting something to the value it initially was).

rir/src/utils/EnumSet.h
o-: implicit conversion. This function gets called if you pass an EnumSet to a function expecting a StoreType.

I needed this, because I wanted to keep the initial effect part of the template. But a complex type like EnumType cannot be a template parameter. Only a primitive type can.

So when you write:

```Instruction<Effects::Any()>```

Then (at compile time) the `Any` function gets evaluated, this returns an EnumSet, then using the implicit conversion it gets converted to the underlying integer of the bitset. Then finally in the constructor of the instruction, the implicit `Enum(StorageType)` constructor is used to convert it back to an enum set :) yay for zero cost abstractions.

rir/src/compiler/pir/instruction.h
o-: mask as in masking a bitset

o-: (which the function was not doing :) so I fixed that part)

rir/src/compiler/pir/instruction.h
charig: I meant the `maskEffectsOnNonObjects()` method. 
The updateType on arithmetic ops is only ensuring that it is still safe to "hide", and otherwise it "reset" those flags to the original state. 
The name `maskEffectsOnNonObjects` sounded a little low-level to me, we are reseting those flags.

PR #366, opened 2019-03-13
promargs must not have missing args

o-:
the promargs list on the context must not have missing arguments.
the reason (at least one of them) is that nargs() blindly counts
the length of the list, thus would add missing arguments to the
tally.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -1495,11 +1495,13 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP arglist = R_NilValue;
             auto names = (Immediate*)pc;
             advanceImmediateN(n);
+            bool hasMissing = false;
             for (long i = n - 1; i >= 0; --i) {
                 SEXP val = ostack_pop(ctx);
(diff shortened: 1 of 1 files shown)


o-:
tbr @JanJecmen

rir/src/interpreter/interp.cpp
```diff
@@ -1495,20 +1495,44 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
...
                     cptr->cloenv = res;
-                    if (cptr->promargs == symbol::delayedArglist)
-                        cptr->promargs = arglist;
+                    if (cptr->promargs == symbol::delayedArglist) {
+                        auto promargs = arglist;
+                        if (hasMissing) {
+                            // For the promargs we need to strip missing
+                            // arguments from the list, otherwise nargs()
+                            // reports the wrong value.
+                            promargs = Rf_shallow_duplicate(arglist);
```
JanJecmen, line 1518: This allocates so res should be protected? (if the gc doesn't know about cptr)

o-: contexts are from R and they are gc roots

PR #367, opened 2019-03-13
more prcecise dead store optimization using environment escape analysis

o-:
also integrate env access flags into the effect system


o-:
tbr

Issue #368, opened 2019-03-13
PIR_DEBUG_STYLE=GraphViz has no effect in PIR_DEBUG=PrintOptimizationPasses


o-:
@Jakobeha

Jakobeha:
Should be fixed, see my new PR

PR #369, opened 2019-03-13
Fix for #368

Jakobeha:
This makes all calls to `ClosureVersion::print` use the currently set debug style. It doesn't affect printing CSSA or AfterAllocater because those use `printCode`


o-:
thanks

PR #370, opened 2019-03-15
don't rsync recommended packages

o-:
it seems that the currently there is some build breakage in the
recommended packages. let's disable syncing for now, so we use
our known good version of the recommended packages from the cache.


o-:
@guido let's see if this is a workaround for our build problems

PR #375, opened 2019-03-18
[wip] Fix phi placement

o-:


rir/src/compiler/analysis/verifier.cpp
```diff
@@ -214,6 +225,21 @@ class TheVerifier {
                 }
             });
             if (slow) {
+                std::unordered_set<BB*> inp;
+                for (auto in : cfg(bb->owner).immediatePredecessors(bb))
+                    inp.insert(in);
+                phi->eachArg([&](BB* bb, Value*) { inp.erase(inp.find(bb)); });
```
JanJecmen: this can backfire if bb is not found:
> The iterator pos must be valid and dereferenceable. Thus the end() iterator (which is valid, but is not dereferenceable) cannot be used as a value for pos.

rir/src/compiler/pir_tests.cpp
```diff
@@ -394,7 +394,106 @@ bool testPir2Rir(const std::string& name, const std::string& fun,
...
 }
 
+class MockBB : public BB {
+    class MockCode : public pir::Code {
+      public:
+        MockCode(BB* e, size_t s) : Code() {
+            entry = e;
+            nextBBId = s;
+        }
+        ~MockCode() { entry = new MockBB; }
```
mhyee: ???

rir/src/compiler/pir_tests.cpp
```diff
@@ -394,7 +394,106 @@ bool testPir2Rir(const std::string& name, const std::string& fun,
...
+            code.entry = this;
+    }
+    static void reset() {
+        code.entry = nullptr;
+        code.nextBBId = 0;
+    }
+};
+MockBB::MockCode MockBB::code = MockCode(nullptr, 0);
+
+bool testCfg() {
```
mhyee, line 423: A bit of ASCII art here would make it easier to follow the examples.

mhyee:
Also, I gave up trying to understand what `scope_resolution` and `phi_placement` are doing.

o-:
@mhyee It's not that complicated... scope analysis tracks all `stvars`. this allows you to statically approximate ldvars. the scope analysis will tell you all possible values for a particular load. from that set of inputs, you need to then construct ssa variables, if possible.

rir/src/compiler/pir_tests.cpp
o-: :) don't ask. `~Code` insists on deleting something

mhyee: Now that I looked at the [code](https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/pir/code.cpp#L28) it makes sense. Please add a comment, because by itself, this line is very 😖 .

rir/src/utils/Set.h
```diff
@@ -0,0 +1,65 @@
+#ifndef RIR_SET_H
```
charig, line 1: Why???

o-: * std::set :  needs "operator<" implemented and is generally big overhead
* std::unordered_set :  needs hash implemented, easy to get wrong, big overhead for small sets
* SmallSet, just needs `operator==` and is faster and less memory overhead for small sets

o-: just wanted something simpler...

PR #376, opened 2019-03-19
Cleanup

JanJecmen:
Opinions?
* fuse ldvar and force instructions if they are adjacent and the force is the single use of the load
* try to lower pir to rir such that blocks executed one after another appear in the bc stream next to each other
* make visibility explicit even for extracts and loads
* remove unused instruction

### DIFF
--- a/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
@@ -875,16 +875,6 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 EMPTY(PirCopy);
 #undef EMPTY
 
-            case Tag::IsObject: {
-                cs << BC::isobj();
-                break;
-            }
-
-            case Tag::IsEnvStub: {
-                cs << BC::isstubenv();
-                break;
-            }
-
 #define SIMPLE(Name, Factory)                                                  \
     case Tag::Name: {                                                          \
         cs << BC::Factory();                                                   \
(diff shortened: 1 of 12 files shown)


rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -782,7 +789,9 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
 
             case Tag::LdVar: {
                 auto ldvar = LdVar::Cast(instr);
-                cs << BC::ldvarNoForce(ldvar->varName);
+                cs << (ldvar->fusedWithForce
```
charig, line 796: If when coming from pir forces are always explicit, do we need to lower to `ldVar` at all?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -875,16 +885,6 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
                 EMPTY(PirCopy);
 #undef EMPTY
 
-            case Tag::IsObject: {
```
charig, line 882: This was my fault. I left them because @o- did not like the fusing of the is instructions into a typetest and we were planning to rollback

o-:
lgtm

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
JanJecmen: I guess there can be some reordering of instructions that can move the load and force apart

PR #378, opened 2019-03-19
Omit deopt branch

o-:


o-:
just an additional printing flag, tbr

mhyee:
Please update https://github.com/reactorlabs/rir/blob/master/documentation/debugging.md

One comment: in gdb, when I want to print a function, I have to go `p c->printStandard(std::cerr, true)`. I can't omit the flag. So now I guess I have to write `p c->printStandard(std::cerr, true, true)`?

o-:
> One comment: in gdb, when I want to print a function

yeah, don't have a better solution. because default args (or proxy functions) are usually optimized away and not available in gdb either. that's why i didn't bother....

o-:
you can do `1,1` instead of `true,true` ;)

PR #380, opened 2019-03-20
fix force dominance analysis in loops and with deopt exits

o-:
1. creating a promise in a loop should reset the state every time a new
promise is created.
2. deopt exits with unforced promises should not make analysis state
ambiguous, because either we exit with the unforced promise, or we force
and then exit later with the forced version (both things cannot happen by the construction of the graph).

Also there were small issues:
* CallSafeBuiltin did block context removal.
* Cleanup elided envs on framestates (which leads to deopt with elided env)
* Promise inliner didn't correctly update the new promise it creates (somtimes updating it before it's defined) and also deleted the original promise even if it was still in use.


o-:
bugfix

PR #381, opened 2019-03-20
Default type should include the lazy option

charig:
Safe arithmetic ops does not use reflection.

### DIFF
--- a/rir/src/compiler/opt/elide_env_spec.cpp
+++ b/rir/src/compiler/opt/elide_env_spec.cpp
@@ -55,6 +55,8 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
                     });
                     next = ip + 1;
                     i->type.setNotObject();
+                    i->type.setNonLazy();
+                    i->effects.reset(Effect::Reflection);
                 }
 
                 // Speculatively elide envs on forces that only require them in
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -238,7 +238,7 @@ class Instruction : public Value {
     bool usesDoNotInclude(BB*, std::unordered_set<Tag>);
     bool unused();
 
-    virtual void updateType() {};
+    virtual void updateType(){};
 
     virtual void printEnv(std::ostream& out, bool tty) const;
     virtual void printArgs(std::ostream& out, bool tty) const;
@@ -580,11 +580,13 @@ class FLIE(LdFun, 2, Effects::Any()) {
     SEXP hint = nullptr;
 
     LdFun(const char* name, Value* env)
-        : FixedLenInstructionWithEnvSlot(RType::closure, {{PirType::any()}},
+        : FixedLenInstructionWithEnvSlot(PirType(RType::closure).nonLazy(),
+                                         {{PirType::any()}},
                                          {{Tombstone::closure()}}, env),
           varName(Rf_install(name)) {}
     LdFun(SEXP name, Value* env)
-        : FixedLenInstructionWithEnvSlot(RType::closure, {{PirType::any()}},
+        : FixedLenInstructionWithEnvSlot(PirType(RType::closure).nonLazy(),
+                                         {{PirType::any()}},
                                          {{Tombstone::closure()}}, env),
           varName(name) {
         assert(TYPEOF(name) == SYMSXP);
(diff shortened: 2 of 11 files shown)


rir/src/compiler/transform/insert_cast.cpp
```diff
@@ -9,10 +9,11 @@ pir::Instruction* InsertCast::cast(pir::Value* v, PirType t, Value* env) {
     if (v->type.maybePromiseWrapped() && !t.maybePromiseWrapped()) {
         return new pir::Force(v, env);
     }
-    if (v->type == RType::logical && t == NativeType::test) {
+    if (v->type == PirType(RType::logical).nonLazy() && t == NativeType::test) {
```
o-, line 12: this looks strange. Why would `PirType(RType::logical)` create a lazy value by default? you should probably fix the constructor instead

rir/src/compiler/pir/type.h
```diff
@@ -224,6 +227,8 @@ struct PirType {
         assert(isRType());
         PirType t = *this;
         t.flags_.set(TypeFlags::isScalar);
+        t.flags_.reset(TypeFlags::lazy);
```
o-, line 230: no, that's wrong. scalar does not imply eager! you could have a promise that is guaranteed to return a scalar.

rir/src/compiler/pir/type.h
```diff
@@ -174,7 +176,8 @@ struct PirType {
     static PirType val() {
         return PirType(vecs() | list() | RType::sym | RType::chr | RType::raw |
                        RType::closure | RType::prom | RType::code | RType::env |
-                       RType::missing | RType::unbound | RType::ast);
+                       RType::missing | RType::unbound | RType::ast)
+            .nonLazy();
```
o-, line 180: same comment as bellow: this `nonLazy` should not be neccessary. if this combination returns a lazy value, then the constructor needs to be fixed to not return a lazy value.

rir/src/compiler/pir/type.cpp
```diff
@@ -83,8 +83,16 @@ PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
     }
 
     if (PirType::vecs().isSuper(*this)) {
-        if (Rf_length(e) == 1)
+        if (Rf_length(e) == 1) {
             flags_.set(TypeFlags::isScalar);
+            flags_.reset(TypeFlags::lazy);
```
o-, line 88: that should not be here i guess

rir/src/compiler/pir/singleton_values.h
```diff
@@ -97,7 +99,7 @@ class Tombstone : public Value {
             assert(false);
     }
     static Tombstone* closure() {
-        static Tombstone cls(RType::closure);
+        static Tombstone cls(PirType(RType::closure).nonLazy());
```
o-, line 102: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -1188,7 +1193,7 @@ class VLIE(Call, Effects::Any()), public CallInstruction {
                                        srcIdx) {
         assert(fs);
         pushArg(fs, NativeType::frameState);
-        pushArg(fun, RType::closure);
+        pushArg(fun, PirType(RType::closure).nonLazy());
```
o-, line 1198: dito (and all above...)

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -55,6 +55,8 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
                     });
                     next = ip + 1;
                     i->type.setNotObject();
+                    i->type.setNonLazy();
```
o-, line 58: this is done by `updateType`. replace all those with `i->updateType()`

rir/src/compiler/pir/type.h
```diff
@@ -242,7 +247,7 @@ struct PirType {
         return t;
     }
 
-    PirType forced() const {
+    PirType nonLazy() const {
```
o-, line 250: that's a terrible name! your `nonLazy` function removes the `lazy` flag and the `promiseWrapped` flag. In all the rest of the pir we call a `nonLazy` or `eager`, a value that is evaluated but maybe still promise wrapped.  how about `asValue`?

PR #387, opened 2019-03-20
Simple range

Jakobeha:
(testing GitLab runner)


rir/src/ir/insns.h
```diff
@@ -286,6 +291,13 @@ DEF_INSTR(aslogical_, 0, 1, 1, 0)
  */
 DEF_INSTR(asbool_, 0, 1, 1, 0)
 
+/**
+ * asint_ :: pop object stack, convert to integer scalar and push. Ceils/floors
+ *           if real, 0 or 1 if logical, throws an NA error if another type. For
```
o-: I would recommend creating two instructions. floor and ceil. that seems simpler and less confusing.

rir/src/interpreter/interp.cpp
```diff
@@ -2510,6 +2525,81 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
...
+                    }
+                }
+            } else { // Everything else
+                x = NA_INTEGER;
+                isNaOrNan = true;
+            }
+            if (isNaOrNan) {
+                Rf_errorcall(getSrcAt(c, pc - 1, ctx), "NA/NaN argument");
+            }
+            res = Rf_allocVector(INTSXP, 1);
```
o-: here you could check, if `NO_REFERENCES(val)` and also val is a simple scalar, you can store the result into val instead of creating a new vector.

rir/src/compiler/opt/delay_instr.cpp
```diff
@@ -17,7 +17,12 @@ void DelayInstr::apply(RirCompiler&, ClosureVersion* function,
             auto next = ip + 1;
 
             if (!i->hasEnv() && !i->hasEffect() && !Phi::Cast(i) &&
-                !i->branchOrExit()) {
+                !i->branchOrExit() &&
+                // TODO: Find out why moving these instructions breaks simple
+                // ranges (seems to mess up the SSA allocator).
+                // For now, these are only used by compiled specials, and never
+                // get optimized by being delayed anyways
+                !Inc::Cast(i) && !Dec::Cast(i) && !SetShared::Cast(i)) {
```
o-: is this still the case when you rebase?

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,143 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+            //   }
+            // }
+
+            CodeStream& cs = ctx.cs();
+            BC::Label fwdBranch = cs.mkLabel();
+            BC::Label endBranch = cs.mkLabel();
+
+            // i' <- m
+            compileExpr(ctx, start);
+            cs << BC::setShared() << BC::asint(false);
```
o-: ensureNamed should be enough

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,143 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+            // n' <- n
+            compileExpr(ctx, end);
+            cs << BC::setShared();
```
o-: dito

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,143 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+                return false;
+            }
+
+            SEXP start = args[0];
+            SEXP end = args[1];
+            if (TYPEOF(start) != INTSXP && TYPEOF(start) != LGLSXP &&
+                (TYPEOF(start) != REALSXP || XLENGTH(start) == 0 ||
+                 *REAL(start) != (int)*REAL(start))) {
+                return false;
+            }
```
o-, line 194: so start has to be a constant, whereas end can be evaluated, right? what's the reason for this choice?

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,143 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+
+            // for(i in m:n) {
+            //   ...
+            // }
+            // =>
+            // i' <- m
+            // n' <- n
+            // if (i' > n') {
```
o-, line 202: if i and n are constants you could do this statically (ie. only compile the relevant part)

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,143 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+            }
+
+            SEXP start = args[0];
+            SEXP end = args[1];
+            if (TYPEOF(start) != INTSXP && TYPEOF(start) != LGLSXP &&
+                (TYPEOF(start) != REALSXP || XLENGTH(start) == 0 ||
+                 *REAL(start) != (int)*REAL(start))) {
+                return false;
+            }
+
```
o-, line 195: love the comments in this function. thanks

rir/src/compiler/opt/delay_instr.cpp
Jakobeha: This was since Saturday, unless any changes were made then

rir/src/ir/Compiler.cpp
Jakobeha: If `start` is an (actual, not in R types) integer than the list will guaranteed either only contain integers (even if `end` is a real), or raise an error (if `end` isn't a number). If `start` isn't an integer than the list could contain non-integers, and then we'd have to `inc` and `dec` reals (which would be possible but more complicated). The best way to check if `start` is an integer is by looking if its an integer literal. Ultimately, most programs will have `for (i in <constant>:n)`, so its easier to only handle those cases.

rir/src/ir/Compiler.cpp
Jakobeha: yes, but I don't think it will really affect performance, maybe I can try after this is merged

rir/src/ir/Compiler.cpp
o-: ok, I got it. actual integer literals are not very common, since you have to explicitly write the trailing `L`. but maybe you could also do this to real literals, which happen to be an integer?

Jakobeha: Sorry I didn't use the right term. This checks if `start` is a literal which is a "true" integer, so if `start` is an R-type real literal it can still compile the simple range.

rir/src/compiler/opt/delay_instr.cpp
Jakobeha: Nevermind, looks like this section is completely gone in the latest `master`

rir/src/ir/Compiler.cpp
o-: right, now I actually read the thing.

Jakobeha:
Submodule is different for this and other PRs, but my submodule is at `3.5.1-rir-patch` and there aren't any actual changes

rir/src/interpreter/interp.cpp
```diff
@@ -2834,7 +2941,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP vec = ostack_at(ctx, 1);
             SEXP val = ostack_at(ctx, 2);
 
-            if (MAYBE_SHARED(vec)) {
+            if (MAYBE_REFERENCED(vec)) {
```
o-: this should stay maybe_shared

rir/src/interpreter/interp.cpp
```diff
@@ -2872,7 +2979,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP mtx = ostack_at(ctx, 2);
             SEXP val = ostack_at(ctx, 3);
 
-            if (MAYBE_SHARED(mtx)) {
+            if (MAYBE_REFERENCED(mtx)) {
```
o-: this should stay maybe_shared

rir/src/interpreter/interp.cpp
```diff
@@ -2912,7 +3019,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP val = ostack_at(ctx, 2);
 
             // Fast case
-            if (NOT_SHARED(vec) && !isObject(vec)) {
+            if (NO_REFERENCES(vec) && !isObject(vec)) {
```
o-: this should stay maybe_shared

rir/src/interpreter/interp.cpp
```diff
@@ -2963,7 +3070,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
                 }
             }
 
-            if (MAYBE_SHARED(vec)) {
+            if (MAYBE_REFERENCED(vec)) {
```
o-: this should stay maybe_shared

rir/src/interpreter/interp.cpp
```diff
@@ -3005,7 +3112,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP val = ostack_at(ctx, 3);
 
             // Fast case
-            if (NOT_SHARED(mtx) && !isObject(mtx)) {
+            if (NO_REFERENCES(mtx) && !isObject(mtx)) {
```
o-: this should stay maybe_shared

rir/src/interpreter/interp.cpp
```diff
@@ -3071,7 +3178,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
                 }
             }
 
-            if (MAYBE_SHARED(mtx)) {
+            if (MAYBE_REFERENCED(mtx)) {
```
o-: this should stay maybe_shared

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,141 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+            //   }
+            // }
+
+            CodeStream& cs = ctx.cs();
+            BC::Label fwdBranch = cs.mkLabel();
+            BC::Label endBranch = cs.mkLabel();
+
+            // i' <- m
+            compileExpr(ctx, start);
+            cs << BC::floor();
```
o-: missing ensureNamed

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,141 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+            compileExpr(ctx, end);
+            cs << BC::ensureNamed();
+            // if (i' > n')
+            cs << BC::dup2() << BC::gt();
+            cs.addSrc(R_NilValue);
+            cs << BC::brfalse(fwdBranch);
+            // {
+            // n' <- ceil(n') - 1
+            cs << BC::ceil() << BC::dec();
+            // while
```
o-, line 235: missing ensureNamed

rir/src/ir/Compiler.cpp
```diff
@@ -145,6 +145,141 @@ Code* compilePromise(CompilerContext& ctx, SEXP exp);
...
+                             cs << BC::swap() << BC::dec() << BC::swap();
+                             // ...
+                             compileExpr(ctx, body);
+                             // }
+                         });
+            // } else {
+            cs << BC::br(endBranch) << fwdBranch;
+            // n' <- floor(n') + 1
+            cs << BC::floor() << BC::inc();
+            // while
```
o-, line 256: missing ensure named

o-: inc should not be used here, since you don't modify in-place. use just `push 1; add`

rir/src/ir/Compiler.cpp
o-: dec should not be used here, since you don't modify in-place. use just `push 1; sub`

o- (commented):
So, I remembered, that `inc_` is supposed to update in-place, even if the refcount is 1.

That was a feature, not a bug, to be able to increment the loop counter without creating a new vector all the time.

So the idea is we compile the for loop as follows:
```
%i = 0
while (%i < size(seq))
  ...
  %i++
```

now, we need to mark the (stack) value `%i` ensureNamed, because otherwise the `<` comparison might override it (values with refcount 0 can be reused). But then the i++ at the end of the loop, should still be able to increment i without duplicating it. so the inc bytecode is allowed to destructively modify it

rir/src/ir/Compiler.cpp
Jakobeha: Not sure, if `start` is a constant can't it be modified in place? If it's a variable it should already be named and then `floor` will either copy or not modify

rir/src/ir/Compiler.cpp
Jakobeha: Not sure if ensureNamed is needed, I understand `push 1; sub` though

rir/src/ir/Compiler.cpp
Jakobeha: Similar to above

rir/src/ir/Compiler.cpp
o-: yes, and the copy will have refcount of 0. an you use this value multiple times

rir/src/ir/Compiler.cpp
o-: the resulting value `n'` is used in the loop (ie. multiple times). ensureNamed is needed!

rir/src/ir/Compiler.cpp
o-: please be a bit careful with the refcounts in general. those are not fun bugs to debug and it's easy to get an error to accidentally pass tests.

Jakobeha:
I avoided the `ensureNamed` problems you mentioned above by putting `setShared` before each `stvar`, so it would always copy in the following `inc`. I tried using the `ensureNamed`s instead, but they failed the tests. This passes the tests, it copies every time `i` is incremented in the loop, but this might be necessary because modifying `i` in place would modify the variable it got set to

o-:
> I tried using the ensureNameds instead, but they failed the tests. This passes the tests, it copies every time i is incremented in the loop,

I don't want to merge on the grounds that "it passes tests". we need to understand what is going on.
 
> but this might be necessary because modifying i in place would modify the variable it got set to

I am not sure that this is the case. maybe you can get a hold on this if you form a hypothesis what the refcount should be at every step of your loop and then compare what it actually is.

for example, I think in `stvar("i", i')` the refcount of `i'` should be at least 1 (due to the ensureNamed) and thus `stvar` should further increase the refcount to 2. This should prevent the `inc` from overriding.

o-:
this branch currently modifies the gnur submodule. I believe this is an accident.

Jakobeha:
The submodule is weird, the current submodule on `3.5.1-rir-patch`, `master`, and this branch is `13994bc`. `bd9fc9` doesn't even exist.

mhyee:
> The submodule is weird, the current submodule on `3.5.1-rir-patch`, `master`, and this branch is `13994bc`. `bd9fc9` doesn't even exist.

Maybe there was something in your local `gnur` repo? You could try rebasing your local branch on master and then force pushing.

Jakobeha:
I replaced `setShared` with `ensureNamed`, as you mentioned above. However some of the PIR compiled code fails, probably because PIR removes the `ensureNamed` but it wouldn't remove `setShared`. So I merged with #410 to see if that fixes

o-:
this is a bit of a mess right now. I would like to merge, but I can't since this includes half of the yet to be reviewed #416 plus various commits which are neither in master nor in any pending PR, but from some intermediate versions of my PR. I don't know how, but this needs to be cleaned up...

Jakobeha:
I thought this needed #416 in order for this to work (I now know that I don't), so we can merge after that gets merged. I'm not sure about the other commits, I merged with some previous PRs (like `fixIncReally`) which eventually got merged into master.

Next time, what should I do when I have a PR which might depend on another PR?

o-:
> This needs #416 in order to work, so we can merge after that gets merged.

#418, which I now merged might be enough?

> I'm not sure about the other commits, I merged with some previous PRs which I needed (like fixIncReally) which eventually got merged into master.

Ok, I hope that this pr really does not diverge from #416.

> Next time, what should I do when I have a PR which depends on another PR?

There is no generic solution and it depends a bit on personal taste. But merging in another feature branch greates a huge mess, when that other feature branch gets force-pushed.

so you did something like this:

```
-------- master
  \   \   
   \    --c-- b1
    \            \
      ------------- b2
```

where b1 is my branch and c is a commit in my branch. you merged into your b2.

now if I change c and force push, this is akin to replacing b1, with a completely new branch b1', so when you merge again you get this:

```
---------------- master
  \   \    \
   \   \    -- c' -- b1'
    \    --c-- b1      \
     \            \     \
       -------------------- b2
```

if you do this a couple of times it get's basically impossible to figure out what comes from where. Once b1' is merged (and the old b1 abandoned), you will have this:

```
-------------- c' -- master
  \   \
   \    --c-----
    \            \
     -------------------- b2
```

where you will accidentally drag in the diff between c and c'. You will pretty much have to manually go through the textual diff and sort out everything that you captured by accident.

o-:
what people usually do is, to make their own branch based on the other branch instead of master.
basically if you have a branch based on master, but you need a fix from f, you say: `git rebase f`

This changes the graph from:

```
-------- master
  \   \
   \    --c-- f
    \
      ------------- b2

```

to:

```
-------- master
  \
    --c-- f
           \   
            --- b2

```

after an update to f:

```
-------- master
 \  \
  \   ---c'---f'
   \
     --c-- f
             \ 
              --- b2
```

you say again: `git rebase --onto f' f b2`, and git will do this:

```
-------- master
 \  \
  \   ---c'---f'
   \           \
     --c-- f    \
                 \
                  --- b2
```

o-:
you have to get used to it and adapt the workflow a bit. For example, you really need to avoid interference between your branch and the f branch, otherwise rebase will have a hard time to do this cleanly.

usually what helps is to squeze everything into one commit, before rebaseing.

for example what I usually do (assuming I am in a branch which is based on upstream/feature)
```
git rebase upstream/feature --interactive
```

this allows you to edit, combine and reorder the commits in your branch. Once my branch is clean (usually just 1 or 2 combined commits) I will fetch the new feature branch and update:

```
git fetch upstream
git rebase --onto upstream/feature <oldBranchBaseId>
```

rebase will then apply all your commits, one-by-one on top of the updated branch. If anything fails, then you can edit, `git add mergedFile` and then `git rebase --continue`.

o-:
A great way of getting started with this workflow is to use `git rebase master` when you want to update your topic branch with the newest commits from master. This command is basically just doing:


from

```
------master
  \
    ---- b
```
to

```
------master
        \   
         ---- b
```

for your b branch, instead of this (what merge does):

```
------master
  \     \   
    ---- b
```

o-:
if you ever get lost `git reflog` is your friend. It shows you the last versions you had your HEAD pointing to. A comited version can never get lost. you can always get it back using a combination of checkout and reset.

rir/src/ir/Compiler.cpp
mhyee: I'm bumping this. I don't know how much it affects performance (execution time is probably unaffected, compilation time probably increases because now the loop is twice as large), but trying to read the PIR output is a bit annoying, when there are effectively two versions of a for loop.

I was scratching my head and wondering why my one line program had two loops in it...

Jakobeha:
@myhee I can make it compile one loop if you want. The reason I did 2 loops
is because with only 1, I would need to put a condition inside the loop
which checks whether to increment or decrement. I needed to handle the
decrement case because in `for (i in 1:n)`, I can't determine if `n` is
negative, and if so the index will go down.

Alternatively, for constants I can determine if the index will go up or
down, so I would only need to compile one loop without a condition. But PIR
would take care of that if we get better constantfolding and dead branch
removal.

PR #388, opened 2019-03-20
Debug printing fixes

mhyee:
I had these in my working branch, but figured I might as well split them off in case other people need them.


o-:
(canceled the gitlab ci manually, since it is clogged already...)

PR #391, opened 2019-03-21
Remove asserts from release build

charig:


o-:
this also removes the asserts from ci runs :(

charig:
The `debug` and `debugopt` builds are for testing asserts I suppose...

o-:
> The debug and debugopt builds are for testing asserts I suppose...

they are too slow, we run half of ci in release. ci jobs are defined in .gitlab-ci.yml and .travis.yml

o-:
so please make sure that asserts are not turned off on the ci

rir/src/compiler/translations/pir_2_rir/stack_use.cpp
```diff
@@ -46,12 +49,16 @@ void StackUseAnalysisState::AbstractStack::matchContents(
         if (operator[](i) != other.operator[](i)) {
             if (operator[](i)->isInstruction() &&
                 other.operator[](i)->isInstruction()) {
-                auto my = Instruction::Cast(operator[](i))->hasSingleUse();
-                auto their =
-                    Instruction::Cast(other.operator[](i))->hasSingleUse();
-                assert(my && Phi::Cast(my));
-                assert(their && Phi::Cast(their));
-                assert(my == their && "Stack mismatch.");
+                assert(Instruction::Cast(operator[](i))->hasSingleUse() &&
```
o-: ehm, how about using a temporary variable to store `operator[](i)`? (which is already the imho the ugly version of `(*this)[i]`, but that's not from your pr, ik)

rir/src/interpreter/interp.cpp
```diff
@@ -296,6 +298,8 @@ static RIR_INLINE SEXP createLegacyArgsList(const CallContext& call,
     }
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-variable"
```
o-: You can also just completely remove those asserts. Those are really from the beginning of time, when I wasn't exactly sure how setjmp/longjmp work.

o-: If they would fail, the C compiler would have been very drunk...

rir/src/ir/CodeVerifier.cpp
```diff
@@ -234,6 +234,8 @@ void CodeVerifier::calculateAndVerifyStack(Code* code) {
     code->stackLength = max.ostack;
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-variable"
 void CodeVerifier::verifyFunctionLayout(SEXP sexp, InterpreterInstance* ctx) {
     assert(TYPEOF(sexp) == EXTERNALSXP and "Invalid SEXPTYPE");
```
o-: if the code verifier fails, the code is seriously broken and should not be executed. I think it would be a good idea to convert those asserts into Rf_error. The same goes for the pir verifier.

o- (approved):
verifier error should always lead to abort. otherwise lgtm

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -163,7 +162,8 @@ class TheVerifier {
 
         if (auto mk = MkArg::Cast(i)) {
             auto p = mk->prom();
-            assert(p->owner->promise(p->id) == p);
+            if (p->owner->promise(p->id) != p)
```
o-: sorry, I don't want to drag this out unnecessarily. But if you actually look at what the verifier does, whenever there is a verification error, it sets ok to false. Then the callees currently do `assert(Verfier::verify....`.

So although you now changed a bunch of assumes into errors, you do not check the final verifier result with those changes.

can you please (instead of erroring on some conditions inside the verification functions), just error on the final `ok` result??

PR #393, opened 2019-03-21
fixed missing implicit args in pir compiled function

Jakobeha:
Also added a regression test:

```r
f <- pir.compile(rir.compile(function(a,b,c) a))
g <- rir.compile(function() {
  f()
  f(1)
  f(1,2)
  f(1,2,3)
})

rir.disassemble(f)
stopifnot(g()==1)
```

Calling `g` fails an assertion in call_context.cpp: `i < passedArgs`. `ldarg` assumes that we pass the loaded argument. However, when calling a function implicitly this isn't always true. This PR makes it so, when `i >= passedArgs`, a missing implicit argument is returned instead.

Perhaps we should add this guard `ldarg` instead

### DIFF
--- a/rir/src/interpreter/call_context.h
+++ b/rir/src/interpreter/call_context.h
@@ -72,8 +72,11 @@ struct CallContext {
     bool hasNames() const { return names; }
 
     Immediate implicitArgIdx(unsigned i) const {
-        assert(implicitArgs && i < passedArgs);
-        return implicitArgs[i];
+        assert(implicitArgs);
+        if (i < passedArgs)
+            return implicitArgs[i];
+        else
+            return MISSING_ARG_IDX;
     }
 
     bool missingArg(unsigned i) const {
--- a/rir/tests/pir_regression_missing.R
+++ b/rir/tests/pir_regression_missing.R
@@ -77,6 +77,7 @@ g <- rir.compile(function() {
   f(1,2,3)
 })
 
-rir.disassemble(f)
+stopifnot(g()==1)
+pir.compile(g)
 stopifnot(g()==1)
 
--- a/rir/tests/pir_regression_missing.R
+++ b/rir/tests/pir_regression_missing.R
@@ -68,3 +68,15 @@ stopifnot(h()==3)
 stopifnot(h()==3)
 stopifnot(h()==3)
 stopifnot(h()==3)
+
+f <- pir.compile(rir.compile(function(a,b,c) a))
+g <- rir.compile(function() {
+  f()
+  f(1)
+  f(1,2)
+  f(1,2,3)
+})
+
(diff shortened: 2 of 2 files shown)


Jakobeha:
This is failing a test in GitLab, `test-debug`, but I have no idea why. The only change affects an assertion failure.

o-:
@Jakobeha I think your fix is masking a bug.  because the rir dispatch should ensure that we are not calling a pir version which expects more arguments than we pass. because doing so will produce incorrect results with default arguments. can you try to see why we are dispatching to this version?

Jakobeha:
We have a guard for this, but it only runs if the closure takes default arguments: https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp#L87.

```r
    if (closure->formals().hasDefaultArgs()) {
        ...
        if (!ctx.assumptions.includes(Assumption::NotTooFewArguments)) {
            logger.warn("TODO: don't know how many are missing");
            return fail_();
        }
    }
```

o-:
@Jakobeha anyway, you can't merge this. As I said this is a dispatching bug and not a ldarg bug. Check `dispatch` or `matches` implementations in the interpreter. somebody is calling a version that should not be called.

Jakobeha:
Here is a different fix - it forces PIR to recompile the function if the # of missing arguments is different, with the correct # of missing arguments.

However, as the test shows, GNU-R throws an error if a function tries to reference a missing variable, while PIR just returns the missing value. I know the semantics for this are very weird - e.g. GNU-R *won't* throw an error if the function passes the variable to a function at least in some cases. This is already an issue in PIR. I believe the solution is, whenever a value is forced, RIR must also check if the value is missing and throw an error if so - PIR can still remove forces, but it can't remove this check (unless it knows the value won't be missing), so we would need an extra opcode/instruction like `ForceMissing`.

o-:
hey, I am sorry. I think I your initial patch was correct after all and my review was wrong.

In particular

> because the rir dispatch should ensure that we are not calling a pir version which expects more 
> arguments than we pass. because doing so will produce incorrect results with default arguments. 

is not correct. If pir does assume that no arguments are missing, it does set the `NotTooFewArguments` assumption. if the function has no default argumetns, it does not need this assumption. In your regression test pir did not use that assumption, thus it is correct to call the version with too few arguments.

the difference is, for stack args we have https://github.com/reactorlabs/rir/blob/master/rir/src/interpreter/interp.cpp#L746 to supply the missing args

but for implicit args we don't have this mechanism and thus loading a missing arg can happen, and just returning R_MissingArg in that case is the correct thing to do.

So, please merge  bc95bc3.

PR #394, opened 2019-03-21
 delay environments into deopt branches and GVN 

o-:


rir/src/compiler/opt/delay_instr.cpp
```diff
@@ -16,35 +16,6 @@ void DelayInstr::apply(RirCompiler&, ClosureVersion* function,
             auto i = *ip;
             auto next = ip + 1;
 
-            if (!i->hasEnv() && !i->hasEffect() && !Phi::Cast(i) &&
-                !i->branchOrExit()) {
-                Instruction* usage = i->hasSingleUse();
-                if (usage && usage->bb() != bb) {
```
charig, line 22: I guess this removal has something to do with the phi placement changes, but why removing also the else part that just delays non effectful instructions?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -250,29 +251,26 @@ void Instruction::replaceUsesAndSwapWith(
 }
 
 bool Instruction::usesAreOnly(BB* target, std::unordered_set<Tag> tags) {
-    bool answer = true;
-    Visitor::run(target, [&](Instruction* i) {
+    return Visitor::check(target, [&](Instruction* i) -> bool {
```
charig, line 254: Why continue executing after we already know the answer?

rir/src/compiler/pir/value.h
```diff
@@ -43,7 +43,8 @@ class Value {
     virtual SEXP asRValue() const { assert(false && "Not a singleton"); }
 
     bool producesRirResult() const {
-        return type != PirType::voyd() && type != NativeType::context;
+        return type != PirType::voyd() &&
+               (type.isRType() || type == NativeType::test);
```
charig, line 47: Ok, but I wonder, was this failing anywhere?

rir/src/compiler/pir_tests.cpp
```diff
@@ -612,6 +612,75 @@ bool testCfg() {
...
+    }\n\
+    return (sum);\n\
+}\n";
+
+static bool testMandelbrot() {
+    pir::Module m;
+    SEXP env = compileToRir("", mandelbrot, R_GlobalEnv);
+    SEXP fun = CAR(FRAME(env));
+    // warmup
+    rirApplyClosure(R_NilValue, fun, R_NilValue, R_GlobalEnv);
```
charig, line 669: Do we need the warmup?

rir/src/compiler/pir_tests.cpp
```diff
@@ -815,19 +884,23 @@ static Test tests[] = {
...
+    Test("Test mandelbrot is scope-resolved", &testMandelbrot),
 };
 
 } // namespace
 
 namespace rir {
 
 void PirTests::run() {
+    size_t oldconfig = Rir2PirCompiler::MAX_INPUT_SIZE;
+    Rir2PirCompiler::MAX_INPUT_SIZE = 3000;
```
charig, line 896: speedup?

rir/src/utils/EnumSet.h
```diff
@@ -81,6 +81,10 @@ class EnumSet {
         return set_ == s.set_;
     }
 
+    constexpr RIR_INLINE EnumSet operator~() const {
+        return EnumSet(~set_ & Any());
```
charig, line 85: Why `& Any()`?

rir/src/compiler/opt/load_elision.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <unordered_set>
+
+namespace rir {
+namespace pir {
+
+struct ALoad {
+    // cppcheck-suppress noExplicitConstructor
+    ALoad(LdVar* ld) : origin(ld), env(ld->env()), name(ld->varName) {}
+    // cppcheck-suppress noExplicitConstructor
+    ALoad(LdFun* ld) : origin(ld), env(ld->env()), name(ld->varName) {}
```
charig, line 20: They are two absolutely equal constructors?

rir/src/compiler/opt/gvn.cpp
```diff
@@ -0,0 +1,204 @@
...
+                storeNumber(i, nextNumber, {nextNumber});
+            } else {
+                // We allow instructions with those effects to be deduplicated.
+                auto effects =
+                    i->effects & (~(Effects(Effect::Error) | Effect::Warn |
+                                    Effect::Visibility | Effect::Force));
+                if (effects.empty())
+                    computeNumber(i);
+                else
+                    assignNumber(i);
```
charig: Why do we care at all? Why not just adding the effects check at the beggining?

rir/src/compiler/opt/gvn.cpp
```diff
@@ -0,0 +1,204 @@
...
+            }
+        };
+
+        while (changed) {
+            changed = false;
+            DominatorTreeVisitor<>(dom).run(cls->entry, [&](BB* bb) {
+                for (auto i : *bb)
+                    computeGN(i);
```
charig, line 154: If it is a dominator tree, why would be need to iterate?

rir/src/compiler/opt/gvn.cpp
```diff
@@ -0,0 +1,204 @@
...
+                            if (!i->bb()->before(firstInstr, i))
+                                continue;
+                        } else {
+                            if (!dom.dominates(firstInstr->bb(), i->bb()))
+                                continue;
+                        }
+                    }
+                    i->replaceUsesWith(first);
+                    // Make sure this instruction really gets removed
+                    i->effects.reset();
```
charig, line 198: Why this? Because of the error, etc effects?

charig (approved):

rir/src/compiler/opt/delay_instr.cpp
o-: I think there is just no gain in doing that. I think I added this initially as sort of a demo pass, to see what the effect annotations allow us to do.

rir/src/compiler/pir/instruction.cpp
o-: yes that was my thinking. your implementation checks every instruction, regardless. my implementation stops after the first counterexample instruction

rir/src/compiler/pir/value.h
o-: it was just wrong, why shouldn't i change it? it also caused gvn to run on things like frame state instructions

rir/src/compiler/pir_tests.cpp
o-: yes

rir/src/compiler/pir_tests.cpp
o-: no the test fails otherwise

rir/src/utils/EnumSet.h
o-: to clear the bits outside of the valid range. otherwise the fast equality check (using integer machine equality) would fail

rir/src/compiler/opt/load_elision.cpp
o-: spot the difference :)

rir/src/compiler/opt/gvn.cpp
o-: sorry, I don't understand the question. and which beginning?

rir/src/compiler/opt/gvn.cpp
o-: as you can see on a simple if-then-else example, dominator tree order does not guarantee that all inputs are already assigned a number. tbh, I am not even sure if it is faster than depth first.

rir/src/compiler/opt/gvn.cpp
o-: the idea is, that we ignore error/warning for gvn. the reasoning is that the error/warning would already have happened, in the instruction we use to replace with. but then the intruction we replaced, won't be removed, since it is not effectless. so I clear the effects here, such that cleanup pass can remove it

rir/src/compiler/pir_tests.cpp
o-: (we need to speculate on the bitwise builtins)

rir/src/compiler/pir_tests.cpp
o-: (on travis, where we set the limit to 700, to speed up the runs)

rir/src/compiler/opt/gvn.cpp
charig: I meant...it seems that for any instruction with "real effect" we do not need to do more than `assignNumber`. You have this above in the same function:
```
auto i = Instruction::Cast(v);
            if (!i) {
                assignNumber(v);
                return;
            }
```
why not just add `|| i->effects()`

rir/src/compiler/pir_tests.cpp
charig: The default feedback enables speculation

o-: not for functions

rir/src/compiler/opt/gvn.cpp
o-: ok, i can do that, if you prefer

PR #396, opened 2019-03-23
small fix in insert_cast

Jakobeha:
fixes a bug where type isn't updated if the instruction isn't `Phi`

### DIFF
--- a/rir/src/compiler/transform/insert_cast.cpp
+++ b/rir/src/compiler/transform/insert_cast.cpp
@@ -31,10 +31,7 @@ void InsertCast::apply(BB* bb) {
     auto ip = bb->begin();
     while (ip != bb->end()) {
         Instruction* instr = *ip;
-        Phi* p = nullptr;
-        if ((p = Phi::Cast(instr))) {
-            p->updateType();
-        }
+        instr->updateType();
         instr->eachArg([&](InstrArg& arg) {
             while (!arg.type().isSuper(arg.val()->type)) {
                 auto c = cast(arg.val(), arg.type(), env);


rir/src/compiler/transform/insert_cast.cpp
```diff
@@ -31,10 +31,7 @@ void InsertCast::apply(BB* bb) {
     auto ip = bb->begin();
     while (ip != bb->end()) {
         Instruction* instr = *ip;
-        Phi* p = nullptr;
-        if ((p = Phi::Cast(instr))) {
-            p->updateType();
-        }
+        instr->updateType();
```
charig, line 34: This is weird because phi's updateType is the only changing the type of the arguments. What kind of issue did you find?

Jakobeha: I added a `SetShared` instruction with an integer input, followed by an instruction which expected an integer. The `SetShared` normally updates its return type, but it didn't because `updateType` wasn't called, so `rir_2_pir` tried to cast its result type (`val?`) to an integer for the next instruction, and this failed.

charig: The change does not seem to hurt. I'm only afraid it may be hiding other issue. Because `SetShared` inherits the type from its argument at creation time. So, if you previously had an integer and then an operation expecting an int, and you only add a `setShared` in between, I do not see how the failure can happen

Jakobeha: The other type got resolved to an integer later. It was the index in the simple-range - it's a PIR SSA value, so PIR could infer its type, but the cast was being added by the compiler before any phases.

It is a really small thing, I don't think it really matters. Maybe just take note of this in the future.

PR #397, opened 2019-03-23
P2r code cleanup

JanJecmen:
This adds a code buffer for the backend and then performs some peephole optimizations on it mostly to clean up some stack shuffling instructions


o-:
I am ok with merging this. But honestly I would much prefer if our allocator would produce decent code in the first place.

JanJecmen:
Agreed, this was just a quick fix..
Honestly, I am not sure how to go about the allocator producing good code, it's kind of circular: you want to decide stack/register based on the stack shuffling needed, but you only know the stack shuffling once you decide where to allocate stuff
But for starters, maybe an api that would tell you, given some stack contents and an instruction, what is the optimal way to pick/pull/put/pop such that the instruction can be executed
Or do you have something particular in mind?

PR #398, opened 2019-03-24
Keyboard Interrupt Handler

scottcarroll:
Adds functionality to RIR to check for SIGINTs. Does so in a fashion similar to gnu-r. Performance and when to check for termination signals is kept in mind with this addition. An example of functionality added:
> f <- function() repeat NULL
> f()
^C

This now results in rir exiting the loop gracefully.


rir/src/interpreter/interp.cpp
```diff
@@ -45,12 +45,12 @@ static RIR_INLINE SEXP getSrcForCall(Code* c, Opcode* pc,
 #define INSTRUCTION(name)                                                      \
     op_##name: /* debug(c, pc, #name, ostack_length(ctx) - bp, ctx); */
 #define NEXT()                                                                 \
-    (__extension__({ goto* opAddr[static_cast<uint8_t>(advanceOpcode())]; }))
+    (__extension__({                                                           \
```
mhyee: Was this just a whitespace change? We'd prefer to keep changes minimal.

rir/src/interpreter/interp.cpp
```diff
@@ -45,12 +45,12 @@ static RIR_INLINE SEXP getSrcForCall(Code* c, Opcode* pc,
...
+        goto* opAddr[static_cast<uint8_t>(advanceOpcode())]; }))
 #define LASTOP                                                                 \
     {}
 #else
 #define BEGIN_MACHINE                                                          \
-    loop:                                                                      \
```
mhyee: Should this have been removed?

rir/src/interpreter/interp.cpp
```diff
@@ -358,8 +358,17 @@ static RIR_INLINE SEXP rirCallTrampoline(const CallContext& call, Function* fun,
     return rirCallTrampoline(call, fun, symbol::delayedEnv, arglist, ctx);
 }
 
-#pragma GCC diagnostic push
```
mhyee: It looks like these two lines didn't get merged properly. Putting them back (in the right place) will probably fix the build errors on Travis.

rir/src/interpreter/interp.cpp
```diff
@@ -358,8 +358,17 @@ static RIR_INLINE SEXP rirCallTrampoline(const CallContext& call, Function* fun,
...
-#pragma GCC diagnostic ignored "-Wunused-variable"
+int count = 0;
+
+// Interrupt Signal Checker - Allows for Ctrl - C functionality to exit out
+// of infinite loops
+void checkuserinterrupt() {
+    if (++count > 1000) {
```
mhyee: It might be useful to define this as a macro, similar to what [GNU-R does](https://github.com/reactorlabs/gnur/blob/R-3.5.1-rir-patch/src/main/eval.c#L5402). I think we only use it in one place, but it makes it easier to understand the code. Otherwise it just looks like a random "[magic constant](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants)."

rir/src/interpreter/interp.cpp
```diff
@@ -358,8 +358,17 @@ static RIR_INLINE SEXP rirCallTrampoline(const CallContext& call, Function* fun,
     return rirCallTrampoline(call, fun, symbol::delayedEnv, arglist, ctx);
 }
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-variable"
+int count = 0;
```
mhyee: Two things here:

1. This should be `unsigned`, because we know negative values can never occur. More importantly, _signed_ integer overflow is undefined behaviour. _Unsigned_ integer overflow is perfectly valid.
2. This should be declared as `static`. Otherwise some other file can "see" the variable. If it's static, then it's only visible in this file.

rir/src/interpreter/interp.cpp
```diff
@@ -358,8 +358,17 @@ static RIR_INLINE SEXP rirCallTrampoline(const CallContext& call, Function* fun,
...
+
+// Interrupt Signal Checker - Allows for Ctrl - C functionality to exit out
+// of infinite loops
+void checkuserinterrupt() {
```
mhyee: Our convention is to use camel case, so `checkUserInterrupt`.

rir/src/interpreter/interp.cpp
```diff
@@ -358,8 +358,17 @@ static RIR_INLINE SEXP rirCallTrampoline(const CallContext& call, Function* fun,
...
+    if (++count > 1000) {
+        R_CheckUserInterrupt();
```
mhyee, line 371: Should we also have `R_PendingFinalizers`? What does it do?

rir/src/interpreter/interp.cpp
scottcarroll: No sorry this was removed by mistake

rir/src/interpreter/interp.cpp
scottcarroll: Yes this was just a whitespace change

rir/src/interpreter/interp.cpp
scottcarroll: R_PendingFinalizers is a function inside of GNU-R memory.c. It checks whether finalizers are pending and if they are, runs garbage collection on what is needed inside of R. GNU-R only calls this function when running arbitrary code should be safe. In both cases it's if the boolean IMMEDIATE_FINALIAZERS is true, but I don't see where this boolean/macro is being created. I would imagine adding this to Rir would be safe, but I'm not sure.

mhyee: > It checks whether finalizers are pending and if they are, runs garbage collection on what is needed inside of R.

Yeah, that sounds like we'll want to have it.

> In both cases it's if the boolean IMMEDIATE_FINALIAZERS is true

Actually, the macro is `#ifndef IMMEDIATE_FINALIZERS` which means "if `IMMEDIATE_FINALIZERS` is _not_ defined."

`IMMEDIATE_FINALIZERS` is defined at compile time, probably as a flag. I don't think we're setting it, which means `R_RunPendingFinalizers()` is called.

@o-, @JanJecmen thoughts?

mhyee:
Looks great, thanks!

We changed our continuous integration setup since you opened this PR, and it requires us to create branches from the main repo rather than forks. So I'll open a PR that's a duplicate of this one, just for CI to run, and if it passes, I'll close that PR and merge this one.

PR #399, opened 2019-03-25
Small interp stuff

o-:


rir/src/common.h
```diff
@@ -4,8 +4,7 @@
 #include <cassert>
 #include <cstdint>
 
-// TODO force inlining for clang & gcc
-#define RIR_INLINE __attribute__((always_inline)) inline
+#define RIR_INLINE inline
```
charig, line 7: Yes. According to this: http://gcc.gnu.org/onlinedocs/gcc/Inline.html, we should also start moving macros to inline functions

rir/src/interpreter/interp.cpp
```diff
@@ -1007,11 +1007,13 @@ enum op { PLUSOP, MINUSOP, TIMESOP, DIVOP, POWOP, MODOP, IDIVOP };
     do {                                                                       \
         SEXP a = ostack_at(ctx, 0);                                            \
         SEXP b = ostack_at(ctx, 1);                                            \
-        if (TYPEOF(a) == res_type && NO_REFERENCES(a)) {                       \
+        if (NO_REFERENCES(a)) {                                                \
```
charig, line 1010: Yes. I do not why but I noticed we do much more checks than `gnu-r`.

charig (approved):

rir/src/common.h
o-: I fear we will actually regress a bit over this. but I was getting sick of figuring out where to put the RIR_INLINE's and in the end just making it worse. So lets try this for once and then see what happens.

PR #400, opened 2019-03-25
Force constant promises for better dispatch from GNUR

Jakobeha:
Currently, when dispatching, if an argument is an unbound promise we "give up" and don't assume anything about it. In this PR, we inspect the source of the promise. If it's simply a constant, we will "force" the promise, so that we can dispatch an eager non-obj version of the function.

This makes it so when we run code like `f(10); f(10); f(10)` in GNUR (e.g. from the REPL), PIR can infer that `f` is being called with an eager non-obj, and compile/dispatch a better version of it.

Later, we could also check variables referring to constants by loading them from the environment. If the variable is a promise, instead of forcing it, we'd either give up or "safe force" it again. But in the latter, we should probably have a counter, so we don't waste too much time only to realize we can't actually compute the promise without possible side effects.

I looked at how GNU-R forces promises, from https://github.com/reactorlabs/gnur/blob/c575dedb863481e6b6e904525a44a0fc2d65e7a6/src/main/eval.c.

### DIFF
--- a/rir/src/interpreter/safe_force.cpp
+++ b/rir/src/interpreter/safe_force.cpp
@@ -18,8 +18,8 @@ SEXP safeEval(SEXP e, SEXP rho) {
             // TODO: Is this possible? Should it be assert(false)?
             return R_UnboundValue;
         }
-        Case(BCODESXP) { assert(false); }
-        Case(EXTERNALSXP) { assert(false); }
+        Case(BCODESXP) { return R_UnboundValue; }
+        Case(EXTERNALSXP) { return R_UnboundValue; }
         // TODO : some code (eg. serialize.c:2154) puts closures into asts...
         //        not sure how we want to handle it...
         // Case(CLOSXP) {
--- a/rir/src/interpreter/safe_force.cpp
+++ b/rir/src/interpreter/safe_force.cpp
@@ -8,7 +8,7 @@
 
 namespace rir {
 
-SEXP safeEval(SEXP e, SEXP rho) {
+static SEXP safeEval(SEXP e, SEXP rho) {
     Match(e) {
         // Function application
         Case(LANGSXP) { return R_UnboundValue; }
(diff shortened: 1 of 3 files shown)


rir/src/interpreter/safe_force.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "R/RList.h"
+#include "R/Sexp.h"
+#include "R/Symbols.h"
+#include "R/r.h"
+
+#include <assert.h>
+
+namespace rir {
+
+SEXP safeEval(SEXP e, SEXP rho) {
```
o-: please mark functions only used within one cpp file as static

o- (approved):

Jakobeha:
Applied, should I merge now?

Jakobeha:
Applied changes, should I merge?

o-:
> Applied changes, should I merge?

sure

o-:
@Jakobeha  this seems to regress quite a lot: http://rflies.rir.o1o.ch/changes/?tre=10&rev=3efc5ea1af8e7e62dc5d6145deacae055299dfcd&exe=1&env=1

I think you have to somehow make your check faster.  For many callsites and many argumetns this whole branch (https://github.com/reactorlabs/rir/pull/400/files#diff-0aeab869947b153815cf78fca363e967R38) here will be executed on every invocation. that's just too much overhead and you need to have a quicker check.

something like 
```
if (PRVALUE(a) != R_UnboundValue)
    return PRVALUE(a);
auto t = TYPEOF(PRCODE(a));
if (t == SYMSXP || t == LANGSXP || t == PROMSXP || t == BCODESX || t == EXTERNALSXP)
    return R_UnboundValue;
```
might do the trick. this should speed up the case, where we don't win anything and any further work just adds overhead.

o-:
Also, an additional note: this can only happen when calling from R to rir/pir, right?

so we should probably only do this when calling from R to rir, and never when calling rir -> rir

Jakobeha:
`safeForcePromise` will only be run when we call with stack args, so it shouldn't ever happen in RIR (although later we should probably safe-force the implicit args from RIR too).

It would happen when calling from PIR to PIR though (if PIR already set the promise value it will simply return that, but otherwise it will try to safe force). I added an explicit flag in PIR to `call` which prevents the latter, although I'm not sure if that would add more overhead.

PR #401, opened 2019-03-26
no checkpoints for invisible

o-:
fix a regression: we don't want to insert checkpoints, just to get the
visibility status correct.


o-:
bugfix

PR #402, opened 2019-03-27
Updated safe force

Jakobeha:
Applied suggestions, this should hopefully run faster, it also won't try to safe force promises from PIR calls, only from GNUR calls


o-:
A suggestion: instead of adding it as a flag to CallContext (which probably does impact perf), just make a `safeForceArgs` function, that you call in all the places where you just set the flag to true.

PR #404, opened 2019-03-27
Debug-only opcodes

Jakobeha:
Adds opcodes which will be removed in the release build, which can be inserted by the compiler and used to debug RIR:

- `tmpSet`: Pops and stores an SEXP in a debug pool, at the given index (get a new index via `DebugPool::addTmp`). Doesn't protect the SEXP
- `tmpGet`: Reads and pushes an SEXP from the debug pool, at the given index
- `assert`: Pops an SEXP, asserts it's `R_TrueValue`
- `print`: Pops an SEXP and prints it
- `printStack`: Prints the top 5 stack contents

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -238,6 +238,7 @@ class Instruction : public Value {
     bool usesAreOnly(BB*, std::unordered_set<Tag>);
     bool usesDoNotInclude(BB*, std::unordered_set<Tag>);
     bool unused();
+    virtual bool isDebug() const { return false; }
 
     virtual void updateType(){};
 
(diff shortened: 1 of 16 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -240,6 +240,7 @@ class Instruction : public Value {
     bool usesAreOnly(BB*, std::unordered_set<Tag>);
     bool usesDoNotInclude(BB*, std::unordered_set<Tag>);
     bool unused();
+    virtual bool isDebug() const { return false; }
```
o-, line 243: not neccessary. instead, set effects to `Effects::Any()` to ensure they are not removed.

o- (commented):
I like assert and print. I am not so sure about the others. Can you comment a bit, what you need them for?

print stack you can replace with `pick n; print`*

then the debug pool you can replace with `stvar_super_ someFreshGlobalName`

o-:
actually `print` can be replaced by `call_builtin_ 1 .Internal(Print)`

PR #405, opened 2019-03-27
Removed setshared and fixed `inc`

Jakobeha:


o-:
wait, wait,  d719120 is not necessary! the idea here is, if you have
```
a <- c(1,2,3)
a[[1]] <- 4
```
the vector 123 has only one name (`a`) thus we can override it in place. whereas
```
a <- c(1,2,3)
b <- a
a[[1]] <- 4
```
the vector `b` must stay unchanged, thus we cannot override in place.

so `MAYBE_SHARED` is the correct test here!

o-:
actually, did you look at the failing test?

```
> ## for(var in seq) .. when seq is modified  "inside" :
> x <- c(1,2); s <- 0; for (i in x) { x[i+1] <- i + 42.5; s <- s + i }
```

that should make it pretty clear, why your first commit broke this test and why your second commit avoids the breakage, but is not actually a solution. it seems that the for loop sequence is a bit special, in that it should never be changed while iterating over it.

so the setshared bc actually does have a legitimate use here. but that's probably the only place.

PR #406, opened 2019-03-28
Fix named counting

o-:
fix and improve static reference count

to compute the static reference count we need a static analysis. The
reason is that in

    create a
    loop {
      consume a
    }

we must know that a is used multiple times. Our old one-pass analysis
would think that it is used only once, even though it is in a loop.

on the other hand we can improve the precision with this commit. In both
cases:

    create a
    if (...)
       consume a
    else
       consume a

and

    loop {
      create a
      ...
      consume a
    }

the value a can have only one use at runtime

### DIFF
--- a/rir/src/compiler/analysis/refrence_count.h
+++ b/rir/src/compiler/analysis/refrence_count.h
@@ -1,0 +1,90 @@
+#ifndef PIR_REFERENCE_COUNT_H
+#define PIR_REFERENCE_COUNT_H
+
+#include "../pir/pir_impl.h"
+#include "generic_static_analysis.h"
+
+namespace rir {
+namespace pir {
+
+struct AUses {
+    std::unordered_map<Instruction*, size_t> uses;
+    AbstractResult mergeExit(const AUses& other) { return merge(other); }
+    AbstractResult merge(const AUses& other) {
+        AbstractResult res;
+        for (auto u : other.uses) {
+            if (!uses.count(u.first)) {
+                uses.emplace(u);
+                res.update();
+            } else if (uses.at(u.first) < u.second) {
+                // used in both branches, use count is the maximum
+                uses.at(u.first) = u.second;
+                res.update();
+            }
+        }
+        return res;
+    }
+    void print(std::ostream&, bool) const {
+        // TODO
+    }
+};
+
+class StaticReferenceCount : public StaticAnalysis<AUses> {
+  public:
+    StaticReferenceCount(ClosureVersion* cls, LogStream& log)
+        : StaticAnalysis("StaticReferenceCountAnalysis", cls, cls, log) {
+        Visitor::run(code->entry, [&](Instruction* i) {
+            if (Phi::Cast(i)) {
+                i->eachArg([&](Value* v) {
+                    if (auto a = Instruction::Cast(v))
+                        alias[i].insert(a);
+                });
+            }
+        });
+    }
+
+  protected:
+    std::unordered_map<Instruction*, SmallSet<Instruction*>> alias;
+
+    AbstractResult apply(AUses& state, Instruction* i) const override {
+        AbstractResult res;
+        if (Phi::Cast(i))
+            return res;
+
+        if (i->needsReferenceCount()) {
+            // This value was used only once in a loop, we can thus
+            // reset the count on redefinition.
+            if (state.uses.count(i) && state.uses.at(i) == 1) {
+                state.uses[i] = 0;
+                res.update();
+            }
+        }
+
+        auto count = [&](Instruction* i) {
+            auto& use = state.uses[i];
+            if (use < 2) {
+                use++;
+                res.update();
+            }
+        };
+
+        auto apply = [&](Value* v) {
+            if (auto j = Instruction::Cast(v)) {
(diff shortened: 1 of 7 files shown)


rir/src/compiler/analysis/refrence_count.h
```diff
@@ -0,0 +1,109 @@
...
+  protected:
+    std::unordered_map<Instruction*, SmallSet<Instruction*>> alias;
+
+    AbstractResult apply(AUses& state, Instruction* i) const override {
+        AbstractResult res;
+        if (Phi::Cast(i))
+            return res;
+
+        if (i->needsReferenceCount()) {
+            // This value was used only once in a loop, we can thus
```
JanJecmen, line 74: Don't see the connection to any loop? Is it that if `uses[i] == 1` we know we are in a loop?

rir/src/compiler/analysis/refrence_count.h
```diff
@@ -0,0 +1,109 @@
...
+            // reset the count on redefinition.
+            if (state.uses.count(i) && state.uses.at(i) == 1) {
+                state.uses[i] = 0;
+                res.update();
+            }
+        }
+
+        auto count = [&](Instruction* i) {
+            auto& use = state.uses[i];
+            if (use < 2) {
```
JanJecmen, line 84: BTW, `NAMED_MAX` is 3 in our version of R, but that doesn't matter here I presume?

rir/src/compiler/pir/instruction.h
```diff
@@ -653,6 +656,8 @@ class FLIE(LdVar, 1, Effects() | Effect::Error | Effect::ReadsEnv) {
     size_t gvnBase() const override {
         return hash_combine(InstructionImplementation::gvnBase(), varName);
     }
+
+    bool needsReferenceCount() const override { return false; }
```
JanJecmen, line 660: Why doesn't the `LdVar` need refcount?

rir/src/utils/Set.h
```diff
@@ -60,6 +60,8 @@ class SmallSet {
     iterator end() { return container.end(); }
     const_iterator cbegin() const { return container.cbegin(); }
     const_iterator cend() const { return container.cend(); }
+    const_iterator begin() const { return container.cbegin(); }
+    const_iterator end() const { return container.cend(); }
```
JanJecmen, line 64: Why are these also constant iterators?

JanJecmen (approved):

rir/src/compiler/analysis/refrence_count.h
o-: so the idea here is if we have:
```
loop {
  produce a
  ...
  consume a
  ...
}
```
then on the second iteration in the analysis when we see `produce a` the count will be 1. in this case it is safe to reset it to 0, since it was only used once. but if there are 2 (or more) `consume a` we should not reset it, because it really needs to be marked as a shared value

rir/src/compiler/analysis/refrence_count.h
o-: hmm, good point. at least rir only discriminates between 0,1 and more. I don't know of any situation where we could do something else if it is 2 vs. 3, do you?

rir/src/compiler/pir/instruction.h
o-: ldvar values already have a guaranteed named count of 1. ensure named will bump the refcount to 1, so it should never be neccessary for ldvar values.

rir/src/utils/Set.h
o-: tbh i don't fully understand the interplay of constant and iterators. suggestions?

mhyee: [Looking at the C++ API for vector iterators, this looks right](https://en.cppreference.com/w/cpp/container/vector/begin).

`begin()`/`end()` can be const or non-const, and it depends on the variable you're declaring. In the examples below, the compiler will pick the right version of `begin()`/`end()`:

```
for (const auto& b : bar) {
  // uses the const versions of begin()/end()
  // b is const, not allowed to modify b or bar
}
for (auto& b : bar) {
  // uses the non-const versions of begin()/end()
  // b is non-const, can modify it
}
```

But sometimes it's not clear:

```
for (auto it = bar.begin(); it != bar.end(); ++it) {
  // it is a pointer, and you can modify it and *it
}
for (const auto it = bar.begin(); it != bar.end(); ++it) {
  // this won't compile, because it is const so ++it is illegal!
}
for (const auto it = bar.cbegin(); it != bar.cend(); ++it) {
  // now this works
  // allowed to modify it, but not *it (or bar)
}
```

o-: ok, so cbegin/cend is just for when the compiler fails to pick the right one. thanks

rir/src/compiler/analysis/refrence_count.h
JanJecmen: Not really.. the only thing is we could sometimes decrement the named maybe? Because 2 really should mean 2 names of the value, not at least 2.

o-: yeah, this might be an interesting optimization to add, if we find such cases

PR #407, opened 2019-03-28
fix inc_ and setShared_

o-:
inc can only modify in-place if the reference count is zero. Thanks
@Jakobeha for spotting the bug.

Also setShared should (as the name suggests) not randomly increment, but
actually just ensure that named is at least 2.

also remove unused bc makeUnique

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -2190,13 +2190,13 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP val = ostack_top(ctx);
             assert(TYPEOF(val) == INTSXP);
             int i = INTEGER(val)[0];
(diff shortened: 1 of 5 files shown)


rir/src/interpreter/interp.cpp
```diff
@@ -2190,7 +2190,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             SEXP val = ostack_top(ctx);
             assert(TYPEOF(val) == INTSXP);
             int i = INTEGER(val)[0];
-            if (MAYBE_SHARED(val)) {
+            if (!NO_REFERENCES(val)) {
```
mhyee: Can you flip the condition? Double negatives confuse me. :)

o-: who wouldn't not hate them too?

PR #409, opened 2019-03-28
[wip] initial version of native backend

o-:
disabled by default


o-:
probably not worth reviewing yet. I should polish it a bit at least. Mainly put it here to test if it builds on the ci and others want to play with it...

PR #410, opened 2019-03-29
inc is intended to increment in-place

o-:
I remembered, that inc_ is supposed to update in-place, even if the
refcount is 1.

That was a feature, not a bug, to be able to increment the loop counter
without creating a new vector all the time.

So the idea is we compile the for loop as follows:

    %i = 0
    while (%i < size(seq))
      ...
      %i++

now, we need to mark the (stack) value %i ensureNamed, because
otherwise the < comparison might override it (values with refcount 0
can be reused). But then the i++ at the end of the loop, should
still be able to increment i without duplicating it. so the inc
bytecode is allowed to destructively modify it

but, the pir version of inc was broken, because the pir optimizer did
not preserve this invariant, that an inc'd value is not used anymore
after the operation.

for now I replaced it by a normal add. this might have potential
performance issues, but we need to solve them differently.

### DIFF
--- a/rir/src/compiler/analysis/refrence_count.h
+++ b/rir/src/compiler/analysis/refrence_count.h
@@ -8,7 +8,14 @@ namespace rir {
 namespace pir {
 
 struct AUses {
-    std::unordered_map<Instruction*, size_t> uses;
+    enum Kind {
+        None = 0,
+        Once = 1,
+        Multiple = 2,
+        Destructive = 3,
+    };
+
+    std::unordered_map<Instruction*, Kind> uses;
     AbstractResult mergeExit(const AUses& other) { return merge(other); }
     AbstractResult merge(const AUses& other) {
         AbstractResult res;
@@ -73,17 +80,24 @@ class StaticReferenceCount : public StaticAnalysis<AUses> {
         if (i->needsReferenceCount()) {
             // This value was used only once in a loop, we can thus
             // reset the count on redefinition.
-            if (state.uses.count(i) && state.uses.at(i) == 1) {
-                state.uses[i] = 0;
+            if (state.uses.count(i) && state.uses.at(i) == AUses::Once) {
+                state.uses[i] = AUses::None;
                 res.update();
             }
         }
 
         auto count = [&](Instruction* i) {
             auto& use = state.uses[i];
-            if (use < 2) {
-                use++;
+            switch (use) {
+            case AUses::None:
+                use = AUses::Once;
                 res.update();
+                break;
+            case AUses::Once:
+                use = AUses::Multiple;
+                res.update();
+                break;
+            default: {}
             }
         };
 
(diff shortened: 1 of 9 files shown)


o-:
fixes broken #407

o-:
just realized that the same problem applies to subassign, they are also allowed to update inplace. so I had to actually fix it properly at the pir level. whenever an input to a in-place modifying bc is used elsewhere, we need to bump the named count to 2.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -706,12 +706,12 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
...
     });
 
-    std::unordered_map<Instruction*, bool> needsEnsureNamed;
+    std::unordered_map<Instruction*, size_t> needsRefcount;
     {
         StaticReferenceCount analysis(cls, log);
         for (auto& u : analysis.result().uses) {
             if (u.second > 1)
-                needsEnsureNamed[u.first] = u.second;
+                needsRefcount[u.first] = u.second - 1;
```
Jakobeha: If an instruction is used multiple times, but not by subassign, does it still need `setShared`?

rir/src/compiler/analysis/refrence_count.h
```diff
@@ -98,8 +98,28 @@ class StaticReferenceCount : public StaticAnalysis<AUses> {
...
+        // multiple times, we need to bump the refcount even further.
+        switch (i->tag) {
+        case Tag::Subassign1_1D:
+        case Tag::Subassign2_1D:
+        case Tag::Subassign1_2D:
+        case Tag::Subassign2_2D:
+        case Tag::Inc:
+            if (auto input = Instruction::Cast(i->arg(0).val())) {
+                if (state.uses.count(input) && state.uses.at(input) == 2) {
+                    state.uses[input] = 3;
```
Jakobeha: maybe rename `uses`, since an instruction will have `uses = 2` if it's used 2+ times but not by a special opcode, and 3+ if it's used 2* by a special opcode.

Also, if `uses = 1` would the special opcode still override? I understand instructions are processed in order, but is this handled:
```r
while (...) {
   use(i)
   i[[n]] <- ...
}
```

rir/src/ir/insns.h
```diff
@@ -242,7 +242,7 @@ DEF_INSTR(add_, 0, 2, 1, 0)
 DEF_INSTR(uplus_, 0, 1, 1, 0)
 
 /**
- * inc_ :: increment tos integer
+ * inc_ :: increment tos integer (destructive! ignores refcount)
```
Jakobeha, line 245: Also should mention this for subassign

rir/src/interpreter/interp.cpp
```diff
@@ -2188,15 +2188,19 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
...
-            assert(TYPEOF(val) == INTSXP);
-            int i = INTEGER(val)[0];
-            if (NO_REFERENCES(val)) {
-                INTEGER(val)[0]++;
-            } else {
+            SLOWASSERT(TYPEOF(val) == INTSXP);
+            // Inc_ destructively modifies TOS, even if the refcount is 1. This
+            // can be used to perform `++i` on a value on the stack. The old i
+            // value will be overwritten (generally only do this if you are sure
+            // that this is the last copy on the stack).
```
Jakobeha, line 2171: I don't think this is necessary because you mention the same in `insns.h`. Otherwise, also mention for `subassign`

o-: I think this is *very* neccessary, since we both tried to "fix" this

rir/src/compiler/analysis/refrence_count.h
o-: the static analysis is a fix-point computation, not one pass. so it will take two passes:

first pass it goes from `0 -> 1 -> 2`, then second pass `2 -> 2 -> 3`

mhyee:
Typo in the filename: `refrence_count.h` -> `reference_count.h`

PR #411, opened 2019-03-29
type inference pass

o-:
Add a type inference pass that can deal with recursive dependencies
introduced by phis.

Also improve and fix the type annotations on some instructions.

The improved types do not change much yet, but they are important for
native compilation.

### DIFF
--- a/rir/src/compiler/opt/inline.cpp
+++ b/rir/src/compiler/opt/inline.cpp
@@ -211,18 +211,26 @@ class TheInliner {
                         if (ld) {
                             Value* a = arguments[ld->id];
                             if (auto mk = MkArg::Cast(a)) {
-                                // We need to cast from a promise to a lazy
-                                // value
-                                auto cast =
-                                    new CastType(a, RType::prom,
-                                                 mk->isEager()
-                                                     ? mk->eagerArg()
-                                                           ->type.forced()
-                                                           .promiseWrappedVal()
-                                                     : ld->type);
-                                ip = bb->insert(ip + 1, cast);
-                                ip--;
-                                a = cast;
+                                if (!ld->type.maybePromiseWrapped()) {
+                                    // This load already expects to load an
+                                    // eager value. We can just discard the
+                                    // promise altogether.
+                                    assert(mk->isEager());
+                                    a = mk->eagerArg();
+                                } else {
+                                    // We need to cast from a promise to a lazy
+                                    // value
+                                    auto type = mk->isEager()
+                                                    ? mk->eagerArg()
+                                                          ->type.forced()
+                                                          .orPromiseWrapped()
+                                                    : ld->type;
+                                    auto cast =
+                                        new CastType(a, RType::prom, type);
+                                    ip = bb->insert(ip + 1, cast);
+                                    ip--;
+                                    a = cast;
+                                }
                             }
                             ld->replaceUsesWith(a);
                             next = bb->remove(ip);
--- a/rir/src/compiler/translations/pir_2_rir/liveness.cpp
+++ b/rir/src/compiler/translations/pir_2_rir/liveness.cpp
@@ -9,6 +9,7 @@ namespace rir {
 namespace pir {
 
 LivenessIntervals::LivenessIntervals(unsigned bbsSize, CFG const& cfg) {
+    maxLive = 0;
 
     // temp list of live out sets for every BB
     std::unordered_map<BB*, std::set<Value*>> liveAtEnd(bbsSize);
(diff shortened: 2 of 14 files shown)


rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -964,8 +965,11 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
             }
 
             case Tag::MkArg: {
-                cb.add(BC::promise(ctx.cs().addPromise(
-                    getPromise(ctx, MkArg::Cast(instr)->prom()))));
+                auto p = MkArg::Cast(instr)->prom();
+                unsigned id = ctx.cs().addPromise(
+                    getPromise(ctx, MkArg::Cast(instr)->prom()));
```
mhyee: `getPromise(ctx, p);`?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -1420,9 +1426,8 @@ rir::Function* Pir2Rir::finalize() {
 
     assert(signature.formalNargs() == cls->nargs());
     ctx.push(R_NilValue);
-    size_t localsCnt = compileCode(ctx, cls);
+    auto body = compileCode(ctx, cls);
```
mhyee, line 1426: What was the reasoning for making these changes in this file?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -699,6 +699,7 @@ size_t Pir2Rir::compileCode(Context& ctx, Code* code) {
 
     LastEnv lastEnv(cls, code, log);
     std::unordered_map<Value*, BC::Label> pushContexts;
+    std::unordered_map<Promise*, unsigned> promMap;
```
mhyee: What is the purpose of `promMap`?

rir/src/compiler/opt/inline.cpp
```diff
@@ -211,18 +211,26 @@ class TheInliner {
...
-                                    new CastType(a, RType::prom,
-                                                 mk->isEager()
-                                                     ? mk->eagerArg()
-                                                           ->type.forced()
-                                                           .promiseWrappedVal()
-                                                     : ld->type);
-                                ip = bb->insert(ip + 1, cast);
-                                ip--;
-                                a = cast;
+                                if (!ld->type.maybePromiseWrapped()) {
```
mhyee, line 214: Is this fixing a bug somewhere?

rir/src/compiler/translations/pir_2_rir/liveness.cpp
```diff
@@ -9,6 +9,7 @@ namespace rir {
 namespace pir {
 
 LivenessIntervals::LivenessIntervals(unsigned bbsSize, CFG const& cfg) {
+    maxLive = 0;
```
mhyee: Where is this being used?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: ok, this is slightly unrelated, because I pulled those changes out of my native branch.

The api was a bit strange in the first place. Why should a compile method return as the result the number of locals the code needs? that is very accidental api design.

I had to change it for native compilation and thought, why not fix it already now..

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: it's probably dead code in this pr. removing.

rir/src/compiler/opt/inline.cpp
o-: no, but it's a bit faster. In the case the promise is eager, and the callee was already compiled under the assumption that it will receive an already evaluated promise.

Instead of:
```
// caller
val  e = ldconst 123
prom p = mkpromise(e, ...)
//callee
val^ l = castType(p, lazy)
val  v = force(l)
         use(v)
```

we directly generate this

```
val  e = ldconst 123
prom p = mkpromise(e, ...)
val  v = force(e)
         use(v)
```

both of those will end up being converted to 

```
val  e = ldconst 123
         use(e)
```

just the fromer should avoid some extra steps

rir/src/compiler/translations/pir_2_rir/liveness.cpp
o-: admittedly not in this pr. do you mind?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
mhyee: 👍

rir/src/compiler/translations/pir_2_rir/liveness.cpp
mhyee: Sorry, but I'd prefer having this change in the commit that uses it (otherwise it looks like dead code). This one is small enough that it shouldn't be too much trouble... (famous last words)

PR #412, opened 2019-03-29
PIR tests in R

Jakobeha:
Adds a new function, `pir.check(f, CheckType...)`, which will compile `f` and inspect the PIR to make sure it adheres to each `CheckType`.

For example, `pir.check(f, NoLoad, NoStore)` returns `TRUE` if `f` is a closure which can be compiled and its PIR doesn't have any loads or stores, `FALSE` otherwise. So `stopifnot(pir.check(f, NoLoad, NoStore))` fails if `f`'s loads/stores can't be eliminated.

I tested this by reusing most of the `pir_tests`, I also added a few new tests.

### DIFF
--- a/rir/src/compiler/test/PirCheck.h
+++ b/rir/src/compiler/test/PirCheck.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "R/Symbols.h"
+#include "common.h"
 #include <list>
 
 namespace rir {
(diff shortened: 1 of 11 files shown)


rir/src/compiler/test/PirCheck.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "R/Symbols.h"
+#include "common.h"
+#include <list>
+
+namespace rir {
+
+#define LIST_OF_PIR_CHECKS(V)                                                  \
```
o-, line 9: I mean, this trick is nice and all. But in general macros are a place of last resort in C++. To be used if nothing else works. In this case I think a simple static array of tests would have worked just fine. somthing like 
```
typedef pair<string, function<...>>;
constexpr static  Test tests[] = {
  Test("NoLoad", &noLoad),
  Test("NoStore", [&](ClosureVersion* cls){Visitor::check(cls, ...))}
```

PR #414, opened 2019-03-31
int and real scalar type assumptions

Jakobeha:


rir/src/compiler/pir/type.h
```diff
@@ -345,6 +346,36 @@ struct PirType {
         return t_.r.includes(o.t_.r);
     }
 
+    void addAssumptions(Assumptions& given, int idx) const {
+        if (!maybeObj())
+            given.setNotObj(idx);
+        if (given.isEager(idx) && given.isNotObj(idx) && isScalar()) {
```
o-: I don't understand this. why are some properties read from `given`, some other properties from type, then some other properties are updated. I see no logic. If this is the particular ad-hoc thing, that the builder needs, then better just keep this code at the builder site. because putting it on type suggests, that this is some generic functionality, when I really don't understand what the reusable property of this function would be...

rir/src/compiler/pir/type.h
```diff
@@ -345,6 +346,36 @@ struct PirType {
...
+            assert(isRType());
+            if (t_.r == RTypeSet(RType::real))
+                given.setSimpleReal(idx);
+            if (t_.r == RTypeSet(RType::integer))
+                given.setSimpleInt(idx);
+        }
+    }
+
+    void loadAssumptions(const Assumptions& given, int idx) {
```
o-: this seems at least a more or less a reusable thing. But I don't particularly like that it's here. Does it make sense to tightly couple assumptions and types? would it be better to keep those two things unrelated?

o- (commented):
looks good, modulo comments

o-:
thanks, no need to ask for another review, when I already said that it's good

PR #416, opened 2019-04-02
Prevent stvar from bumping named count and fix static reference counts

o-:
The main changes here are:

### Fix named count for stvar

this was completely bogus. in ir/compiler.cpp we used to emit an `BC::ensureNamed()` before `stvar`. This means before stvar, the named count was already 1 and the stvar would *always* bump it to 2...

I changed this with the following consequences:

* the static refcount analysis was revealed to be broken
* the subassign case in ir/compiler.cpp was revealed to be broken

### Issues with subassign in RIR

In this case:

```
     a <- 1
     f <- function()
         a[[1]] <- 2
```
in the inner function, we load `a`, then do an `subassign`, which will override in-place, because the named count is 1. Only after the subassign, we store back `a` in the inner scope and realize that named is now 2 (too late).

To fix this problem I added a `ldVarForUpdate` bytecode that bumps the named count if the variable is loaded from an outer scope.

The same needs to be done in Pir2Rir, when we cannot statically rule out that the vector is from an outer scope.

### Issues with static refcount analysis

1. the merge was not forming a lattice, taking long to converge in some cases
2. trying to figure out if subassign is changing a value is bound to be doomed. a simple counterexample produced by jakobeha for the inc_ bytecode shows this (see comment below). the problem is we would have to also track bindings, which we don't want to do outside of the scope analysis. Therefore we completely remove that part from the refcount analysis.
3. it was very imprecise, often way too conservative (e.g. ensureNamed for values which flow into stvar, which is guaranteed to increment refcount)
4. the abstract states get huge for big functions. we need an upper limit where we give up to not take forever on gitlab-ci.

### DIFF
--- a/rir/src/compiler/analysis/reference_count.h
+++ b/rir/src/compiler/analysis/reference_count.h
@@ -8,29 +8,92 @@ namespace rir {
 namespace pir {
 
 struct AUses {
+
+    /*
+     * Lattice:
+     *
+     *                    Multiple
+     *                     /      \
+     *   AlreadyIncremented        Once
+     *                     \      /
+     *                       None
+     *
+     * Definitions:
+     *   reuse            :=   override if refcount is == 0
+     */
+
     enum Kind {
-        None = 0,
-        Once = 1,
-        Multiple = 2,
-        Destructive = 3,
+        // 1) Static refcount is safe
+        None,
+        Once,               // At least one use, which reuses input
+        AlreadyIncremented, // Refcount is known to be 1 (ie. used by stvar)
+
+        // 2) Need an ensureNamed
+        Multiple, // At least two uses, at least one (but the last) is reusing
     };
 
-    std::unordered_map<Instruction*, Kind> uses;
+    static Kind merge(Kind a, Kind b) {
+        if (a == b)
+            return a;
+
+        Kind bigger = a > b ? a : b;
+        Kind smaller = a < b ? a : b;
+        if (smaller == None)
+            return bigger;
+
+        if (bigger == Multiple || bigger == AlreadyIncremented)
+            return Multiple;
+
+        assert(false);
+        return Multiple;
+    };
+
+    bool overflow = false;
+    std::map<Instruction*, Kind> uses;
     AbstractResult mergeExit(const AUses& other) { return merge(other); }
     AbstractResult merge(const AUses& other) {
         AbstractResult res;
-        for (auto u : other.uses) {
-            if (!uses.count(u.first)) {
-                uses.emplace(u);
+
+        if (overflow)
+            return res;
+
+        if (other.overflow) {
+            uses.clear();
+            overflow = true;
+            res.taint();
+            return res;
+        }
+
+        auto myPos = uses.begin();
+        auto theirPos = other.uses.begin();
+
+        while (theirPos != other.uses.end()) {
+            if (myPos == uses.end() || myPos->first > theirPos->first) {
+                // we are missing this entry
+                auto updated = uses.insert(*theirPos);
+                myPos = updated.first;
                 res.update();
-            } else if (uses.at(u.first) < u.second) {
-                // used in both branches, use count is the maximum
-                uses.at(u.first) = u.second;
-                res.update();
+                myPos++;
+                theirPos++;
+            } else if (myPos->first == theirPos->first) {
+                if (myPos->second != theirPos->second) {
+                    auto m = merge(myPos->second, theirPos->second);
+                    if (m != myPos->second) {
+                        myPos->second = m;
+                        res.update();
+                    }
+                }
+                myPos++;
+                theirPos++;
+            } else {
+                assert(myPos->first < theirPos->first);
+                // other branch does not have this entry
+                myPos++;
             }
         }
         return res;
     }
+
     void print(std::ostream&, bool) const {
         // TODO
     }
@@ -51,14 +114,14 @@ class StaticReferenceCount : public StaticAnalysis<AUses> {
                         if (auto a = Instruction::Cast(v)) {
                             if (Phi::Cast(a)) {
                                 for (auto otherAlias : alias[a]) {
-                                    if (otherAlias->needsReferenceCount() &&
+                                    if (otherAlias->minReferenceCount() < 1 &&
                                         !alias[i].includes(otherAlias)) {
                                         changed = true;
                                         alias[i].insert(otherAlias);
                                     }
                                 }
                             } else {
-                                if (a->needsReferenceCount() &&
+                                if (a->minReferenceCount() < 1 &&
                                     !alias[i].includes(a)) {
                                     changed = true;
                                     alias[i].insert(a);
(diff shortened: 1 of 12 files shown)


Jakobeha:
I found another bug in the reference count, which is currently affecting simple ranges - whenever we find an, `LdVar`, we need to ensure there aren't any potential later `LdVar`s, otherwise it needs multiple references.

Example:

```
%0.0 = LdVar i
%0.1 = Inc %0.0
%0.2 = LdVar i
```

In this case, `%0.2` could get corrupted.

If `i` was converted into a PIR variable, there would be a `Phi` so the analysis would realize that it's used multiple times and insert a `setShared` before the `inc`

o-:
hey, @mhyee I am merging this tbr, because it finally fixes the compilation time issues that caused the ci to pile up jobs and run into timeouts.

rir/src/ir/insns.h
```diff
@@ -49,6 +49,12 @@ DEF_INSTR(set_env_, 0, 1, 0, 0)
  */
 DEF_INSTR(ldfun_, 1, 0, 1, 0)
 
+/**
+ * ldvar_:: take immediate CP index of symbol, finding binding in env and push.
```
mhyee, line 53: `ldvar_` -> `ldvar_for_update_`

Might also be clearer to move this below the `ldvar_` definition and document it similarly to `ldvar_noforce_`, e.g. "like ldvar_ but increment named count if the variable is not local."

rir/src/ir/Compiler.cpp
```diff
@@ -142,8 +142,8 @@ class CompilerContext {
 };
 
 Code* compilePromise(CompilerContext& ctx, SEXP exp);
-void compileExpr(CompilerContext& ctx, SEXP exp);
-void compileCall(CompilerContext& ctx, SEXP ast, SEXP fun, SEXP args);
+void compileExpr(CompilerContext& ctx, SEXP exp, bool voidContext = false);
```
mhyee, line 145: Looking at the code, I'm having a hard time understanding _when_ I should be setting this flag to `true` or `false`. So either the API or documentation (or both) should be improved.

rir/src/interpreter/interp.cpp
```diff
@@ -1690,6 +1694,49 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
...
+            }
+
+            // if promise, evaluate & return
+            if (TYPEOF(res) == PROMSXP)
+                res = promiseValue(res, ctx);
+
+            if (res != R_NilValue) {
+                if (isLocal) {
+                    ENSURE_NAMED(res);
+                } else {
```
mhyee, line 1730: `} else if (NAMED(res) < 2) {`

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -917,7 +961,10 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
 
             case Tag::LdVar: {
                 auto ldvar = LdVar::Cast(instr);
-                cb.add(BC::ldvarNoForce(ldvar->varName));
+                if (needsLdVarForUpdate.count(instr))
+                    cb.add(BC::ldvarNoForce(ldvar->varName));
```
mhyee, line 965: Should this be swapped with the other case? "if needs ldvarForUpdate, add ldvarNoForce, else add ldvarForUpdate" ??

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -1243,12 +1290,12 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
...
-            if (instr->needsReferenceCount()) {
-                if (needsRefcount[instr] == AUses::Multiple)
-                    cb.add(BC::ensureNamed());
-                else if (needsRefcount[instr] == AUses::Destructive)
-                    cb.add(BC::setShared());
-            }
+            if (instr->minReferenceCount() < 2 && needsSetShared.count(instr))
+                cb.add(BC::setShared());
+            else if (instr->minReferenceCount() < 1 &&
+                     (refcountAnalysisOverflow ||
```
mhyee, line 1296: Is `ensureNamed` the right thing to do when the analysis overflows? I haven't looked at the analysis (yet), but I would expect overflow => give up => pick the most conservative option => `setShared`.

rir/src/compiler/analysis/reference_count.h
```diff
@@ -8,29 +8,92 @@ namespace rir {
...
+        Kind smaller = a < b ? a : b;
+        if (smaller == None)
+            return bigger;
+
+        if (bigger == Multiple || bigger == AlreadyIncremented)
+            return Multiple;
+
+        assert(false);
+        return Multiple;
+    };
```
mhyee, line 49: Unnecessary semicolon.

rir/src/compiler/analysis/reference_count.h
```diff
@@ -8,29 +8,92 @@ namespace rir {
...
-        Once = 1,
-        Multiple = 2,
-        Destructive = 3,
+        // 1) Static refcount is safe
+        None,
+        Once,               // At least one use, which reuses input
+        AlreadyIncremented, // Refcount is known to be 1 (ie. used by stvar)
+
+        // 2) Need an ensureNamed
+        Multiple, // At least two uses, at least one (but the last) is reusing
```
mhyee, line 32: Nitpick: comment not lined up with the other ones above.

rir/src/compiler/analysis/reference_count.h
```diff
@@ -8,29 +8,92 @@ namespace rir {
...
+
+        if (bigger == Multiple || bigger == AlreadyIncremented)
+            return Multiple;
+
+        assert(false);
+        return Multiple;
+    };
+
+    bool overflow = false;
+    std::map<Instruction*, Kind> uses;
```
mhyee, line 52: If this needs to be ordered, a comment saying so would be nice, otherwise someone might change it to `unordered_map`.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: the refcount analysis is only concerned with NAMED 0 vs. 1.  To decide 1 vs. 2 a simpler and more conservative approach is needed, because a precise approach would need to track bindings...

rir/src/compiler/analysis/reference_count.h
o-: this is clang format unfortunately

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
mhyee: Sorry to nitpick again, but can you add a comment for this? I still think the overflow behaviour is a bit surprising. (Just sneak it into your next commit or commit directly to master.)

PR #417, opened 2019-04-02
scope analysis uses wrong state for recursive lookup

o-:
recursive lookup in scope analysis is bogus, since the state of the
initial instruction is used to query the state of earlier instructions.

for example in:

  a <- 1
  b <- a
  a <- 2
  a + b

we would resolve "b" in the last line to "ldvar a", then we use the same
state again to lookup "a", which would return 2 instead of 1.

Thus the recursive lookup is bogus and needs to be removed.


o-:
just bugfixes

PR #419, opened 2019-04-03
Simple Ranges

Jakobeha:
Should fix all diffs, most of the extra changes are just clang format


o-:
cool, just merge when green!

PR #421, opened 2019-04-03
Refactor liveness

mhyee:
I moved the liveness analysis because I'll need it for loop peeling.

I also added more documentation, to prevent other people from going on a wild goose chase like I did.

I did not fix the "you can only query after an instruction" because there is no need for it (yet); it'll be in the loop peeling PR.

You can look at the individual commits to separate the move from the changes.

### DIFF
--- a/rir/src/compiler/analysis/liveness.cpp
+++ b/rir/src/compiler/analysis/liveness.cpp
@@ -1,6 +1,6 @@
 #include "liveness.h"
-#include "../../pir/bb.h"
-#include "../../pir/instruction.h"
+#include "../pir/bb.h"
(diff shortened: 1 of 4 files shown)


mhyee:
Travis timed out, and then I remembered I wanted to make more commits long after the other CI tests had finished...

I'm generally opposed to inheriting from things in `std`, and the last commit rips this out. But if you disagree, I can remove the commit from the PR.

o-:
> I'm generally opposed to inheriting from things in std, and the last commit rips this out. But if you disagree, I can remove the commit from the PR.

fully agree. actually, in general I prefer delegation over inheritance whenever possible.

PR #422, opened 2019-04-03
debugopt still too slow for pre-push, use releaseassert instead

mhyee:


mhyee:
I cancelled the CI to save CPU time, since this change only affects the pre-push hook and not the actual tests.

charig (approved):

charig:
Thanks!

PR #423, opened 2019-04-04
avoid allocating a real arglist for binops

o-:
when the binops do not dispatch, the builtins do not leak the argslist.
Thus it should be safe to not allocate them on the R heap, but just
create them locally.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -933,6 +933,26 @@ static R_INLINE int R_integer_times(int x, int y, Rboolean* pnaflag) {
 enum op { PLUSOP, MINUSOP, TIMESOP, DIVOP, POWOP, MODOP, IDIVOP };
 #define INTEGER_OVERFLOW_WARNING "NAs produced by integer overflow"
 
+static SEXPREC createFakeSEXP(SEXPTYPE t) {
+    SEXPREC res;
+    res.attrib = R_NilValue;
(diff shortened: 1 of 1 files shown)


o-:
@JanJecmen what do you think about this one? think it's safe? reasonable?

rir/src/interpreter/interp.cpp
```diff
@@ -930,6 +930,25 @@ static R_INLINE int R_integer_times(int x, int y, Rboolean* pnaflag) {
 enum op { PLUSOP, MINUSOP, TIMESOP, DIVOP, POWOP, MODOP, IDIVOP };
 #define INTEGER_OVERFLOW_WARNING "NAs produced by integer overflow"
 
+static SEXPREC createFakeSEXP(SEXPTYPE t) {
+    SEXPREC res;
+    res.attrib = R_NilValue;
+    res.gengc_next_node = R_NilValue;
+    res.gengc_prev_node = R_NilValue;
+    res.sxpinfo.gcgen = 1;
+    res.sxpinfo.mark = 1;
```
JanJecmen, line 942: Maybe sth like `res.sxpinfo.named = 2;` just to be safe?

JanJecmen (approved):
Hey, I like the idea, never ocurred to me to fake SEXPs.. I had a look at R binops and don't see why it wouldn't work.
I am curious if it has any performance impacts, I got the impression that asking R for a cons cell is pretty fast

PR #424, opened 2019-04-04
Fixed static_cast_ assertion

Jakobeha:
Previously, if we had a promise which resolved to a missing, we wouldn't remove `NoExplicitlyMissingArgs`. The problem was, my old branch changed that. This changes it back.

However, that could actually be a bug. If a promise resolves to a missing arg, wouldn't that assumption be wrong?

Another potential problem, I overloaded `PirType == RType`, and that was used in some places to get `PirType == PirType(RType)`, which might've had different behavior. I changed my overload to `PirType.isRType(RType)` to prevent confusion and in case that was the bug.

### DIFF
--- a/rir/src/compiler/pir/type.h
+++ b/rir/src/compiler/pir/type.h
@@ -233,6 +233,9 @@ struct PirType {
     RIR_INLINE constexpr bool isRType() const {
         return flags_.includes(TypeFlags::rtype);
     }
+    RIR_INLINE bool isRType(const RType& o) const {
+        return isRType() && t_.r == o;
+    }
     RIR_INLINE constexpr bool maybe(RType type) const {
         return isRType() && t_.r.includes(type);
     }
@@ -329,10 +332,6 @@ struct PirType {
 
     static const PirType voyd() { return PirType(NativeTypeSet()); }
     static const PirType bottom() { return optimistic(); }
-
-    RIR_INLINE bool operator==(const RType& o) const {
-        return isRType() && t_.r == o;
-    }
 
     RIR_INLINE bool operator==(const NativeType& o) const {
         return !isRType() && t_.n == o;
--- a/rir/src/compiler/util/ConvertAssumptions.cpp
+++ b/rir/src/compiler/util/ConvertAssumptions.cpp
@@ -35,9 +35,9 @@ void writeArgTypeToAssumptions(Assumptions& assumptions, Value* arg, int i) {
         if (assumptions.isEager(i) && assumptions.isNotObj(i) &&
             value->type.isScalar()) {
             assert(value->type.isRType());
-            if (value->type == RType::real)
+            if (value->type.isRType(RType::real))
                 assumptions.setSimpleReal(i);
-            if (value->type == RType::integer)
+            if (value->type.isRType(RType::integer))
                 assumptions.setSimpleInt(i);
         }
     }
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -570,11 +570,11 @@ static void addDynamicAssumptionsFromContext(CallContext& call) {
                     notObj = false;
                     isEager = false;
                 }
+            } else if (arg == R_MissingArg) {
+                given.remove(Assumption::NoExplicitlyMissingArgs);
             }
             if (isObject(arg)) {
                 notObj = false;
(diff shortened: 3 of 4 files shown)


o-:
did you check if this change improves pidigits perf?

my theory was more, that your change accidentally disabled some optimization.

charig:
A couple of small comments on this one:

1. Maybe is time to make Assumptions a real class. Or move the methods from ConvertAssumption to the struct.

2. Changing the initial assumptions when compiling (rir2pir) triggers this problem with the dispatch chain. I already observed that. It also leads to the situation in which the lowering has to resorts to the baseline in a couple of cases. It would be good to know why in some few cases changing that initial assumptions make the system fail.

o-:
> Maybe is time to make Assumptions a real class. Or move the methods from ConvertAssumption to the struct.

why? I fail to see the connection to this PR.

> Changing the initial assumptions when compiling (rir2pir) triggers this problem with the dispatch chain ...

for all I know, we don't even know what's the problem. I would like to know what is the reason that pidigits regressed. And i would like to know by direct observation. and not by changing other parts of the system and then it disappears for who knows what reason.

If that is not possible, then the correct short term fix is to just revert #414

charig:
> > Maybe is time to make Assumptions a real class. Or move the methods from ConvertAssumption to the struct.
> 
> why? I fail to see the connection to this PR.

The original PR, that triggerd this one, was already merged and introduces the ConvertAssumption files.

> 
> > Changing the initial assumptions when compiling (rir2pir) triggers this problem with the dispatch chain ...
> 
> for all I know, we don't even know what's the problem. I would like to know what is the reason that pidigits regressed. And i would like to know by direct observation. and not by changing other parts of the system and then it disappears for who knows what reason.
> 
> If that is not possible, then the correct short term fix is to just revert #414

This is exactly what I said.  Removing the assert from pir2rir when there is an assumption mismatch is conceptually sound but it was there because there was still an issue not fully understood. So, it may not be only a problem in pidgits.

o-:
> This is exactly what I said.

well, my point was, we do not know if this is the problem. I agree that it's a good candidate. but, step 1. identify the actual problem, then find theories on why it happens.

> Removing the assert from pir2rir when there is an assumption mismatch is conceptually sound but it was there because there was still an issue not fully understood.

I didn't realize that we removed that assert in #414. In that case, that is indeed a good candidate. can we please add that assert again then!

If I look at: https://github.com/reactorlabs/rir/pull/414/files#diff-4ea831d48309a267f16a0867327b2357R666
then it seems quite strange. e.g. the write arg type to assumption doesn't actually look at the type, but instead only does something for MkArg...

incidentally this was my exactly my comment on the review, "writeArgTypeToAssumptions" and "readArgType" are ad-hoc and not reusable. and it seems the problem is that writeArgTypeToAssumptions was written for the rir2pir pass and is useless in the other case.

Jakobeha:
> I didn't realize that we removed that assert in #414. In that case, that is indeed a good candidate. can we please add that assert again then!

I removed the assert because it was failing and static call was reverting to baseline. Now that I look back that's probably related to why the change made pidgits slower.

Another theory is, #414 makes it so we only assume the first 3 arguments, previously we would assume the first 4. I had to do this because we store flags in a 16-bit uint, so 3 args = 16 assumptions flags.

But now that I'm looking at `EnumSet`, using all 16 bits might be the problem. Even if `EnumSet` doesn't store any extra information, I know that bitshifts and overflows in C can cause problems. e.g. shifting a signed negative integer is implementation-defined (https://stackoverflow.com/questions/28681131/right-shift-for-signed-integer)

Jakobeha:
For now we have `8` extra bits in the `Assumptions` class, so I'll make the assumptions larger and add back the 4th and 5th arg, see if that fixes the assert.

I could change `Assumptions` entirely, which will probably move `ConvertAssumptions` too. I initially was looking at creating a runtime `DynamicType` class, which would contain the assumptions of each arg type. But this would've made the assumptions take much more space, be slower, and just add a lot of complexity. I figured just adding some extra assumptions would be better.

o-:
> For now we have 8 extra bits in the Assumptions class, so I'll make the assumptions larger and add back the 4th and 5th arg, see if that fixes the assert.

As I and guido said, the problem is probably due to a mismatch between static and dynamic dispatch.

sorry for repeating, but please first identify the problem, before proposing a fix.

PR #425, opened 2019-04-05
fix performance regressions introduced in #417

o-:
some part of the scope analysis state can be queried at any position and only the actual bindings are flow-sensitive.

### DIFF
--- a/rir/src/compiler/analysis/scope.cpp
+++ b/rir/src/compiler/analysis/scope.cpp
@@ -8,44 +8,52 @@ namespace pir {
 
 ScopeAnalysis::ScopeAnalysis(ClosureVersion* cls, Promise* prom, Value* promEnv,
                              const ScopeAnalysisState& initialState,
-                             size_t depth, LogStream& log)
-    : StaticAnalysis("Scope", cls, prom, initialState, log), depth(depth),
-      staticClosureEnv(promEnv) {}
+                             ScopeAnalysisResults* globalState, size_t depth,
+                             LogStream& log)
+    : StaticAnalysis("Scope", cls, prom, initialState, globalState, log),
+      depth(depth), staticClosureEnv(promEnv) {}
 
-void ScopeAnalysis::lookupGlobal(const ScopeAnalysisState& state, Value* v,
-                                 const LoadMaybe& action,
-                                 const Maybe& notFound) const {
+void ScopeAnalysis::lookup(Value* v, const LoadMaybe& action,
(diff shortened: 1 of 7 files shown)


rir/src/compiler/analysis/scope.h
```diff
@@ -130,26 +130,49 @@ class ScopeAnalysis : public StaticAnalysis<
...
     // arguments and so on.
-    void lookup(const ScopeAnalysisState& envs, Value* i, const LoadMaybe&,
-                const Maybe& notFound = []() {}) const;
-    void lookup(const ScopeAnalysisState& envs, Value* i,
-                const ValueMaybe& action,
-                const Maybe& notFound = []() {}) const {
-        lookup(envs, i, [&](AbstractLoad load) { action(load.result); },
-               notFound);
+    //
+    // global version, can be called if the state does not correspond with i
```
charig: value?

rir/src/compiler/analysis/scope.cpp
```diff
@@ -12,48 +12,72 @@ ScopeAnalysis::ScopeAnalysis(ClosureVersion* cls, Promise* prom, Value* promEnv,
...
+    if (!instr)
+        return notFound();
 
-    // If this is a ldvarsuper, we perform a superget on the abstract
-    // environment and if possible recurse on the result
-    if (auto sld = LdVarSuper::Cast(instr)) {
-        action(superLoad(state, sld->varName, sld->env()));
-        return;
-    }
+    lookupGlobal(state, instr, action, [&]() {
```
charig: Lookup calls lookupGlobal. So...I can't find the semantics difference between calling `lookupAt` and `lookupGlobal` in terms of the correspondence between the state and the current instruction

o-: `lookupGlobal <= lookupAt`

something that is valiad at a particular position, is also valid globally, lookupAt also does lookupGlobal

charig: Ok, I got it now, the issue is that you are defining the local lookup in the `not found lambda` of the `lookupGlobal` call. I still do not get use to this code style. Just random comment addition: I know we discussed this since its introduction and I can not come to a better alternative but lookup is pretty confusing in this context.

charig (approved):

o-:
@charig can you take another look at my newest commit? I found another issue that was not fixed in #417, which caused more problems. So I had to slightly revise the strategy.

PR #426, opened 2019-04-05
added path fix to downloadBenchs.sh

Jakobeha:
Would set the path for simple benchmarks, also explicitly sets the vanilla R path, and sets other paths with more explicit locations


Jakobeha:
I cancelled the travis builds so they don't add to the queue

charig (approved):

PR #428, opened 2019-04-06
Safe force args in RIR -> RIR/PIR calls

Jakobeha:
This lets us infer assumptions for implicit calls made in RIR, so now we can infer simple arguments like constants at any callsite.

Modifies how `callImplicit`s are compiled, so that now a "safe forced" version of the compiled `SEXP` is pushed into the promise pool after each promise. Then the interpreter uses this to generate assumptions

### DIFF
--- a/rir/src/interpreter/safe_force.cpp
+++ b/rir/src/interpreter/safe_force.cpp
@@ -8,7 +8,7 @@
 
 namespace rir {
 
-static RIR_INLINE SEXP safeEval(SEXP e, SEXP rho) {
+SEXP safeEval(SEXP e, SEXP rho) {
     SEXPTYPE t = TYPEOF(e);
     if (t == LANGSXP || t == SYMSXP || t == PROMSXP || t == BCODESXP ||
         t == EXTERNALSXP) {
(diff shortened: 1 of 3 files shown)


o-:
there is a spimpler and more elegant way of achieving what you want, that is already mostly there. A call instruction (also the call_implicits) have a field of `availableAssumptions` (https://github.com/reactorlabs/rir/blob/master/rir/src/ir/BC_inc.h#L90). As you can see here (https://github.com/reactorlabs/rir/blob/master/rir/src/interpreter/interp.cpp#L2019) the dispatching starts with those assumptions as given. So you don't need to pre-evaluate promises or anything. The only missing piece is to statically populate those assumptions, which is currently not done yet. As you see in (https://github.com/reactorlabs/rir/blob/master/rir/src/ir/Compiler.cpp#L1004) the assumptions start out empty.

Jakobeha:
I applied the changes

o-:
looks good. waiting for the benchmark machine to become operational again before merging

o- (approved):

PR #429, opened 2019-04-08
Review416

o-:
address review on #416


mhyee:
Thanks, the comment about void contexts helps a lot!

PR #432, opened 2019-04-08
use the new run.sh script in rbenchmarking repo

o-:


.gitlab-ci.yml
```diff
@@ -78,7 +78,11 @@ test_flaky:
...
+    - git clone https://github.com/charig/ReBench.git -b envVarsSupport && cd ReBench
+    - pip install --user .
+    - export PATH=$HOME/.local/bin:$PATH
+    - cd ..
+    - git clone https://github.com/reactorlabs/rbenchmarking
     - cd build_release
-    - ../tools/downloadBenchs.sh ../build_release
-    - ../tools/runBenchs.sh --installRebench --travis
-    - PIR_DEOPT_CHAOS=1 ../tools/runBenchs.sh --travis
+    - ../rbenchmarking/run.sh bin ../external/custom-r/bin --warmup 1 e:RIR e:PIR -r -SN --iterations 2
```
charig: Why do we need warmup for testing?

charig (approved):
Modulo some small changes to RBenchmarking so that it also work on macOS

.gitlab-ci.yml
charig: RBenchmarking updated: here an example of the slightly different command from travis https://github.com/reactorlabs/RBenchmarking/blob/master/.travis.yml#L89

charig (approved):

.gitlab-ci.yml
o-: why is it in Setup?

charig: My original idea was that the setup scripts are not strictly necessary but auxiliary. You have a rebench.conf in the root dir and you can run it manually if you know where the implementations are. That said, I have no strong opinions about it, feel free to move it elsewhere.

PR #434, opened 2019-04-08
Removed verification that predecessor block isn't handled in phi

Jakobeha:
The above check won't be true if there's an input which enters a loop, which is a phi, but then it isn't used by the loop. Very rough example:

```r
y = 0
if (foo)
  x = a
else
  x = b
while (y < 10) {
  y = y + x
  ... # x doesn't get modified
}
...
```

At one point during rir_2_pir, the SSA looks like:

```
BB0:
  %0.0(y) = 0
  if foo, goto BB1, else BB2
BB1:
  %1.0(x) = a
  goto BB3
BB2:
  %2.0(x) = b
  goto BB3
BB3:
  %3.0(x) = Phi(%1.0, %2.0, %3.0)
  %3.1(y) = Phi(%0.0, %3.2)
  %3.2(y) = Add(%3.1, %3.0)
  ...
BB[N-1]:
  if (%3.2 < 10), goto BB3, else BB[N]
BB[N]
  ...
```  

All predecessor blocks are handled here. But the Phi contains an input to itself, so in rir_2_pir it gets removed, and the SSA becomes:

```
...
  %3.0(x) = Phi(%1.0, %2.0)
...
```

now the predecessor block [N-1] is no longer handled in %3.0, and when the slow verifier gets run, it sees this.

The problem might also be, we need another BB before the BB in the loop (BB3), which contains the single Phi %3.0.

This was something I got when implementing simple ranges with 1 loop. I had a stack value which would either increment or decrement the loop counter, and another which would be TRUE if the loop was going up and FALSE if it was going down. Both of these values were used inside the loop, but not modified, and both were assigned in separate branches which ran before the loop.

### DIFF
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -244,16 +244,6 @@ class TheVerifier {
                         inp.erase(pos);
                     }
                 });
-                if (!inp.empty()) {
-                    std::cerr << "Error at instruction '";
-                    i->print(std::cerr);
-                    std::cerr << " the following predecessor blocks are not "
-                                 "handled in phi: ";
-                    for (auto& in : inp)
-                        std::cerr << in->id << " ";
-                    std::cerr << "\n";
-                    ok = false;
-                }
             }
         }
 


o-:
hmm that's an interesting edge case

>  But the Phi contains an input to itself, so in rir_2_pir it gets removed, 

intuitively i would say that this is the error. where does it get removed?

the phi should know for every input what it's value is. and if for one input it is unchanged, then this should be indicated by the phi being it's own input. or am I wrong?

Jakobeha:
That makes sense. Actually the input doesn't get removed, it simply doesn't get added, at https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp#L96:

```cpp
if (in != p) {
  p->addInput(incomBB, in);
}
```

I tried removing this, and it causes a lot more problems. Also realized that `in != p` in more cases than just this one, and in all the other cases it seems to work fine. If I make it so the `Phi` is an input of itself *specifically* in the above case, by overriding the verifier to add it instead of raising "unhanded predecessor", then the example above works.

I know now that removing the check isn't the right solution, because if the predecessor isn't handled it messes up the stack. But I'm not sure whether we should 1) add a BB before the loop, to preserve that a phi can't be its own input, or 2) somehow handle this case, so that now (at least with what we've seen so far) it's the only case where a phi can be its own input.

mhyee:
I can imagine a similar situation like the following:

```
x_1 = 1
while (something) {
  x_2 = phi(x_1, x_3)
  x_3 = x_2 + 0
  # read x_3
  # something else
}
```

Here, `x_3 = x_2 + 0` could get optimized out, so the phi becomes `x_2 = phi(x_1, x_2)`. Now we have a phi inside the loop with itself as an input, and it's not possible to move it outside the loop.

But maybe this case isn't a problem, because the phi gets rewritten to `x_2 = phi(x_1)` and then ultimately removed because it's a single-input phi?

o-:
> it simply doesn't get added, at https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp#L96:

I see. so it only concerns stack values that we convert to phi's. what kind of errors do you get, when you remove that test? one of the problems will probably be, that you will create phis, which *only* have themselves as input, which is not well-founded.

it seems to me that the problem is rather the shape of the graph here. we have something like:

```
a       b   
  \   /   
    m <-. 
    |   |   
     \_/ 
```

if you would ensure that every loop starts with it's own block (which is kind of related to the preheader discussion we had earlier, then you would get:

```
a       b   
  \   /   
    m 
    |
    l <-.
    |   |   
     \_/ 

```

and the problem might just go away

mhyee:
What if your graph is:
```
         |
     --> L <--
   /     |     \
  |    /   \    |
  \ _ /     \ _ /
```

Preheader won't help you here.

I think the code would be something like:
```R
x <- 1
repeat {
  print(x)
  if (something) break
  if (something_else)
    x <- x + 0
  else
    x <- x + 1
}
```

I need to look at the output again, but simple ranges might end up producing something like this, with two loop "bodies."

o-:
good point. so maybe the solution really is to remove this `if (in != p) {` condition. and then add a cleanup pass that removes phis, which have only themselves as input

Jakobeha:
I tried removing the condition before and it caused some errors. Now I removed the condition and all the tests seem to work? Anyway, I'll try out the gnuR tests.

PR #435, opened 2019-04-09
Allow phis which take themselves as input

Jakobeha:
I also tried to ensure that cleanup would remove phis which take themselves and one other value, since they could just be replaced with the other value. I'm not sure if it fully works, but at least in this example:

```r
f <- rir.compile(function(a) {
  x <- a
  y <- 0
  while (y < 10) {
    y <- y + x
  }
  y
})
```

the variable x doesn't create a self-referencing phi. Also in this example:

```r
f <- rir.compile(function(a) {
  y <- 0
  if (a == 1)
    x <- 2
  else
    x <- 3
  while (y < 10)
    y <- y + x
  y
})
```

The variable x is merged in a BB before the loop, and that phi (not self-referencing) is reused in the loop.

### DIFF
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -187,9 +187,25 @@ class TheVerifier {
         if (auto phi = Phi::Cast(i)) {
             phi->eachArg([&](BB* input, Value* v) {
                 if (auto iv = Instruction::Cast(v)) {
-                    if (iv == phi) {
+                    if (slow && iv == phi && phi->nargs() == 2) {
                         // Note: can happen in a one-block loop, but only if it
                         // is not edge-split
+                        //
+                        // nargs == 2 because this is OK:
+                        // a   b
+                        //  \ /
+                        //   m<-.
+                        //   \_/
+                        // but this:
+                        //   a
+                        //   |
+                        //   m<-.
+                        //   \_/
+                        // should be cleaned up to
+                        //   a
+                        //   |
+                        //   /<-.
+                        //   \_/
                         std::cerr << "Error at instruction '";
                         i->print(std::cerr);
                         std::cerr << "': input '";
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -187,7 +187,7 @@ class TheVerifier {
         if (auto phi = Phi::Cast(i)) {
             phi->eachArg([&](BB* input, Value* v) {
                 if (auto iv = Instruction::Cast(v)) {
-                    if (input == phi->bb() && iv != phi) {
+                    if (input == phi->bb()) {
                         // Note: can happen in a one-block loop, but only if it
                         // is not edge-split
                         std::cerr << "Error at instruction '";
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -187,34 +187,7 @@ class TheVerifier {
         if (auto phi = Phi::Cast(i)) {
             phi->eachArg([&](BB* input, Value* v) {
                 if (auto iv = Instruction::Cast(v)) {
-                    if (slow && iv == phi && phi->nargs() == 2) {
-                        // Note: can happen in a one-block loop, but only if it
-                        // is not edge-split
-                        //
-                        // nargs == 2 because this is OK:
-                        // a   b
-                        //  \ /
-                        //   m<-.
-                        //   \_/
-                        // but this:
-                        //   a
-                        //   |
-                        //   m<-.
-                        //   \_/
-                        // should be cleaned up to
-                        //   a
-                        //   |
-                        //   /<-.
-                        //   \_/
-                        std::cerr << "Error at instruction '";
-                        i->print(std::cerr);
-                        std::cerr << "': input '";
-                        iv->printRef(std::cerr);
-                        std::cerr << "' phi has itself as input\n";
-                        ok = false;
-                    }
-
-                    if (input == phi->bb()) {
+                    if (input == phi->bb() && iv != phi) {
                         // Note: can happen in a one-block loop, but only if it
                         // is not edge-split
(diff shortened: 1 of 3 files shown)


o-:
> I'm not sure if it fully works

Then I would suggest you add testcases. I am confident you can figure it out!

One remark: the verifier is run after every pass. So with your strategy, the verifier will complain, before cleanup comes around. I would write an extra pass (not an actual compiler pass, but just a single iteration over the code after the initial rir2pir) to clean it up.

Also I am not convinced by the new verifier condition. Imho any phi which has less inputs than predecessors is broken.

Jakobeha:
I can "test" this by making the verifier fail on all self-referencing phis, and committing. If all the
tests pass (particularly test-debug), then it's probably fine. Currently we don't ever get a case where we need a self-referencing phi, that only happens with simple ranges. Looking at the code though, it really seems like the cleanup would remove any phis which reference themselves and one other value.

> The verifier is run after every pass. So with your strategy, the verifier will complain, before cleanup comes around

I make the cleanup use the "slow" verifier, which is currently only run at the end of all optimizations. However I wasn't sure if this was explicit. I could also not verify, because the cleanup should spot the exact same situation and fix it (although the whole point is if it doesn't).

> I would write an extra pass (not an actual compiler pass, but just a single iteration over the code after the initial rir2pir) to clean it up.

The problem is, what if there's a self-referencing phi with 3 inputs, but the other 2 the inputs get merged? Then the phi is self-referencing with 2 inputs. I could write the extra pass and then handle each case where a phi's inputs can get merged, but I think cleanup would be easier.

o-:
> I make the cleanup use the "slow" verifier, which is currently only run at the end of all optimizations. 

the idea of the verifier is, that code which does not pass verification is invalid and no pass is allowed to ever generate invalid code. I know we don't always run the slow verifier, but assume that the slow verifier must always pass after every pass. it is not ok to fix it up later in some other pass.

the reason is that some passes rely on properties that are verified. and we don't want to enforce a particular pass order for the whole thing to be correct.

> The problem is, what if there's a self-referencing phi with 3 inputs, but the other 2 the inputs get merged?

I see, good point.

UPDATE: Ignore this, I realized it's what you are doing....

I am still sort of confused about the whole thing. If I have this:

```
   a = 1
  loop:
   p = phi(a,p)
   print p
   goto loop
```

that should be valid pir. it's stupid, but it is valid.

you can clean it up, by just removing the phi and replacing it with a.

but I still don't see what your example would be, where you would have a legitimate case with a phi that has less inputs than predecessors.

o-:
actually looking at the code, you only weakened the predecessor rule for phi inputs. that seems reasonable. so this looks all good. feel free to merge, but remove the `//TODO` comment first.

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -187,18 +187,7 @@ class TheVerifier {
...
-                        std::cerr << "Error at instruction '";
-                        i->print(std::cerr);
-                        std::cerr << "': input '";
-                        iv->printRef(std::cerr);
-                        std::cerr << "' phi has itself as input\n";
-                        ok = false;
-                    }
-
-                    if (input == phi->bb()) {
+                    if (input == phi->bb() && iv != phi) {
```
o-: while I agree that a phi can be it's own input, a phi can never have it's own block as input. this would violate edge-split. so this should just be `if (input == phi->bb())`.

o-:
btw, I checked and llvm also produces phi's as their own input in the case described by @myhe.

Take this program:

```
    1 int main() {
    2   int x = 1;
    3   while (x) {
    4     while (x) {
    5       x = 0;
    6     }
    7   }
    8   return x;
    9 }
```

compile with `clang -O0 -emit-llvm t.c -c -S && opt -mem2reg -S t.ll`

and you get:

```
  br label %1

; <label>:1:                                      ; preds = %7, %0
  %.0 = phi i32 [ 1, %0 ], [ %.1, %7 ]
  %2 = icmp ne i32 %.0, 0
  br i1 %2, label %3, label %8

; <label>:3:                                      ; preds = %1
  br label %4

; <label>:4:                                      ; preds = %6, %3
  %.1 = phi i32 [ %.0, %3 ], [ 0, %6 ]
  %5 = icmp ne i32 %.1, 0
  br i1 %5, label %6, label %7

; <label>:6:                                      ; preds = %4
  br label %4

; <label>:7:                                      ; preds = %4
  br label %1

; <label>:8:                                      ; preds = %1
  ret i32 %.0
```

on the other hand we produce the following pir:

```
BB0
  real$'  %0.0  = LdConst          [1] 1 
  goto BB1
BB1
  real$'  %1.0  = Phi              %0.0:BB0, %4.0:BB6
  t       %1.1  = AsTest           %1.0
  void            Branch           %1.1 -> BB2 (if true) | BB3 (if false)
BB2
  goto BB4
BB3
  nil'    %3.0  = LdConst          NULL 
  void            Invisible        
  void            Return           %3.0
BB4
  real$'  %4.0  = Phi              %1.0:BB2, %9.0:BB9
  t       %4.1  = AsTest           %4.0
  void            Branch           %4.1 -> BB5 (if true) | BB6 (if false)
BB5
  void            Branch           %4.1 -> BB7 (if true) | BB8 (if false)
BB6
  goto BB1
BB7
  real$'  %7.0  = LdConst          [1] 0 
  goto BB9
BB8
  goto BB9
BB9
  real$'  %9.0  = Phi              %4.0:BB8, %7.0:BB7
  goto BB4
```

where we have a merge block at the end, that merges the conditional write, with the unchanged phi and then we get two recursive phis with each other as inputs.

So I actually wonder if we ourselves should somehow manage to merge the two phis? I wonder if this is exactly the case described on page 64 at http://ssabook.gforge.inria.fr/latest/book-v1.pdf ?

rir/src/compiler/analysis/verifier.cpp
Jakobeha: You're right, I got confused with the "input BB" vs. the BB where the input value is actually located

PR #436, opened 2019-04-09
Simple range uses 1 loop

Jakobeha:
This is rebased on https://github.com/reactorlabs/rir/pull/435

### DIFF
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -187,9 +187,25 @@ class TheVerifier {
         if (auto phi = Phi::Cast(i)) {
             phi->eachArg([&](BB* input, Value* v) {
                 if (auto iv = Instruction::Cast(v)) {
-                    if (iv == phi) {
+                    if (slow && iv == phi && phi->nargs() == 2) {
                         // Note: can happen in a one-block loop, but only if it
                         // is not edge-split
+                        //
+                        // nargs == 2 because this is OK:
+                        // a   b
+                        //  \ /
+                        //   m<-.
+                        //   \_/
+                        // but this:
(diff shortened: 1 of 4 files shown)


mhyee:
Thanks. Just to confirm, you're basically saving `delta` (+1 or -1) and `cond` (true or false) and then compiling the loop as:

```
while ( (i' < n') == cond ) {
  // body
  i' <- i' + delta
}
```

(Just a nitpick, but the comments describing this aren't super clear.)

Do you know if GNU R does this or something else? Also, was this originally blocked because of the whole phi self-input thing?

I would also be interested in how well the PIR optimizer handles this, especially for cases where `i` and `n` are constants.

Jakobeha:
Yes, I should fix the comments.

I think GNUR doesn’t rewrite for loops, it stores vectors 1:n as an altrep at runtime, so this is different.

This was blocked because of the self-input problem. In regular code, PIR has an extra BB with the incoming phi before the loop branch, so the phi doesn’t need to reference itself. It didn’t do this in simple ranges. I think because in “regular” code, there are initially a lot of bbs, and the variables in the loop are all loaded, so there aren’t any phis. Then, when pir merges bbs and lifts the variables into SSA, it generates the incoming bb in a separate phi, and doesn’t merge that with the loop bb. But the simple range variables are already on the stack, so pir generates the phi in rir2pir, and it’s already in the loop bb.

Right now PIR doesn’t really help with constantfolding. It doesn’t even remove the `== cond`. Maybe we can improve the constantfolding next, it doesn’t seem like it would be that hard to get rid of the condition, and replace the +1 or -1 with inc or dec.

mhyee (approved):
Looks good, just waiting for CI to finish. Thanks!

PR #438, opened 2019-04-10
re-enable asserts

o-:
we disabled asserts to speed up execution. In fact it has regressed. For
example I just measured in mandelbrot disabling asserts is a 7%
regression (!!!).

Since no one is investigating the cause, let us just enable them again,
as there are only downsides.


charig (approved):
Ok for me. But, is it necessary to remove the configuration? Wouldn't it be enough just to run the `releaseassert` version on the benchmarking machine? I still think it is useful to have build configurations for both settings

o-:
>  I still think it is useful to have build configurations for both settings

ok. I would like to have the default be the fast version. is it ok if I add `realeasenoassert`

charig:
Sure

PR #439, opened 2019-04-10
Less effects

Jakobeha:
- ~~Removed visibility flag for operations and subassigns. The instructions themselves don't seem to affect visibility. They're generated from functions which do affect visibility, but when we compile those functions in RIR we compile the instructions and then explicitly add a visibility instruction (or in the case of some instructions like binops, we compile the args and they should add the visibility).~~

- Removed warning and error effects from binops/unops whose arguments are a "safe" type. Most operations require both arguments to be numerics or numeric vectors. Comparison operations allow their arguments to be characters, strings, symbols, and quoted expressions like `quote(a + b)`. The quoted expressions seem little weird, because they're `code` values (I didn't allow `ast` values, not sure how those can be created), but expressions like `quote(a + f(b, c)) <= "foo` actually work (that example yields `TRUE`).

- Instructions which remove effects in `updateType` now print their effects.


Jakobeha:
For now I removed the visibility changes, so this only removes the errors / warnings and prints effects. This also means I had to disable tests, because the instructions aren't removed entirely, but I already know they get warning and error effects removed.

rir/src/compiler/pir/instruction.h
```diff
@@ -368,6 +393,7 @@ class InstructionImplementation : public Instruction {
     bool exits() const override final { return CF == Controlflow::Exit; }
     bool branches() const override final { return CF == Controlflow::Branch; }
     bool branchOrExit() const override final { return branches() || exits(); }
+    bool canRemoveEffects() const override { return false; }
```
o-: I don't quite understand this method. what does it say about the instruction?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -87,20 +92,107 @@ void Instruction::printGraphBranches(std::ostream& out, size_t bbId) const {
...
+        }
+
+        env()->printRef(out);
+
+        if (tty)
+            ConsoleColor::clear(out);
+    }
+}
+
+void Instruction::printEffects(std::ostream& out, bool tty) const {
```
o-: I have an alternative proposal: what about printing every effect on every instruction, but somehow abbreviate and group them into categories, such that it fits in 2-3 characters. For example:

`Any` => `*`

`Force | Reflection | LeakArg | ExecuteCode` => `#` 
`Force | ExecuteCode` => `>`

`Warn | Error  | Visibility` => `e`
`Warn | Visibility` => `w`
`Visibility` => `v`

Reads/writes/leaks env are already in the color of the env, they don't have to be printed.

ChangesContext, TriggerDeopt are specific to certain instructions and never removed, so I would not print them either.

rir/src/compiler/pir/instruction.h
```diff
@@ -193,6 +193,27 @@ class Instruction : public Value {
         }
     }
 
+    // Won't produce errors if all arguments are type argt (and no env)
+    // Won't produce warnings if all arguments are argt and scalar (and no env)
```
o-: update comment

rir/src/compiler/pir/instruction.h
```diff
@@ -193,6 +193,27 @@ class Instruction : public Value {
...
+    void maskWarnErrOn(PirType argt) {
+        bool noErr = true;
+        bool noWarn = true;
+        eachArg([&](Value* v) {
+            if (mayHaveEnv() && env() == v)
+                return;
+            if (!v->type.isA(argt)) {
+                noErr = false;
+                noWarn = false;
+            } else if (!v->type.isScalar()) {
```
o-: ok, this is strange. this is going to bite somebody, since it is so unexpected. why is this extra case needed? This seems to be highly specific to particular instructions. wouldn't it be better to just add this to the custom `updateType` implementation of the instructions that need it?

Jakobeha:
One thing I realized, the comment above the effects implies that an effect contains all lower-order effects, but we use an `EnumSet` to store all effects. Is it OK if I refactor this into `maxEffect`, or is the comment wrong?

rir/src/compiler/pir/instruction.cpp
Jakobeha: I can abbreviate them. Although I still think I should print out all effects (except maybe changescontext, triggerdeopt, read/write/leak env), because I can imagine getting a really confusing bug because an instruction has a strange set of effects.

rir/src/compiler/analysis/dead_store.h
```diff
@@ -178,7 +178,7 @@ class DeadStoreAnalysis {
                 }
             } else if (i->readsEnv()) {
                 observeFullEnv(i->env());
-            } else if (i->exits() || i->effects.contains(Effect::ExecuteCode)) {
+            } else if (i->exits() || i->maxEffect() >= Effect::ExecuteCode) {
```
o-: why? we don't treat the effects as ordered anymore. the effects are a set of atomar effects, so it does not make sense the compare them individually. It used to be that e.g. writesEnv would include readsEnv. then it used to make sense to compare individual effects. but now we have sets. so `readsEnv < readsEnv|writesEnv` but `readsEnv <> writesEnv`, they are incomparable.

rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -55,7 +55,6 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
                     });
                     next = ip + 1;
                     i->type.setNotObject();
-                    i->effects.reset(Effect::Reflection);
```
o-: ?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -62,7 +69,65 @@ void printPaddedTypeAndRef(std::ostream& out, const Instruction* i) {
...
+            ConsoleColor::red(out);
+        else if (eff.contains(Effect::LeaksEnv))
+            ConsoleColor::magenta(out);
+        else if (eff.contains(Effect::ReadsEnv))
+            ConsoleColor::yellow(out);
+    }
+    eff.reset(Effect::ReadsEnv);
+    eff.reset(Effect::WritesEnv);
+    eff.reset(Effect::LeaksEnv);
+    Effects allEff = Effects::Any();
```
o-: `static constexpr Effects allEff = Effects::Any() & ~(Effect::ReadsEnv | Effect::WritesEnv | Effect::LeaksEnv);`

rir/src/compiler/pir/instruction.h
```diff
@@ -81,9 +81,8 @@ struct InstrArg {
 enum class HasEnvSlot : uint8_t { Yes, No };
 
 // Effect that can be produced by an instruction.
-// This is a trivial lattice, any effect with higher order contains all the
-// lower order effects.
-enum class Effect : uint8_t {
+enum class Effect : int8_t {
```
o-: ups, thanks for removing stale comment

rir/src/compiler/pir/instruction.h
```diff
@@ -112,6 +111,11 @@ enum class Effect : uint8_t {
     FIRST = Visibility,
     LAST = ExecuteCode,
 };
+// Effects an instruction causes itself. These effects can lead to
+// lower effects, e.g. if an instruction executes more code, that can cause
```
o-: > These effects can lead to lower effects

no, this I would like to absolutely avoid! if an instruction has effect `executeCode` but not `warning`, this means, we know that the code it executes does not print any warning. I mean we anyway don't do such fancy things. But I would like to stress that we should always set all the effect flags that an instruction can have and never assume that one of them subsumes the other.

My suggestion for grouping effects while printing, was merely since some effects are just not as important if others are present. E.g. if we have Error, then we don't care that much if it can also warning. But nevertheless Error|Warning means it can either show a warning or throw an error (or neither), whereas Error means, just error, no warning.

rir/src/compiler/pir/instruction.h
```diff
@@ -166,15 +178,11 @@ class Instruction : public Value {
...
     bool mightChangeVisibility() const {
-        return effects.includes(Effect::Visibility);
-    }
-    bool mayUseReflection() const {
-        return effects.includes(Effect::Reflection);
-    }
-    bool mayForcePromises() const { return effects.includes(Effect::Force); }
-    bool leaksArg(Value* val) const {
-        return leaksEnv() || effects.includes(Effect::LeakArg);
+        return maxEffect() >= Effect::Visibility;
```
o-: see above. this should be `includes Visiblity`

rir/src/compiler/pir/instruction.h
```diff
@@ -166,15 +178,11 @@ class Instruction : public Value {
...
     }
+    bool mayUseReflection() const { return maxEffect() >= Effect::Reflection; }
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -166,15 +178,11 @@ class Instruction : public Value {
...
+    bool mayForcePromises() const { return maxEffect() >= Effect::Force; }
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -166,15 +178,11 @@ class Instruction : public Value {
...
+    bool leaksArg(Value* val) const { return maxEffect() >= Effect::LeakArg; }
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -208,15 +229,26 @@ class Instruction : public Value {
     }
 
     bool readsEnv() const {
-        return hasEnv() && effects.includes(Effect::ReadsEnv);
+        return hasEnv() && (effects.includes(Effect::ReadsEnv) ||
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -208,15 +229,26 @@ class Instruction : public Value {
...
+                            maxEffect() >= Effect::TriggerDeopt);
     }
     bool changesEnv() const {
-        return hasEnv() && effects.includes(Effect::WritesEnv);
+        return hasEnv() && (effects.includes(Effect::WritesEnv) ||
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -208,15 +229,26 @@ class Instruction : public Value {
...
+                            maxEffect() >= Effect::TriggerDeopt);
     }
     bool leaksEnv() const {
-        return hasEnv() && effects.includes(Effect::LeaksEnv);
+        return hasEnv() && (effects.includes(Effect::LeaksEnv) ||
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -1154,25 +1187,28 @@ SIMPLE_INSTRUCTIONS(V, _)
                   {{lhs, rhs}}, env, srcIdx) {}                                \
         void updateType() override final {                                     \
             maskEffectsAndTypeOnNonObjects(Type);                              \
+            maskEffect(PirType::SafeType().notObject(), Effect::Warn);         \
```
o-: this confused me a lot. Because i was searching for the SafeType function in PirType. Can you put this on the outside? Basically write `BINOP(Mul, lhs()->type | rhs()->type, PirType::num().notObject())` at the outside. Otherwise it is very confusing, especially given that the other `Type` thing is done in this way. Brevity is not always better, if it decreases readability

rir/src/compiler/pir/instruction.h
```diff
@@ -1198,13 +1234,16 @@ BINOP_NOENV(LOr, PirType::simpleScalarLogical());
                                              srcIdx) {}                        \
         void updateType() override final {                                     \
             maskEffectsAndTypeOnNonObjects(arg<0>().val()->type);              \
+            maskEffect(PirType::SafeType().notObject(), Effect::Warn);         \
```
o-: dito

rir/src/compiler/pir/instruction.h
```diff
@@ -152,6 +151,19 @@ class Instruction : public Value {
     }
 
   public:
+    void clearEffects() { effects.reset(); }
+    void clearVisibility() { effects.reset(Effect::Visibility); }
+    void clearLeaksEnv() { effects.reset(Effect::LeaksEnv); }
+    bool hasEffect() const { return !effects.empty(); }
+    bool hasVisibility() const { return effects.contains(Effect::Visibility); }
+    Effect maxEffect() const {
```
o-: remove, there is no such thing as a bigger/smaller effect anymore

rir/src/compiler/pir/instruction.h
```diff
@@ -152,6 +151,19 @@ class Instruction : public Value {
     }
 
   public:
+    void clearEffects() { effects.reset(); }
+    void clearVisibility() { effects.reset(Effect::Visibility); }
+    void clearLeaksEnv() { effects.reset(Effect::LeaksEnv); }
+    bool hasEffect() const { return !effects.empty(); }
+    bool hasVisibility() const { return effects.contains(Effect::Visibility); }
```
o-, line 153: same as mightChangeVisibility

rir/src/compiler/pir/instruction.h
```diff
@@ -152,6 +151,19 @@ class Instruction : public Value {
     }
 
   public:
+    void clearEffects() { effects.reset(); }
+    void clearVisibility() { effects.reset(Effect::Visibility); }
```
o-, line 150: `clearMightChangeVisibility`

rir/src/compiler/opt/elide_env.cpp
```diff
@@ -26,7 +26,6 @@ void ElideEnv::apply(RirCompiler&, ClosureVersion* function, LogStream&) const {
                     if (!envIsNeeded) {
                         i->elideEnv();
                         i->type.setNotObject();
-                        i->effects.reset(Effect::Reflection);
```
o-: ?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -156,10 +156,9 @@ class TheScopeResolution {
                     (CallInstruction::CastCall(i) || Force::Cast(i))) {
                     if (after.noReflection()) {
                         i->elideEnv();
-                        i->effects.reset(Effect::Reflection);
```
o-: ?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -62,7 +69,65 @@ void printPaddedTypeAndRef(std::ostream& out, const Instruction* i) {
     }
 }
 
-bool Instruction::validIn(Code* code) const { return bb()->owner == code; }
+void Instruction::printEffects(std::ostream& out, bool tty) const {
+    Effects eff = effects;
+    if (tty) {
+        if (eff.contains(Effect::WritesEnv))
```
o-: why is this here? It does not make sense to change the color of the effects. the idea is that the environment itself is printed in color, when it is changed.

rir/src/compiler/pir/instruction.cpp
```diff
@@ -62,7 +69,65 @@ void printPaddedTypeAndRef(std::ostream& out, const Instruction* i) {
...
+    allEff.reset(Effect::WritesEnv);
+    allEff.reset(Effect::LeaksEnv);
+    Effects allEffButRefl = allEff;
+    allEffButRefl.reset(Effect::Reflection);
+    if (eff == allEff) {
+        out << "!";
+    } else if (eff == allEffButRefl) {
+        out << "!/r";
+    } else if (eff == Effects::None()) {
+        if (effects != Effects::None()) // Has some env effects
```
o-: ?

o- (changes requested):
I would appreciate if you don't dump completely new changes on top of an already reviewed pr. this will just never let us converge and eats up a lot of time on the reviewers side, since I have to look through the same changes multiple times.

in short: effects are not ordered (any more). effect sets are ordered by set inclusion.

rir/src/compiler/pir/instruction.h
o-: btw. any particular reason to change to signed?

Jakobeha:
I was just confused about the effects, since the comment said they were ordered. I think we should add to the verifier some code to make sure an instruction can't have impossible effect patterns like `!Warning | ExecuteCode`, after this PR

rir/src/compiler/analysis/scope.cpp
```diff
@@ -372,7 +372,7 @@ AbstractResult ScopeAnalysis::doCompute(ScopeAnalysisState& state,
             }
         }
 
-        if (i->mayUseReflection()) {
+        if (i->effects.contains(Effect::Reflection)) {
```
charig, line 375: Why???

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -341,7 +341,7 @@ class ForceDominanceAnalysis : public StaticAnalysis<ForcedBy> {
                 }
             }
 
-            if (i->mayForcePromises())
+            if (i->effects.contains(Effect::Force))
```
charig, line 344: ???

PR #440, opened 2019-04-11
More constant fold

Jakobeha:
Rebased on less-effects

Added constant folding for `Eq` and `Neq` when one of the arguments is a simple logical and the other is a simple scalar, constant folding for `Not` when the arg is a simple scalar or double negation (although double negation technically wouldn't be a constant fold, should it go in cleanup?), and constant folding for `AsInt` when the arg is a simple real or logical.

These particularly help simple ranges. The latter allows the old version which normally has 2 loops to remove the second loop in cases like `for (i in 1:10)`, and the other optimizations remove some dead branches / simplify the current version which always uses 1 loop.


rir/src/compiler/pir/instruction.h
```diff
@@ -1022,25 +1045,27 @@ class FLIE(Extract1_1D, 3, Effects::Any()) {
     }
 };
 
-class FLIE(Extract2_1D, 3, Effects::Any()) {
+class FLIE(Extract2_1D, 3, Effects::Any() & ~Effects(Effect::Visibility)) {
```
o-: you seem to have mixed in stuff from abandoned commits from less-effects, stopping review here...

o- (commented):
I looks like you are treating `&&` and `||` as if they were commutative here. but they are not. `1||true` return `1` whereas `true||1` returns `true`

Jakobeha:
I don't know where. I treat `==` and `!=` as commutative, `&&` and `||` are compiled into `AsLogical -> Branch`.

o-:
got it

o-:
@Jakobeha hey, I am confused. You mentioned that this should benefit while loops. I looked at all the green fields in http://rflies.rir.o1o.ch/changes/?tre=10&rev=eb8572f06a6e2afd0295029a06b023a3639ffa0e&exe=1&env=1#  but if you check the timeline, I think they are only green, because the last run had an outlier.

especially in the micro benchmark examples I would have expected to see it, not? can you double check, if maybe the optimization is somehow not working?

Jakobeha:
They are definitely working, I think it's just that they're not effective. They only remove one or 2 instructions outside the loop. In examples like `for (i in 1:10) ...` they remove an entire branch, but those aren't in the benchmarks.

PR #442, opened 2019-04-11
get rid of unobserved visibility effects

o-:
this slight tweaking of the visibility optimization should actually get rid of most visibility effect flags on random instructions, by just removing all the ones which are shadowed by a later effect.

### DIFF
--- a/rir/src/compiler/analysis/visibility.cpp
+++ b/rir/src/compiler/analysis/visibility.cpp
@@ -30,9 +30,16 @@ AbstractResult VisibilityAnalysis::apply(LastVisibilityUpdate& vis,
 
         if (builtinUpdatesVisibility(builtinId)) {
             isVisibilityChanging();
+            break;
         }
-        break;
+    // fall through
     default:
+        // This instruction might change visibility, thus it's visibility effect
+        // might be observable. But it does not clear previous visibility
+        // instructions, because it might also not change it.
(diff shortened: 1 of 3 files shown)


rir/src/compiler/analysis/visibility.cpp
```diff
@@ -30,9 +30,16 @@ AbstractResult VisibilityAnalysis::apply(LastVisibilityUpdate& vis,
 
         if (builtinUpdatesVisibility(builtinId)) {
             isVisibilityChanging();
+            break;
         }
-        break;
+    // fall through
     default:
+        // This instruction might change visibility, thus it's visibility effect
```
Jakobeha, line 43: It might be simpler to just add `Visible` and `Invisible` to `observable` and not have `last`, I don't see where `last` is used any different than `observable`. Then you would also put this outside of the default case.

Also line 11, `if (vis.last != i) {`, wouldn't that always be false, or when would we update analysis on the same instruction twice?

Jakobeha (approved):
Looks good, I don't see anything wrong with semantics

rir/src/compiler/analysis/visibility.cpp
o-: thanks for the excellent suggestions

Jakobeha (approved):

PR #443, opened 2019-04-11
Disabled recompiling in static call

Jakobeha:
This might be the reason `pidgits` slowed down. I added this in my "scalar assumptions" PR but didn't mention it. I thought that it was a bug, because we had more assumptions but didn't seem like we were creating a closure for those extra assumptions. But it could explain why we recompiled a lot.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -2130,18 +2130,9 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             if (fun->invocationCount() % PIR_WARMUP == 0) {
                 Assumptions assumptions =
                     addDynamicAssumptionsForOneTarget(call, fun->signature());
-                if (assumptions != fun->signature().assumptions) {
+                if (assumptions != fun->signature().assumptions)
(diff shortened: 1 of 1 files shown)


o-:
@Jakobeha ok, but then the question is, why are we compiling a version with more assumptions, but then later we fail to dispatch to it? that still sounds like a bug to me...

rir/src/interpreter/interp.cpp
```diff
@@ -2130,18 +2130,9 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
             if (fun->invocationCount() % PIR_WARMUP == 0) {
                 Assumptions assumptions =
                     addDynamicAssumptionsForOneTarget(call, fun->signature());
-                if (assumptions != fun->signature().assumptions) {
+                if (assumptions != fun->signature().assumptions)
                     // We have more assumptions available, let's recompile
                     dispatchFail = true;
-
```
charig, line 2136: Sorry for the late comment. But I do not understand this change. This does not seem like it was a bug for me but a design decision. What is the precise reason we are taking that decision down?

PR #444, opened 2019-04-11
Fix Visibility in RIR

Jakobeha:
Adds explicit `R_Visible` to binop, extract, and subassign fast cases


rir/src/interpreter/interp.cpp
```diff
@@ -3255,6 +3261,8 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
                         ostack_popn(ctx, 4);
 
                         ostack_push(ctx, mtx);
+                        // This is intentional - subassign enables visibility
+                        R_Visible = (Rboolean) true;
```
o-: 1. the ir/Compiler.cpp emits an explicit visibility change
2. the ir/Compiler.cpp disagrees and says it sets it to invisible

PR #445, opened 2019-04-11
Use fixed binop visibility to remove/replace instructions

Jakobeha:
- Treats instructions which set visibility but aren't `Visible` / `Invisible`, like binops, like `Visible` and `Invisible` in the analysis, so they can remove extra visibility checks.
- Replaces those instructions, if they're unused, with `Visible` or `Invisible`

This is rebased off of `less-effects`. I haven't really tested if this removes too much visibility instructions


o-:
> This is rebased off of less-effects. I haven't really tested if this removes too much visibility instructions

sorry, but I can only review this in isolation if there is a useful diff between less-effects and this branch. for example 1-3 well formulated commits that i can look at in isolation. here the two things are just intermingled.

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -28,10 +28,20 @@ class TheCleanup {
...
                 auto next = ip + 1;
                 bool removed = false;
-                if (!i->branchOrExit() && !i->hasObservableEffects() &&
-                    i->unused()) {
+                if (i->unused() && !i->branchOrExit() &&
+                    !i->hasObservableEffects()) {
                     removed = true;
                     next = bb->remove(ip);
+                } else if (i->unused() && !i->branchOrExit() &&
+                           !i->hasImpureEffects() && i->alwaysSetsVisible() &&
```
o-: please make it explicit what you are testing here. I don't think `hasImpureEffects` is what you want here. Yes, it accidentially does what you want, but I see no reason for this function to keep doing what you want in the future... as soon as we add more effects and change that function, this code will break.

also a lot of code duplication here...

so how about the following:
```
bool dead = i->unused() && !i->branchOrExit);
if (dead && !i->hasObservableEffects()) {
  ..
} else if (dead && i->getObservableEffects() == Effect::Visible && i->alwaysOverridesVisible()) {
    if (!Visible::Cast(i) && !Invisible::Cast(i))
        if (i->alwaysSetsVisible) {
            replace with visible
        } else {
            replace with invisible
    }
} ...

File: rir/src/compiler/pir/instruction.cpp [hunk 2]
```diff
@@ -510,6 +510,22 @@ Instruction* BuiltinCallFactory::New(Value* callerEnv, SEXP builtin,
         return new CallBuiltin(callerEnv, builtin, args, srcIdx);
 }
 
+VisibilityMod CallBuiltin::visibilityMod() const {
+    if (builtinUpdatesVisibility(builtinId)) {
+        return VisibilityMod::AlwaysOverrides;
```
o-: please set it to yes or no. no need for the third category

rir/src/compiler/pir/instruction.h
```diff
@@ -119,6 +119,14 @@ enum class Controlflow : uint8_t {
     Branch,
 };
 
+// How an instruction modifies visibility
+enum class VisibilityMod : uint8_t {
```
o-: Those names are really hard to read. Also they repeat and I don't understand why we need the 4th option. How about this:

```
enum class VisibilityFlag : uint8_t {
  Set,
  Clear,
  Unknown
};
```

also, maybe make this an inner class of Instruction

Jakobeha: We have a lot of other types which are outside of instruction, so I don't think this should be an inner class unless we move all those types inside too?

o-: fair enough. But `VisibilityFlag::Visible` is not a good name. It does not make sense to use scoped enum, and then repeat the scope name again. That's like `Door::doorOpen`, when you can say `Door::open`

Jakobeha: ok, fixed (should've put in the commit message)

rir/src/compiler/pir/instruction.cpp
```diff
@@ -510,6 +510,14 @@ Instruction* BuiltinCallFactory::New(Value* callerEnv, SEXP builtin,
         return new CallBuiltin(callerEnv, builtin, args, srcIdx);
 }
 
+VisibilityFlag CallBuiltin::visibilityFlag() const {
+    return (VisibilityFlag)getFlag(builtinId);
```
o-: that is way too clever.

the dumber the better to read.

```
if (flag == 1)
  return VisibilityFlag::Off;
if (flag == 0)
  return VisibiliyFlag::On;
return VisibilityFlag::Unknown;
```

is so much easier to read an 100x more robust

rir/src/compiler/pir/instruction.h
```diff
@@ -119,6 +119,14 @@ enum class Controlflow : uint8_t {
     Branch,
 };
 
+// How an instruction modifies visibility
+// The order is important - this is equivalent to R's visibility flag
```
o-: refactor CallBuiltin::visibility flag, such that this is not needed

rir/src/compiler/pir/instruction.h
```diff
@@ -1503,6 +1546,8 @@ class VLIE(CallBuiltin, Effects::Any()), public CallInstruction {
         return hash_combine(InstructionImplementation::gvnBase(), blt);
     }
 
+    VisibilityFlag visibilityFlag() const override;
```
o-, line 1548: how about CallSafeBuiltin?

o- (commented):
some final comments, after that its ok to merge

rir/src/compiler/pir/instruction.h
Jakobeha: ? it's already handled

o-: I'm blind :)

PR #447, opened 2019-04-12
No longer optimizes modules unless a new closure is compiled

Jakobeha:
Sometimes when we "compile" a closure we actually just reuse a previously compiled one. But `optimizeModule` was still being run. This PR disables it, and also logs that the closure was reused.

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -222,8 +222,13 @@ SEXP pirCompile(SEXP what, const Assumptions& assumptions,
     logger.title("Compiling " + name);
     pir::Rir2PirCompiler cmp(m, logger);
     cmp.compileClosure(what, name, assumptions,
-                       [&](pir::ClosureVersion* c) {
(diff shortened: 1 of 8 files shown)


rir/src/api.cpp
```diff
@@ -222,8 +222,13 @@ SEXP pirCompile(SEXP what, const Assumptions& assumptions,
     logger.title("Compiling " + name);
     pir::Rir2PirCompiler cmp(m, logger);
     cmp.compileClosure(what, name, assumptions,
-                       [&](pir::ClosureVersion* c) {
+                       [&](pir::ClosureVersion* c, bool isNew) {
                            logger.flush();
+                           if (!isNew) {
+                               logger.debug("Reused already compiled version");
```
o-, line 228: if you put an `assert(false)` here, then we run through the whole test suite without triggering it. So this pr just adds dead code. we don't keep modules between compilations, so I am not sure how you think this could be triggered.

PR #448, opened 2019-04-14
if a call instruction was never called, remove it

o-:
this prevents a dead call instruction (with no call target feedback)
from poisoning the analysis, since a generic call is emited.

### DIFF
--- a/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
+++ b/rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
@@ -322,20 +322,22 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
             auto& feedback = callTargetFeedback.at(callee);
             // If this call was never executed. Might as well compile an
             // unconditional deopt
-            if (!inPromise() && srcCode->funInvocationCount > 0 &&
+            if (!inPromise() && srcCode->funInvocationCount > 1 &&
                 feedback.taken == 0) {
-                // Hack to record the deoptimized call, to ensure we compile
-                // something better on re-optimization.
-                auto deoptPos = pos - 1 - sizeof(ObservedCallees);
-                if (*deoptPos != Opcode::record_call_)
-                    deoptPos = pos;
-                auto fs = insert.registerFrameState(srcCode, deoptPos, stack);
-                insert(new Deopt(fs));
-                stack.clear();
-                break;
+                // To avoid deoptimization loops, we must ensure that on
+                // deoptimization we actually record the new call target.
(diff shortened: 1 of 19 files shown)


rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -1043,6 +1058,19 @@ Value* Rir2Pir::tryTranslate(rir::Code* srcCode, Builder& insert) const {
                 log.failed("Abort r2p due to unsupported bc");
                 return nullptr;
             }
+
+            if (!inPromise() && !insert.getCurrentBB()->isEmpty()) {
+                auto last = insert.getCurrentBB()->last();
+
+                if (Deopt::Cast(last)) {
```
charig, line 1067: Isn't the break above on line 335 enough?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -320,6 +320,21 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
         // currently it does not pay off to put any deopts in there.
         if (callTargetFeedback.count(callee)) {
             auto& feedback = callTargetFeedback.at(callee);
+            // If this call was never executed. Might as well compile an
+            // unconditional deopt
+            if (!inPromise() && srcCode->funInvocationCount > 0 &&
+                feedback.taken == 0) {
+                // Hack to record the deoptimized call, to ensure we compile
```
charig: Are we using this hack anywhere?

charig (approved):

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: yes, without this for example binary tree which has something like:
```
tree = function(n, ...)
  if (n == 1)
    c(...)
  else
    tree(n-1, ...)
```
will optimize before the then-branch is executed. it deoptimizes on the then branch, but before it every ends up in another leaf branch, it will have optimized the function again. thus we will never record the `c` call and will always deoptimize there. the hack ensures that the `record_call` instruction is executed on deoptimization, thus on the next optimization the code will be correct.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
o-: no, break just leaves the compileBC  function. we also must abort all later instructions to avoid compiling dead code.

o-:
@charig you can also wait with reviewing until travis is green to avoid unnecessary work. i am still not sure if this version will now pass...

PR #450, opened 2019-04-15
Verify effects

Jakobeha:
Verifies that instructions don't have impossible combinations of effects.

I wasn't exactly sure what combinations are "impossible", currently I've implemented:
- If an instruction executes code, it must have all other effects *except* affecting the environment (or `LeakArg`)
- If an instruction has `LeakEnv` it must have `LeakArg`

I also fixed up some of the analyses to either remove or check for the effects, so that the verification passes. Again, I'm not sure exactly what effects imply what, or if the assumptions I make are valid, so don't overlook them when reviewing.


Jakobeha:
travis is giving "custom-R repo is dirty" here, although I didn't modify the repo

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -156,10 +156,13 @@ class TheScopeResolution {
                     (CallInstruction::CastCall(i) || Force::Cast(i))) {
                     if (after.noReflection()) {
                         i->elideEnv();
+                        // TODO: Is this OK?
```
o-, line 159: no. why would it. the test is for no reflection. of course a call executes code...

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -156,10 +156,13 @@ class TheScopeResolution {
...
+                        i->effects.reset(Effect::ExecuteCode);
                         i->effects.reset(Effect::Reflection);
                     }
                     if (after.envNotEscaped(i->env())) {
                         i->effects.reset(Effect::LeaksEnv);
+                        i->effects.reset(Effect::LeakArg);
```
o-, line 165: not correct either, the test says nothing about the other arguments

rir/src/compiler/opt/visibility.cpp
```diff
@@ -32,7 +32,8 @@ void OptimizeVisibility::apply(RirCompiler&, ClosureVersion* function,
                     next = bb->remove(ip);
                 }
             } else if (instr->effects.contains(Effect::Visibility)) {
-                if (!visible.observed(instr)) {
+                if (!visible.observed(instr) &&
+                    !instr->effects.contains(Effect::Reflection)) {
```
o-, line 36: why do you think this is necessary?

the visible effect is not about if it does change visibility, it's if the change in visibility is observable.

rir/src/compiler/analysis/verifier.cpp
```diff
@@ -309,6 +309,32 @@ class TheVerifier {
...
+            ok = false;
+        };
+
+        if (eff.contains(Effect::ExecuteCode) && effAndEnv != Effects::Any()) {
+            fail("ExecuteCode implies all other effects except env-based");
+        }
+
+        if (eff.contains(Effect::LeaksEnv) && !eff.contains(Effect::LeakArg)) {
+            fail("LeaksEnv implies LeakArg");
+        }
```
o-, line 337: honestly I don't agree with either of the verifications.

e.g. we could have a call, that we know the target. so the call instruction would have as effects executeCode + whatever the callee does. but that does not have to be Any.

then I am not sure leakArg includes the environment. I guess we are not heavily using it yet. but I can see that we might want leakArg to mean leak some argument other than the env.

rir/src/compiler/opt/elide_env.cpp
```diff
@@ -26,6 +26,7 @@ void ElideEnv::apply(RirCompiler&, ClosureVersion* function, LogStream&) const {
                     if (!envIsNeeded) {
                         i->elideEnv();
                         i->type.setNotObject();
+                        i->effects.reset(Effect::ExecuteCode);
```
o-, line 29: this looks good, but if we do that , we should probably rename the `envOnlyForObj` instruction. because right now the function name only suggests that no env is needed. someone might be tempted to set this to true for an instruction that still executes code, but just doesn't need an env...

PR #451, opened 2019-04-15
fix pir's handling of missing args

o-:
accessing a missing arg throws an error. In pir, we replace ldvar with
ldvar_noforce and force. the former is responsible for checking
missingness. then in scope resolution we replace ldvar with the actual
value. this replacement "forgets" the missingness check. we fix this
with an explicit chkMissing instruction.


o-:
bugfix

PR #453, opened 2019-04-17
support calling closure from R with extra method dispatch arguments

o-:
this should fix the issue where innerBenchmarkLoop in harnes.r is never compiled, since it is called trough UseMethod


o-:
tbr

PR #454, opened 2019-04-17
Improved pre-push hook

Jakobeha:
Small fixes so I don't push unfinished/broken stuff
- Won't push if there are unresolved commits
- Won't push if the build fails

### DIFF
--- a/tools/pre-push.d/90-tests.hook
+++ b/tools/pre-push.d/90-tests.hook
@@ -2,9 +2,9 @@
 
 ROOT=$1
 
-DIFF=$(git diff)
-if [ -n "$DIFF" ]; then
-  echo "please commit before pushing."
+git diff --exit-code > /dev/null && git diff --cached --exit-code > /dev/null
+if test "$?" -ne 0; then
+  echo "please commit or stash your changes before pushing."
   exit 1
 fi
 
--- a/tools/pre-push.d/90-tests.hook
+++ b/tools/pre-push.d/90-tests.hook
@@ -2,9 +2,20 @@
 
 ROOT=$1
 
+DIFF=$(git diff)
+if [ -n "$DIFF" ]; then
+  echo "please commit before pushing."
+  exit 1
+fi
+
 . "${ROOT}/tools/script_include.sh"
 
 build ${ROOT}/tests/build_push_hook ${ROOT} release
+
+if test "$?" -ne 0; then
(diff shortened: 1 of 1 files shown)


tools/pre-push.d/90-tests.hook
```diff
@@ -2,10 +2,21 @@
 
 ROOT=$1
 
+DIFF=$(git diff)
```
mhyee: You can run `git diff --exit-code` which returns `1` if there were any changes and `0` otherwise.

However, you should also check for staged but uncommitted files (i.e. you ran `git add` but not `git commit`): `git diff --cached --exit-code`

Source: https://stackoverflow.com/a/5139672

tools/pre-push.d/90-tests.hook
```diff
@@ -2,10 +2,21 @@
...
+if [ -n "$DIFF" ]; then
+  echo "please commit before pushing."
```
mhyee: Minor wording tweak: "please commit or stash your changes before pushing."

mhyee:
👍 I've definitely pushed broken things because my changes didn't build and the tests ran on the last good build...

What's with the submodule update, why are they included here? If this was a mistake, please also cancel the builds, since the pre-push hook is unrelated to CI.

charig:
Many times I have changes that I do not want to push nor checkout. Sometimes I even have changes in the submodules that I do not want to rollback. So, this is very inconvenient for me. Unless it is extremely useful for others I would suggest to rollback.

o-:
@charig I am actually with you on this

mhyee:
`git stash`? Or even `git push --no-verify`, since that is basically what happens if you had uncommited changes before this PR.

o-:
can't stash dirty submodule and no-verify also skips the tests.

o-:
how about #459

mhyee:
> no-verify also skips the tests.

But if you have uncommitted changes, then the tests are running on those changes, not what you're about to push.

#459 is probably the right compromise, then.

o-:
so currently my situation is that i have a gnur submodule change that removes an `attribute_hidden` for my native branch. so I would have to `git submodule update` whenever i switch back and forth. not a big issue, but also a bit silly given that i am not gonna recompile the module anyway, since on master it works just fine with the extra symbol :)

charig:
Yes, #459 seems to be a good compromise

PR #455, opened 2019-04-17
add an assumption for non reflective arguments

o-:
1. tag all promises as no reflective, if we can rule out that they are reflective
2. set a context assumption if all arguments are non-reflective
3. remove dependency on local environment when forcing arguments under the nonReflectiveArguments assumption

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -69,12 +69,17 @@ pir.check <- function(f, ..., warmup=NULL) {
         as.pairlist(lapply(lapply(as.list(substitute(...())), as.character), as.name))
     if (length(checks) == 0)
         stop("pir.check: needs at least 1 check")
+
+    .Call("pir_check_warmup_begin")
     rir.compile(f)
-    if (is.list(warmup)) {
+    if (!is.null(warmup)) {
+        rir.compile(warmup)
+        pir.compile(warmup)
         for (i in 1:as.numeric(Sys.getenv("PIR_WARMUP", unset="3")))
-            do.call(f, warmup)
+            warmup(f)
     }
     .Call("pir_check", f, checks)
+    #.Call("pir_check_warmup_end")
 }
 
 # creates a bitset with pir debug options
--- a/rir/src/compiler/opt/eager_calls.cpp
+++ b/rir/src/compiler/opt/eager_calls.cpp
@@ -152,6 +152,33 @@ void EagerCalls::apply(RirCompiler& cmp, ClosureVersion* closure,
                     continue;
                 }
 
+                auto availableAssumptions = call->inferAvailableAssumptions();
+
+                // We picked up more assumptions, let's compile a better
+                // version. Maybe we should limit this at some point, to avoid
+                // version explosion.
+                if (availableAssumptions.includes(
+                        Assumption::NoReflectiveArgument) &&
+                    !version->assumptions().includes(
+                        Assumption::NoReflectiveArgument)) {
+                    auto newVersion = cls->cloneWithAssumptions(
+                        version, availableAssumptions,
+                        [&](ClosureVersion* newCls) {
+                            Visitor::run(newCls->entry, [&](Instruction* i) {
+                                if (auto f = Force::Cast(i)) {
+                                    if (LdArg::Cast(f)) {
+                                        f->elideEnv();
+                                        f->effects.reset();
+                                    }
+                                }
+                            });
+                        });
+                    call->hint = newVersion;
+                    assert(call->tryDispatch() == newVersion);
+                    ip = next;
+                    continue;
+                }
+
                 bool allEager = version->properties.includes(
                     ClosureVersion::Property::IsEager);
                 if (!allEager &&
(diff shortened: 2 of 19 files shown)


o-:
tbr, if you find time. this commit should shift the game slightly in our favour, since finally the iterations promise in the benchmarks will not taint the local env anymore. thus I would like everybody to be able to build on this version.

rir/src/compiler/opt/eager_calls.cpp
```diff
@@ -152,6 +152,33 @@ void EagerCalls::apply(RirCompiler& cmp, ClosureVersion* closure,
...
+                    !version->assumptions().includes(
+                        Assumption::NoReflectiveArgument)) {
+                    auto newVersion = cls->cloneWithAssumptions(
+                        version, availableAssumptions,
+                        [&](ClosureVersion* newCls) {
+                            Visitor::run(newCls->entry, [&](Instruction* i) {
+                                if (auto f = Force::Cast(i)) {
+                                    if (LdArg::Cast(f)) {
+                                        f->elideEnv();
+                                        f->effects.reset();
```
o-, line 171: note to self, this is bogus, should only reset reflection. will fix next week

rir/src/runtime/Assumptions.h
```diff
@@ -136,12 +138,14 @@ struct Assumptions {
...
+
     uint8_t unused = 0;
+    uint16_t unused2 = 0;
 };
 #pragma pack(pop)
 
 typedef uint32_t Immediate;
-static_assert(sizeof(Assumptions) == sizeof(Immediate),
+static_assert(sizeof(Assumptions) == 2 * sizeof(Immediate),
               "Assumptions needs to be one immediate arg size");
```
mhyee, line 155: Forgot to update this message.

rir/src/compiler/util/ConvertAssumptions.cpp
```diff
@@ -22,15 +22,17 @@ void readArgTypeFromAssumptions(const Assumptions& assumptions, PirType& type,
...
-    Value* value = arg->followCastsAndForce();
-    if (!MkArg::Cast(value)) {
+    if (auto mk = MkArg::Cast(arg)) {
+        if (mk->isEager()) {
+            if (mk->eagerArg() == MissingArg::instance())
+                assumptions.remove(Assumption::NoExplicitlyMissingArgs);
+            else
+                assumptions.setEager(i);
+        }
+        // if (!mk->noReflection)
```
mhyee, line 32: Should this be deleted? Why is it commented out?

rir/src/compiler/pir/closure.cpp
```diff
@@ -54,7 +54,7 @@ ClosureVersion* Closure::cloneWithAssumptions(ClosureVersion* version,
 ClosureVersion*
 Closure::findCompatibleVersion(const OptimizationContext& ctx) const {
     // ordered by number of assumptions
-    for (auto c = versions.begin(); c != versions.end(); c++) {
+    for (auto c = versions.rbegin(); c != versions.rend(); c++) {
         auto candidate = *c;
```
mhyee, line 58: I'm noticing now that these are probably making unnecessary copies. The copies are small, but still...

So you'll want `const auto& candidate = *c;` and `const auto& candidateCtx = candidate.first;`.

rir/src/runtime/Assumptions.h
```diff
@@ -113,16 +113,14 @@ struct Assumptions {
...
                 return true;
+            if (other.subtype(*this))
+                return false;
         }
 
-        // for std::map we need a complete order, subtype is only partial
+        // we need a complete order, subtype is only partial
         if (missing != other.missing)
-            return missing > other.missing;
-        if (flags.count() != other.flags.count())
```
mhyee, line 114: Why is this removed? Is the comment above (order by number of assumptions) still valid?

rir/src/runtime/DispatchTable.h
```diff
@@ -94,10 +107,19 @@ struct DispatchTable
         }
         std::cout << "\n";
 #endif
-        SLOWASSERT(contains(fun->signature().assumptions));
+
+#ifdef DEBUG_DISPATCH
```
mhyee, line 111: Combine with the `#ifdef` above?

rir/src/runtime/DispatchTable.h
```diff
@@ -67,18 +67,31 @@ struct DispatchTable
...
+            std::cout << "Tried to insert into a full Dispatch table. Have: \n";
+            for (size_t i = 0; i < size(); ++i) {
+                auto e = getEntry(i);
+                std::cout << "* "
+                          << Function::unpack(e)->signature().assumptions
+                          << "\n";
+            }
+            std::cout << "\n";
+            std::cout << "Tried to insert: " << assumptions << "\n";
+            Rf_error("failed");
```
mhyee, line 91: When should we start thinking about growing the dtable?

rir/R/rir.R
```diff
@@ -69,12 +69,17 @@ pir.check <- function(f, ..., warmup=NULL) {
...
-    if (is.list(warmup)) {
+    if (!is.null(warmup)) {
+        rir.compile(warmup)
+        pir.compile(warmup)
         for (i in 1:as.numeric(Sys.getenv("PIR_WARMUP", unset="3")))
-            do.call(f, warmup)
+            warmup(f)
     }
     .Call("pir_check", f, checks)
+    #.Call("pir_check_warmup_end")
```
mhyee, line 82: Why is this commented out?

rir/src/runtime/DispatchTable.h
o-: for now I kind of like it when we get notified that we hit the current limit in the tests, because usually it points to some sort of error, rather than real version explosion...

PR #458, opened 2019-04-18
compile while(...) {} as if(...) do {} while (...)

o-:
this is specifically to give us a preheader to force promises used in
the loop condition. E.g. in:

    function(x)
        ...
        while (i < x) { ... }

our optimizations fail, because `x` is forced in a loop. But if we write
the equivalent

    function(x)
        ...
        if (i < x)
          do { ... } while (i < x)

It's very easy to see where x is forced and also a lot simpler to inline
that promise and to determine that the second x is not forcing at all.


o-:
@mhyee I merged this tbr. I guess the main question is, does it help. and we will see it on the benchmarks :) (also it might only help after #455 )

PR #459, opened 2019-04-18
don't abort push on dirty repo

o-:


mhyee:
I was going to suggest a Y/N prompt, but hey, we can always Ctrl+C.

PR #461, opened 2019-04-21
Vector containing objects fix

Jakobeha:
A vector containing objects isn't an object itself. However, it you extract an element that could be an object.

Currently we don't see any failures because PIR can't ever infer a vector containing objects isn't an object, e.g. `c(a, b, c)` will be marked as an object because PIR doesn't know the type signature of `c`, and there aren't any literals for object vectors.

Also, builtins (currently only `lengths`), do extract elements on their arguments, so we can't mark them as safe unless we know the arguments aren't object-containing vectors.

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -995,8 +995,6 @@ class FLIE(Subassign1_1D, 4, Effects::Any()) {
     Value* idx() { return arg(2).val(); }
     void updateType() override final {
         maskEffectsAndTypeOnNonObjects(lhs()->type | rhs()->type);
-        if (!lhs()->type.isA(PirType::num() | RType::str))
-            type = type.orObject();
     }
 };
 
@@ -1064,9 +1062,9 @@ class FLIE(Extract1_1D, 3, Effects::Any()) {
         auto t = vec()->type;
         if (idx()->type.isScalar())
             t.setScalar();
+        if (PirType(RType::vec).isA(t))
+            t = t.orObject();
         maskEffectsAndTypeOnNonObjects(t);
-        if (!t.isA(PirType::num() | RType::str))
-            type = type.orObject();
     }
 };
 
@@ -1079,9 +1077,10 @@ class FLIE(Extract2_1D, 3, Effects::Any()) {
     Value* vec() { return arg(0).val(); }
     Value* idx() { return arg(1).val(); }
     void updateType() override final {
-        maskEffectsAndTypeOnNonObjects(vec()->type.scalar());
-        if (!vec()->type.isA(PirType::num() | RType::str))
-            type = type.orObject();
+        auto t = vec()->type.scalar();
+        if (PirType(RType::vec).isA(t))
+            t = t.orObject();
+        maskEffectsAndTypeOnNonObjects(t);
     }
 };
 
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -1059,10 +1059,7 @@ class FLIE(Extract1_1D, 3, Effects::Any()) {
     Value* vec() { return arg(0).val(); }
     Value* idx() { return arg(1).val(); }
     void updateType() override final {
-        PirType t = vec()->type.elem();
-        if (!idx()->type.isScalar())
-            t = t.orNotScalar();
-        maskEffectsAndTypeOnNonObjects(t);
+        maskEffectsAndTypeOnNonObjects(vec()->type.subsetType(idx()->type));
     }
 };
 
@@ -1075,7 +1072,7 @@ class FLIE(Extract2_1D, 3, Effects::Any()) {
     Value* vec() { return arg(0).val(); }
     Value* idx() { return arg(1).val(); }
     void updateType() override final {
-        maskEffectsAndTypeOnNonObjects(vec()->type.elem());
+        maskEffectsAndTypeOnNonObjects(vec()->type.extractType(idx()->type));
     }
 };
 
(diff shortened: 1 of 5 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -995,6 +995,8 @@ class FLIE(Subassign1_1D, 4, Effects::Any()) {
     Value* idx() { return arg(2).val(); }
     void updateType() override final {
         maskEffectsAndTypeOnNonObjects(lhs()->type | rhs()->type);
+        if (!lhs()->type.isA(PirType::num() | RType::str))
+            type = type.orObject();
```
o-: ugh, that is not nice. first overshooting and then correcting after the fact, with some partial list of exceptions. please fix it at the source. for example
```
if (! PirType(RType::vector)::isA(lhs()->type))
  maskEffectsAndTypeOnNonObjects ...
```
sounds reasonable to me?

rir/src/compiler/pir/type.cpp
```diff
@@ -80,7 +80,7 @@ void PirType::merge(SEXPTYPE sexptype) {
 PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
     merge(TYPEOF(e));
 
-    if (!Rf_isObject(e)) {
+    if (!isObject(e)) {
```
o-: why

o- (changes requested):
nice catch. can you please add a regression test for both problems you fixed?

Jakobeha:
I can’t add a regression test because the problem is, I don’t think PIR has  a way to assume that vectors containing objects aren’t objects. The only way to create these vectors is with a function like `c`, which PIR can’t infer doesn’t create an argument (especially since the args are objects). I only caught this issue because I added extra inference by safe forcing variables, in another branch.

rir/src/compiler/pir/instruction.h
```diff
@@ -1083,11 +1092,16 @@ class FLIE(Extract1_2D, 4, Effects::Any()) {
...
+    Value* idx1() { return arg(1).val(); }
+    Value* idx2() { return arg(2).val(); }
     void updateType() override final {
-        auto t = arg<0>().val()->type;
-        if (arg<1>().val()->type.isScalar())
+        auto t = vec()->type;
+        if (idx1()->type.isScalar() && idx2()->type.isScalar())
             t.setScalar();
         maskEffectsAndTypeOnNonObjects(t);
+        if (!t.isA(PirType::num() | RType::str))
```
o-: I like the solution you found above. why didn't you apply it here?

rir/src/compiler/pir/instruction.h
```diff
@@ -1099,8 +1113,13 @@ class FLIE(Extract2_2D, 4, Effects::Any()) {
...
               {{PirType::val(), PirType::val(), PirType::val()}},
               {{vec, idx1, idx2}}, env, srcIdx) {}
+    Value* vec() { return arg(0).val(); }
+    Value* idx1() { return arg(1).val(); }
+    Value* idx2() { return arg(2).val(); }
     void updateType() override final {
-        maskEffectsAndTypeOnNonObjects(arg<0>().val()->type.scalar());
+        maskEffectsAndTypeOnNonObjects(vec()->type.scalar());
+        if (!vec()->type.isA(PirType::num() | RType::str))
+            type = type.orObject();
```
o-: same

o- (changes requested):
Hey, thinking about this some more, actually the whole thing seems broken to me in the vector case. (the terminology is a bit confusing. to the user the api to create a vector is called "list").

For example:
```
list(c(1,2), c(2,3))[[2]]
```

returns a double[2], so (1) the type of the result in the case of a vector is not related to the type of the vector at all and (2) it is certainly not guaranteed to be scalar, not even for `[[`.

Can you please investigate the issue more carefully and propose a complete fix. also it would be good to find a way to test those type annotations. We could try to use the TypeTest and instead of deoptimizing, just emit a call to error. So the program will just abort if a typecheck fails. If this happens on a test that is called always with the same types, then we know it is an error. Something like:

```
f <- function(a, b) a[[b]]
f(list(c(1,2)),1)
pir.compile(f, freezeTypes=TRUE)
f(list(c(1,2)),1)  # -> throws error, since we assumed that extract returns scalar
```

Jakobeha:
@o- type testing sounds good, I can use the safe force variable lookup code. One thing I also think would be good is implicit TypeTest assertions in debug mode, for every instruction. So e.g. if a PIR value's type is `int$'` and some compiler flag is enabled, RIR will explicitly check the value is an integer and fail.

For now I think I should just fix the non-scalar case by removing `isScalar`. Then PIR won't be able to optimize num vectors as well. But first we should work on scalar optimization, then maybe in the future, we could e.g. implement generic vector types for better vector inference.

Jakobeha:
Right now I've just implemented the vector fix, so all this branch does is removes unsound type inferences. I'm creating a separate type assertions branch from this branch.

o-:
> For now I think I should just fix the non-scalar case by removing isScalar

why? as I mentioned above, all kind of inference if `vec()` maybe type vector is wrong. so we should not do it. I don't see no point in just fixing one corner case, while leaving the actual problem there.

o-:
sorry, but I really do not want to review this 100x times.

first, let's use some standard terminology here. `[` is called subset, and `[[` is called extract.

Your `elem()` function returns (maybe) the type of a subset, not of a single element extract (which would be scalar for numbers). I think you need distinct `elementType` and `subsetType` functions, because they behave quite differently. Also, if you add helper methods there, you should also correctly handle the object case inside those (again sb. will come along and use it for different stuff).

For example, here are a couple of cases where I think the current version still has issues with:

* the subset of a vector is always a vector. For example: `v[x]` if v is a vector, then result must also be vector. see `typeof(list(1,2,3)[1])`.
* The subset of anything which is not an object, is *never* an object. This holds even for vectors.
* Extracting an element from a vector with a scalar index, will not necessarily return a scalar
* Extracting an element from a number/string will always return a scalar
* The subset and element of a "raw" is always a "raw"

as a sidenote:

The result of extracting something (except when vec() is an object) is never missing, so I think in those cases you can add notMissing

anyway, I would really appreciate if you fire up an R interpreter and try out all the combinations you want to handle in this PR and confirm that they really produce what you expect. And maybe write some tests in C++ code. something like

```
PirType r = RType::vector;
// list(1)[1]
assert(r.subsetType().isA(RType::vector));
assert(!r.subsetType().maybeObject());
assert(!r.orObject().subsetType().isA(RType::vector));
assert(!r.orObject().subsetType().isA(PirType::val());
// list(c(1,2))[[1]]
assert(r.elementType().isA(PirType::val());
assert(!r.elementType().scalar());
assert(!r.elementType.maybeMissing());
assert(!r.elementType.isA(RType::vector));
assert(r.orObject().elementType.maybeMissing());
```

o-:
thanks!

PR #464, opened 2019-04-23
loop invariant code motion

o-:


rir/src/compiler/opt/hoist_instruction.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    } else if (dom.dominates(trg, arg->bb())) {
+                        trg = arg->bb();
+                    } else {
+                        success = false;
+                    }
+                }
+
+                // !trg->isBranch, is a cheap heuristic to avoid moving
+                // something from the loop body to the loop header, which is
+                // usually not great.
```
charig: Why not great to move to the loop header? Where does the loop concept actually appear in all these code? Why we assume a loop header is branch node?

rir/src/compiler/opt/hoist_instruction.cpp
```diff
@@ -0,0 +1,131 @@
...
+            i->eachArg([&](Value* a) {
+                auto arg = Instruction::Cast(a);
+                // cppcheck-suppress knownConditionTrueFalse
+                if (!success || !arg)
+                    return;
+
+                // Try to find a hoisting candidate that is dominated by all
+                // arguments to i
+                // cppcheck-suppress knownConditionTrueFalse
+                if (!trg)
```
charig: target?

rir/src/compiler/opt/pass_definitions.h
```diff
@@ -133,6 +133,8 @@ class PASS(TypeInference);
 
 class PASS(TypeSpeculation);
 
+class PASS(HoistInstruction);
```
charig, line 139: Would you mind adding a comment specifying exactly what does this pass do?

rir/src/utils/configurations.cpp
```diff
@@ -51,6 +51,7 @@ void Configurations::defaultOptimizations() {
         optimizations.push_back(new pir::Constantfold());
         optimizations.push_back(new pir::Cleanup());
         optimizations.push_back(new pir::DelayInstr());
+        optimizations.push_back(new pir::HoistInstruction());
```
charig, line 54: Maybe it is time to change the name of DelayInstr and also maybe be more specific with HoistInstruction? For instance, `DelayLookup`, `HoistEffectless`?

rir/src/compiler/opt/hoist_instruction.cpp
```diff
@@ -0,0 +1,131 @@
...
+                // usually not great.
+                if (!success || trg == bb || trg->isBranch() ||
+                    !dom.dominates(trg, bb)) {
+                    success = false;
+                    return;
+                }
+
+                std::function<bool(BB*)> doesNotReorderEffects = [&](BB* x) {
+                    // empty or loop
+                    if (!x || x == trg)
```
charig: I do not get it. How do you distinguish the first iteration from a loop iteration? How are you sure that there is no another loop that does not start at trg?

rir/src/compiler/opt/pass_definitions.h
o-: to have it become outdated like the others :)

rir/src/utils/configurations.cpp
o-: it's not HoistEffectless. yes I agree that we should review the passes, also check if they actually work, remove stupid ones, cleanup naming and so on. but not really related to this pr

o-:
@charig I changed some stuff, addressed your comments and added some more documentation. feel free to request more changes still. it's not a great optimization. it does not work for instructions without inputs (for example it does not work for LdVar). but it does the trick for hoisting forces out of loops in many occasions...

rir/src/compiler/opt/hoist_instruction.cpp
```diff
@@ -0,0 +1,187 @@
...
+                std::function<bool(BB*)> compute = [&](BB* x) {
+                    if (!x || x == bb || x == target)
+                        return true;
+                    if (!x->isEmpty()) {
+                        // If we hoist over a branch, but one of the
+                        // branches does not need the value, then this will
+                        // waste computation
+                        if (x->last()->branches()) {
+                            if (!dom.dominates(x->trueBranch(), bb) ||
+                                !dom.dominates(x->falseBranch(), bb)) {
```
charig, line 161: Why there is one exception?

rir/src/compiler/opt/hoist_instruction.cpp
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                    return compute(x->next0) && compute(x->next1);
+                };
+                return compute(x->next0) && compute(x->next1);
+            };
+
+            bool success = true;
+            if (i->hasObservableEffects())
+                success = doesNotReorderEffects(target);
+            else if (i->cost() > 0)
```
charig, line 176: Why we have a blacklist but then a cost for ruling out `castType`?

rir/src/compiler/pir/instruction.h
```diff
@@ -937,6 +938,7 @@ class FLIE(Force, 2, Effects::Any()) {
 
 class FLI(CastType, 1, Effects::None()) {
   public:
+    unsigned cost() const override final { return 0; }
```
charig, line 941: cost of what?

charig (approved):

rir/src/compiler/opt/hoist_instruction.cpp
o-: supposedly a loop is something like:

```
a:
  branch b c
b:
    ....
    goto a
c:
```

so to hoist an instruction from `...` to `a:` we need to cross one branch instructions where not both branch targets dominate the loop body.

this is basically a very localized formulation of, only hoist out of one loop.

rir/src/compiler/opt/hoist_instruction.cpp
o-: it's the other way round. castType is free (since it's not an actual instruction) so we can always hoist it (success is always true)

rir/src/compiler/pir/instruction.h
o-: idk :) how expensive to execute, something... currently my idea was just to distinguish between instructions which will be scheduled in rir, and "virtual" ones.

but the idea was that maybe we can extend on it and use it for other optimizations. e.g. an add with environment could be more expensive than plain add. but I am not really sure. I would grow this as we need it...

rir/src/compiler/opt/hoist_instruction.cpp
o-: as I said in the comment, the better solution would be, to compute loop headers. then explicitly hoist instructions there. this would also allow us to hoist instructions without arguments. here I am just using the positions of the arguments of an instruction as a cheap heuristic of where it could be profitable to place an instruction.

maybe this could be achieved by combining the changes here with #362

PR #465, opened 2019-04-24
Elide env with partially undefined vars

o-:
If we have the case:

     e = env()
    loop:
     checkpoint ...
     stvar "x", 42, e
     goto loop

and want to move "e" into the deopt case, we need to create an env,
which is `env(x=phi(unboundValue,42))`, since on the first iteration
of the loop the value is not declared, but on the second iteration it
is.

This PR adds support for exactly that case.

### DIFF
--- a/rir/src/compiler/analysis/abstract_value.cpp
+++ b/rir/src/compiler/analysis/abstract_value.cpp
@@ -11,6 +11,7 @@ AbstractPirValue::AbstractPirValue() : type(PirType::bottom()) {}
 AbstractPirValue::AbstractPirValue(Value* v, Instruction* o,
                                    unsigned recursionLevel)
     : type(v->type) {
+    assert(o || v == UnboundValue::instance());
     vals.insert(ValOrig(v, o, recursionLevel));
 }
 
@@ -124,7 +125,12 @@ AbstractLoad AbstractREnvironmentHierarchy::get(Value* env, SEXP e) const {
         auto aenv = envs.at(env);
         if (!aenv.absent(e)) {
             const AbstractPirValue& res = aenv.get(e);
-            return AbstractLoad(env, res);
+            // UnboundValue has fall-through semantics which cause lookup to
+            // fall through.
+            if (res.maybeUnboundValue())
+                return AbstractLoad(env, AbstractPirValue::tainted());
+            if (!res.isUnboundValue())
+                return AbstractLoad(env, res);
         }
         auto parent = envs.at(env).parentEnv();
         assert(parent);
@@ -158,6 +164,9 @@ AbstractLoad AbstractREnvironmentHierarchy::getFun(Value* env, SEXP e) const {
             // If it might be a closure, we can neither be sure, nor exclude
             // this binding...
             if (res.type.maybe(RType::closure))
+                return AbstractLoad(env, AbstractPirValue::tainted());
+
+            if (res.maybeUnboundValue())
                 return AbstractLoad(env, AbstractPirValue::tainted());
         }
         auto parent = envs.at(env).parentEnv();
--- a/rir/src/compiler/opt/scope_resolution.cpp
+++ b/rir/src/compiler/opt/scope_resolution.cpp
@@ -112,8 +112,9 @@ class TheScopeResolution {
             return val;
         };
 
-        auto tryInsertPhis = [&](AbstractPirValue res, BB* bb,
-                                 BB::Instrs::iterator& iter) -> Value* {
+        auto tryInsertPhis = [&](Value* env, AbstractPirValue res, BB* bb,
+                                 BB::Instrs::iterator& iter,
+                                 bool allowUnbound) -> Value* {
             if (res.isUnknown())
                 return nullptr;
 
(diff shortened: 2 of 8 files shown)


o-:
So, the main idea in this PR is: if we merge two abstract environments with different bindings:  `[a->v]` and `[]` currently we create `[a->T]`. Now we merge this to `[a->(v|missing)]`.

in abstract lookup function in abstract_value.cpp we need to consider the missing case now:
1. all possible values are missing = fall through to the next environment
2. no possible missing = return current env abstract value of binding
3. some missing = give up and return tainted state)

So for the normal scope resolution pass this change is transparent and nothing is gained/lost. Where this really helps is in tryMaterializeEnv, where we can now say, at this point the binding is either uninitialized or a certain value.

There is a bit of noise in this PR, since I changed the scope analysis to record the BB of the write, instead of the writing instruction (for binding updates). The idea was to allow tracking the origin of values. In the end it was not needed, but I still liked the change.

rir/src/compiler/analysis/abstract_value.cpp
```diff
@@ -124,7 +125,12 @@ AbstractLoad AbstractREnvironmentHierarchy::get(Value* env, SEXP e) const {
         auto aenv = envs.at(env);
         if (!aenv.absent(e)) {
             const AbstractPirValue& res = aenv.get(e);
-            return AbstractLoad(env, res);
+            // UnboundValue has fall-through semantics which cause lookup to
+            // fall through.
+            if (res.maybeUnboundValue())
+                return AbstractLoad(env, AbstractPirValue::tainted());
+            if (!res.isUnboundValue())
```
charig, line 132: Can this happen? I do not see it.

rir/src/utils/configurations.cpp
```diff
@@ -107,12 +107,10 @@ void Configurations::defaultOptimizations() {
...
     // ==== Phase 4) Final round of default opts
-    for (size_t i = 0; i < 2; ++i)
+    for (size_t i = 0; i < 3; ++i) {
         addDefaultOpt();
-
-    // Our backend really does not like unused checkpoints, so be sure to remove
-    // all of them here already.
-    optimizations.push_back(new pir::CleanupCheckpoints());
+        optimizations.push_back(new pir::CleanupCheckpoints());
+    }
```
charig, line 113: At this point I would suggest that whenever we need to change this value, at least we add a test that justifies its necessity.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -851,8 +851,6 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
 
                 auto loadArg = [&](Value* what, size_t argNumber) {
                     if (what == UnboundValue::instance()) {
-                        assert(MkArg::Cast(instr) &&
```
charig, line 854: Is this change adding more possibilities that unbound values on phis? If that is the case I would prefer to keep the assert with an or including phis.

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -128,16 +129,31 @@ class TheScopeResolution {
 
             std::unordered_map<BB*, Value*> inputs;
             bool fail = false;
+            Value* initialValue = nullptr;
```
charig: initialValue? Isn't this only here for unbound?

rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -112,8 +112,9 @@ class TheScopeResolution {
             return val;
         };
 
-        auto tryInsertPhis = [&](AbstractPirValue res, BB* bb,
-                                 BB::Instrs::iterator& iter) -> Value* {
+        auto tryInsertPhis = [&](Value* env, AbstractPirValue res, BB* bb,
```
charig, line 115: It is confusing for me the env as the first parameter. Isn't it there only for unbounds? Maybe only one parameter at the end unboundBB and if it is null we know !allowUnbound?

charig (approved):
I wonder if tracking the BB instead of the instruction could introduce any weird interleaving. For instance, what if we move an instruction during scope resolution and then we looking to the abstract state we end up obtaining the wrong bb?

rir/src/compiler/analysis/abstract_value.cpp
o-: Currently probably not. But it does not hurt. e.g. we could in the future track `rm` and implement it by writing unboundValue...

rir/src/utils/configurations.cpp
o-: good point. I wanted to get rid of environments in full mandelbrot. and that was necessary at some point.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: phis, pircopy and mkenv, from the top of my head. do you want an assert on those?

rir/src/compiler/opt/scope_resolution.cpp
o-: well, again, it would work for any initial value... e.g.:

```
mkenv(x=true)
if (...)
  x <- 1
```

here we have `{val=true, origin=nullptr} | {val = 1, origin=x<-1}` as the abstract state

o-: actually no, in this case we set the mkenv as the origin. you convinced me. this should only happen with unbound.

o-:
> I wonder if tracking the BB instead of the instruction could introduce any weird interleaving

good point. gonna revert that part

PR #466, opened 2019-04-24
Better type inference

o-:


o-:
just the last commit. add just one new case to the type inference. also safeBuiltins don't return missing values. there are probably some more instructions where we should reconsider `val()` vs. `val().notMissing()`

rir/src/compiler/opt/types.cpp
```diff
@@ -40,33 +54,25 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
...
-                                if (types.count(val))
-                                    infered = types.at(val).notObject();
-                                else
-                                    infered = val->type.notObject();
-                            }
-                    break;
+                case Tag::CallSafeBuiltin: {
+                    auto c = CallSafeBuiltin::Cast(i);
+                    std::string name = getBuiltinName(getBuiltinNr(c->blt));
+                    if ("bitwiseXor" == name || "bitwiseShiftL" == name) {
```
mhyee: I'm always a bit wary when things like this are hardcoded... Isn't there a list of builtins?

rir/src/compiler/opt/types.cpp
```diff
@@ -40,33 +54,25 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
...
-                        if (auto isO = IsObject::Cast(assumption->condition()))
-                            if (auto val =
-                                    Instruction::Cast(isO->arg<0>().val())) {
-                                if (types.count(val))
-                                    infered = types.at(val).notObject();
-                                else
-                                    infered = val->type.notObject();
-                            }
-                    break;
+                case Tag::CallSafeBuiltin: {
```
mhyee, line 70: Why was the `Assume` case removed?

o-: right, we used to just do the following:
```
t = IsObject(o)
Assume(t)
...
```
but with the type speculation branch, I actually changed this to
```
t = IsObject(o)
Assume(t)
t1 = CastType(t, t->type.notObject())
...
```
and then replace t with t1. so this case is just not needed anymore, since the assumption is already encoded in the castType.

rir/src/compiler/opt/types.cpp
o-: benchmark *cough*. just wanted to see it working :) I am not quite sure what the generic solution would be, since it highly depends on the actual builtin. also for each builtin you would basically need a matrix `t1 x t2 x ... x tn -> t` that tells you for every input combination, what the result type is.

rir/src/compiler/opt/types.cpp
```diff
@@ -40,33 +54,25 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
...
+                    std::string name = getBuiltinName(getBuiltinNr(c->blt));
+                    if ("bitwiseXor" == name || "bitwiseShiftL" == name) {
+                        infered = PirType(RType::integer);
+                        if (getType(c->arg(0).val()).isScalar() &&
+                            getType(c->arg(1).val()).isScalar())
+                            infered.setScalar();
+                        break;
+                    }
                 }
+                // fall through
```
mhyee: Guess we should update this now. :)

rir/src/compiler/opt/types.cpp
mhyee: I'm still unhappy about hardcoding this, especially if it's just for the benchmark.

Are there any other simple cases we could add here?

o-:
> I'm still unhappy about hardcoding this

I don't understand... how else would you do this? there are some builtins that have particular return types. and it is very specific to the actual builtin. the bitwise always return integers. but we will have to come up and write down such ad-hoc rules for every builtin eventually. I don't see any other way...

mhyee:
> I don't understand... how else would you do this? there are some builtins that have particular return types. and it is very specific to the actual builtin. the bitwise always return integers. but we will have to come up and write down such ad-hoc rules for every builtin eventually. I don't see any other way...

I think I'd prefer to see more of the rules, say for the other bitwise operations, sooner rather than later. (EDIT: Which  see you just did, thanks!)

Right now, this smells too much like "implement something that only works for a single benchmark."

mhyee:
Also, I just noticed a typo, `infered` -> `inferred`.

o-:
> Right now, this smells too much like "implement something that only works for a single benchmark."

well, this is what it is. but we gotta start somewhere...

PR #467, opened 2019-04-24
Safe force for stronger tests

Jakobeha:
`pir.check` safe forces arguments, so we can get more assumptions. Also, you can manually enable always safe force with `RIR_STRONG_SAFE_FORCE=1`.


Jakobeha:
Tried to rename this but it confused git

PR #468, opened 2019-04-24
[wip] Force args for tests

Jakobeha:
`pir.check` safe forces arguments, so we can get more assumptions. Also, you can manually enable always safe force with `RIR_STRONG_SAFE_FORCE=1`.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -63,7 +63,7 @@ pir.tests <- function() {
 
 # returns TRUE f, when PIR compiled, satisfies the the given checks (e.g.
 # environment was elided). Max assumptions compiled (+ minimal) are used, if
-# warmup=list(...) will call the function with ... to get better assumptions.
+# warmup=<FUN> will call <FUN> repeatedly to get better assumptions.
 pir.check <- function(f, ..., warmup=NULL) {
     checks <- 
         as.pairlist(lapply(lapply(as.list(substitute(...())), as.character), as.name))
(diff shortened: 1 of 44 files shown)


Jakobeha:
This is mixed up with #461 (only includes the commits it depends on), so if that gets merged I can resolve conflicts afterward.

o-:
@Jakobeha sorry, I am not sure if I understand this PR. can you please add some language as to what you are doing here and why?

and why is it behind a flag? (this has no test coverage this way)

o-:
if this is what I think it is, namely looking up variables at call time, then it is not sound. see this example:

```
function() {
  a <- 1
  f <- function(x) {
    rm("a", envir = sys.frame(-1))
    x
  }
  f(a)
}
```

this should throw an error, but returns 1, when you lookup a already at the callsite `f(a)`

Jakobeha:
- The PR looks up variables when a function is called with them, and recursively "safe" forces their promises, as you guessed.
- This only runs under a flag because it was generating some errors, partly because we were regressing, adding sound assumptions which caused future unsound optimizations (e.g. #461), but now also because it was unsound.
- However, it still lets us get better type inference when it is sound, which could help us write better tests.

I think I should add an explicit flag `forceArgs` to `pir.check`, and then instead of doing this "safe" force, just regularly force the args.

o-:
but we probably do not want to merge it, if it is unsound. we should talk about it offline...

PR #469, opened 2019-04-25
Native

o-:


rir/src/compiler/translations/pir_2_rir/liveness.h
```diff
@@ -0,0 +1,42 @@
...
+namespace rir {
+namespace pir {
+
+struct BBLiveness {
+    uint8_t live = false;
+    unsigned begin = -1;
+    unsigned end = -1;
+};
+
+struct Liveness : public std::vector<BBLiveness> {
```
mhyee: This file should be deleted; leftover from a merge.

rir/src/interpreter/interp.cpp
```diff
@@ -1509,6 +1512,21 @@ void deoptFramesWithContext(InterpreterInstance* ctx,
 SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
                  const CallContext* callCtxt, Opcode* initialPC,
                  R_bcstack_t* localsBase) {
+    if (c->nativeCode) {
+        if (!callCtxt || callCtxt->hasStackArgs() ||
+            (callCtxt->suppliedArgs == 0 &&
+             callCtxt->givenAssumptions.includes(
+                 Assumption::NotTooFewArguments))) {
+            void* args = callCtxt ? (void*)callCtxt->stackArgs : (void*)0xdead;
+            return ((SEXP(*)(Code*, InterpreterInstance*, void*, SEXP,
```
mhyee: Whoa, this is ugly. Can you split this into two lines or something, i.e. one to cast `c->nativeCode` and one to call its result?

mhyee (commented):
This is not an actual review: I skimmed over this super quickly and pretty much skipped `builtins.cpp` and `lower.cpp`.

rir/src/interpreter/interp.cpp
o-: hehe, thanks, yeah, this is all still pretty much written in anger... I'll do another pass, and let ppl know if it's ready for an actual review. I am just expecting to also hit a couple of roadblocks making this work on CI that's why I started with this PR.

PR #470, opened 2019-04-25
Type assertions

Jakobeha:
When `RIR_CHECK_PIR_TYPES=1`, after every PIR instruction with a non-void type is converted into RIR bytecode, an `assert_type_` opcode will be emitted with the instruction's type. This will assert that the top of stack SEXP actually has the instruction's type.

This makes an ugly dependency on PIR, since the RIR interpreter accesses PIR types. It isn't too bad and doesn't create cycles, because `pir/type` doesn't have much dependencies. I think the best solution would be to move `PirType` outside of PIR, maybe into the `ir` folder. Although I'm not sure what everyone else thinks, so I didn't do this yet.

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -63,7 +63,7 @@ pir.tests <- function() {
 
 # returns TRUE f, when PIR compiled, satisfies the the given checks (e.g.
 # environment was elided). Max assumptions compiled (+ minimal) are used, if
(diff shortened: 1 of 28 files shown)


Jakobeha:
This is mixed up with #461 (only includes the commits it depends on), so if that gets merged I can resolve conflicts afterward.

PR #471, opened 2019-04-25
New caching strategy

charig:
Summary of changes:
* Extend Code objects so that they also store the #different `<ldVar/stVar, env>` they contain. This enables to create binding caches of dynamic size eventually. With static size, it enables to optimize based on the fact that there are no collisions whenever this number is < cache size.      
* Extend compiler so that it assigns a unique number to every `<ldVar/stVar, env>`.
* Extend pir2rir so that it assigns a unique number to every `<ldVar/stVar, env>`. This enables to avoid the cache clearing we are performing after every environment change. 
* Extend rir bytecode so every load/stote has its corresponding (load/store)Cache version. The cache versions contain one extra immediate: the unique number calculated. The standard versions do not use the cache at all. This PR does not implement it, but this allow us to avoid all the caching overhead for loads and stores for which PIR can prove are not reused, etc.

TODOs:
* Avoid memsetting when pir does not need it.
* Move from a static array to a dynamic container (R stack mapping) initialized with the precise size needed. 
* It seems like mem setting is one of the most significant sources of performance. This seems to be different when mapping in the R stack. By mapping on the R stack we can also reuse in some RIR trampoline calls (loops) thus avoiding extra memsettings.
* Analysis at the PIR level to use standard instructions (no caching) whenever we know beforehand the values will not be reused enough times.
* What happens with caches fpr promises? Can we do better in PIR?


rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -29,438 +29,6 @@ namespace pir {
 
 namespace {
 
-/*
```
o-, line 32: did you move this stuff unchanged or are there changes in the code you moved into the other file?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -1336,8 +934,8 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
     });
     cb.flush();
 
-    auto localsCnt = alloc.slots();
-    auto res = ctx.finalizeCode(localsCnt);
+    auto res =
```
o-: I feel like some documentation get's lost with this change. why are we passing slots() ? in the old version we knew: because its the locals count...

rir/src/interpreter/interp.cpp
```diff
@@ -1520,11 +1535,13 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
 #endif
 
     assert(c->info.magic == CODE_MAGIC);
+    bool existingLocals = localsBase;
 
     BindingCache bindingCache[BINDING_CACHE_SIZE];
-    memset(&bindingCache, 0, sizeof(bindingCache));
+    bool smallCache = c->bindingsCount <= BINDING_CACHE_SIZE;
+    if (env != symbol::delayedEnv || existingLocals)
```
o-: I don't think existingLocals allows you to not initialize the cache

o-: can you add a comment that the zeroing is delayed to the setEnv, it took me a bit to understand it

rir/src/interpreter/interp.cpp
```diff
@@ -1644,7 +1660,7 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
                     }
                 }
             }
-
+            memset(&bindingCache, 0, sizeof(bindingCache));
```
o-, line 1672: would be nice if we only zero affected bindings

rir/src/interpreter/interp.cpp
```diff
@@ -1232,43 +1232,55 @@ static SEXP seq_int(int n1, int n2) {
...
 
-    Immediate cidx = idx % BINDING_CACHE_SIZE;
-    if (bindingCache[cidx].idx == idx) {
+    Immediate cidx;
+    if (smallCache)
+        cidx = cacheIdx;
+    else
+        cidx = cacheIdx & CACHE_MASK;
+
+    if (bindingCache[cidx].idx == poolIdx) {
```
o-: so we changed the cache to contain entries from different environments (to avoid clearing them on change env).

so this is not correct, you would need to check if the environment is correct too. you would need to add the env the cache is for to the bindingCache cell

rir/src/ir/insns.h
```diff
@@ -54,17 +54,29 @@ DEF_INSTR(ldfun_, 1, 0, 1, 0)
  */
 DEF_INSTR(ldvar_, 1, 0, 1, 0)
 
+/**
+ * ldvar_:: like ldvar.
+ * Stores an additional immediate with a unique number for the cache bindings.
+ */
+DEF_INSTR(ldvar_cache_, 2, 0, 1, 0)
```
o-: I would have prefered a lot, if we would not double the instructions. instead cache idx of 0 (or -1) could mean no caching.

charig: Ok, but do you have a strong reason? Because unless we have some "size" like argument it is usually better to have more instructions with particular semantics than adding an if to every load/store

rir/src/interpreter/interp.cpp
charig: Yes, but far from straightforward...

rir/src/interpreter/interp.cpp
charig: Why? 
In the cache slot we store the unique number determining a pair<symbol,env>.

o-: so assume `<"a", e1>` has id 1 and `<"a", e2>` has id 9. they will both end up in slot 1, the check will pass, since both refer to the symbol "a" and you will read the wrong binding.

rir/src/ir/insns.h
o-: less code duplication (they all do almost the same. I mean in pir you chose to use the same instruction) and most will have a cache

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -0,0 +1,506 @@
...
+
+struct nameAndEnvHash {
+  public:
+    template <typename T, typename U>
+    std::size_t operator()(const std::pair<T, U>& x) const {
+        return std::hash<T>()(x.first) ^ std::hash<U>()(x.second);
+    }
+};
+
+class CachePositionAllocator {
```
o-, line 442: I didn't see this, because I did not see that you added new stuff to the file.

Why do you need to preprocess this? you should be able to do this on the fly. if you already saw a pair, return the id, otherwise assign a new id.

rir/src/interpreter/interp.cpp
charig: No, because `bindingCache[1].idx = 1` or `bindingCache[1].idx = 9` to distinguish both cases.

rir/src/ir/insns.h
charig: Ok, but the distinction was on purpose. PIR is not "executable" -> I do not mind executing ifs at compile time why I may mind execute an if !=-1 for every load/store RIR BC.

Oth, (all) have a cache now. But I do not think that will necessarily be the like that in future versions. Analysis may realize for lot of cases that the cache is not needed (for instance load stores outside loops)

rir/src/compiler/translations/pir_2_rir/allocators.h
charig: Yes, when I started I thought I may need sth fancier, it ended up being pretty simple. Anyway, I do not think one more visitor pass harms at this point and the lowering is more readable like this.

o-: why is it more readable? you don't even need to change the api. just change this class here, to not do a run before

rir/src/interpreter/interp.cpp
o-: https://github.com/reactorlabs/rir/pull/471#discussion_r280716465

o-: duh, sorry. I did not notice that you changed the index to be the cache index instead of the symbol

o-: but the idx field does not contain the cacheIdx, but actually the poolIdx, which is just the symbol, not?

o-: also, this check is not needed on small caches. so it should be
```
if (smallCache || bindingCache[cidx].idx == cacheIdx)
```

and then store the cacheIdx into the field everywhere where you update it

charig: Ah ok, yeah, the only issue was it should store the cacheIdx yes.
The check is still needed right now because we have to avoid loading of uninit variables

rir/src/interpreter/interp.cpp
```diff
@@ -1520,11 +1534,20 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
 #endif
 
     assert(c->info.magic == CODE_MAGIC);
+    bool existingLocals = localsBase;
 
     BindingCache bindingCache[BINDING_CACHE_SIZE];
-    memset(&bindingCache, 0, sizeof(bindingCache));
+    // Position zero is not counted for smallcache because we can not
+    // distinguish slot zero from uninit.
```
o-, line 1541: ehm, use cacheLocation-1 as the index?

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -200,20 +200,22 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
         break;
 
     case Opcode::ldvar_:
-    case Opcode::ldvar_for_update_:
+    case Opcode::ldvar_cache_:
+    case Opcode::ldvar_for_update_cache_:
```
o-, line 204: only saw this now, but in general cached (with d) would be a much better name for all of them. because cache is used as adverb and not verb: perform a lookup, a cached one.

rir/src/interpreter/interp.cpp
```diff
@@ -1232,43 +1232,54 @@ static SEXP seq_int(int n1, int n2) {
...
 
-    Immediate cidx = idx % BINDING_CACHE_SIZE;
-    if (bindingCache[cidx].idx == idx) {
+    Immediate cidx;
+    if (smallCache)
+        cidx = cacheIdx;
+    else
+        cidx = cacheIdx & CACHE_MASK;
+
+    if (bindingCache[cidx].idx == cacheIdx)
```
o-, line 1258: i would go for
```
bool hit = smallCache || bindingCache[cidx].idx == cacheIdx;
if (hit && bindingCache[cidx].loc)
  ...
```

this should be a bit faster for small cache and additionally solve the zero index issue

rir/src/interpreter/interp.cpp
o-: (or see suggestion at https://github.com/reactorlabs/rir/pull/471#discussion_r281169517)

rir/src/interpreter/interp.cpp
```diff
@@ -3316,7 +3414,11 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
         }
 
         INSTRUCTION(guard_fun_) {
+#ifndef UNSOUND_OPTS
```
o-: why is this here?

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -0,0 +1,470 @@
...
+            if (*i == v)
+                return false;
+        for (size_t i = argNumber + 1; i < instr->nargs(); ++i)
+            if (instr->arg(i).val() == v)
+                return false;
+        return true;
+    }
+};
+
+struct nameAndEnvHash {
```
o-: this is very strange. on the one hand you call this nameAndEnvHash, suggesting that this is just for this name/env pair. but then the actual implementation is a template function, which is only necessary if you want to reuse this.

please just use `pairhash<SEXP,Value*>` from common.h

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -0,0 +1,470 @@
...
+
+class CachePositionAllocator {
+  public:
+    typedef size_t SlotNumber;
+
+    size_t numberOfBindings() { return uniqueNumbers.size(); }
+    SlotNumber slotFor(SEXP varName, Value* environment) {
+        std::pair<SEXP, Value*> key =
+            std::pair<SEXP, Value*>(varName, environment);
+        return uniqueNumbers.emplace(key, uniqueNumbers.size() + 1)
```
o-, line 449: this always overrides and never returns the same value?

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -0,0 +1,470 @@
...
+    }
+};
+
+class CachePositionAllocator {
+  public:
+    typedef size_t SlotNumber;
+
+    size_t numberOfBindings() { return uniqueNumbers.size(); }
+    SlotNumber slotFor(SEXP varName, Value* environment) {
+        std::pair<SEXP, Value*> key =
```
o-: `nameAndEnv key`

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -0,0 +1,470 @@
...
+            std::pair<SEXP, Value*>(varName, environment);
+        return uniqueNumbers.emplace(key, uniqueNumbers.size() + 1)
+            .first->second;
+    }
+
+  private:
+    typedef std::pair<SEXP, Value*> nameAndEnv;
```
o-: by convention we write types with first letter capitalized. `NameAndEnv`

o- (approved):

rir/src/compiler/translations/pir_2_rir/allocators.h
charig: emplace does not insert if the key is already defined and returns that key

rir/src/interpreter/interp.cpp
charig: `cacheLocation - 1` implies substracting on every access, I'd prefer to just not use the zero index, its only 16 bytes lost. 

Testing for `loc!=null` enables to use the zero index too, but it implies to add one more check for the nonSmall version (cacheIdx + loc != null) while it adds nothing to the smallcache (previously testing for idx now for loc)

rir/src/compiler/translations/pir_2_rir/allocators.h
o-: nevermind

rir/src/ir/Compiler.cpp
```diff
@@ -458,497 +470,497 @@ bool compileSpecialCall(CompilerContext& ctx, SEXP ast, SEXP fun, SEXP args_, bo
...
+        return true;
+    }
+    Else(break)
+}
 
-        // 3) Special case [ and [[
-
-        if (lhsParts.size() != 2) {
-            return false;
+// Find all parts of the lhs
```
mhyee, line 487: Whoa, what happened to the indentation here?

o-: hah, beatiful. yeah, I had to disable whitespaces in the diff settings to be able to review this PR at all. that's why I did not see it. I think this is the final argument why we should never run clang-format on the whole repo. @charig , can you please revert those whitespace changes?

PR #473, opened 2019-04-26
ban switch fall through

o-:


CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ set(CMAKE_CXX_FLAGS_RELEASE "-O2 -Werror")
 set(CMAKE_CXX_FLAGS_RELEASENOASSERT "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG") 
 set(CMAKE_CXX_FLAGS_DEBUG "-O0 -DENABLE_SLOWASSERT")
 set(CMAKE_CXX_FLAGS_DEBUGOPT "-Og -DENABLE_SLOWASSERT")
-set(CMAKE_CXX_FLAGS "-Wall -Wuninitialized -Wundef -Winit-self -Wcast-align -Woverloaded-virtual -Wctor-dtor-privacy -Wmissing-include-dirs -Wstrict-overflow=5 -std=c++14 -fno-rtti -fno-exceptions")
+set(CMAKE_CXX_FLAGS "-Wall -Wuninitialized -Wundef -Winit-self -Wcast-align -Woverloaded-virtual -Wctor-dtor-privacy -Wmissing-include-dirs -Wstrict-overflow=5 -std=c++14 -fno-rtti -fno-exceptions -Wimplicit-fallthrough")
```
mhyee, line 18: Probably doesn't matter, but might as well keep `.ycm_whatever_config` in sync.

mhyee:
So there were only _three_ fallthroughs that we had to rewrite, and two of them were for `assert(false)`? Banning fallthrough seems like a no-brainer.

I guess we have a lot of situations like
```c++
  switch (foo) {
  case A:
  case B:
  case C:
    // do something
  default:
    // else
  }
```

which are OK.

o-:
> So there were only three fallthroughs that we had to rewrite, and two of them were for assert(false)? 

yeah, I am honestly embarrassed for defending that feature now...

PR #475, opened 2019-04-30
added cleanup handler

Jakobeha:
Adds a function `rirCleanup` which will be called when R exits.

Would be used by #474, also see https://github.com/reactorlabs/gnur/pull/3

### DIFF
--- a/rir/src/api.cpp
+++ b/rir/src/api.cpp
@@ -366,3 +366,5 @@ bool startup() {
 }
 
 bool startup_ok = startup();
+
+void rirCleanup() {}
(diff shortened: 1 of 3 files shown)


rir/src/api.cpp
```diff
@@ -366,3 +366,5 @@ bool startup() {
 }
 
 bool startup_ok = startup();
+
```
charig, line 369: I do not get it. Why do we send to GNU-R and empty function for the cleanup?

Jakobeha: Was for a #474, although I can just use `on.exit`

PR #476, opened 2019-05-01
Binding cache of dynamic size for every activation

charig:
Most of the changes are from #471 .

Single commits since "First simple iteration for a dynamically sized cache" are the only relevant.


rir/src/interpreter/cache.h
```diff
@@ -0,0 +1,126 @@
...
+        cidx = index & CACHE_MASK;
+    return cidx;
+}
+
+#ifdef CACHE_ON_R_STACK
+// TODO: Create a version with a cache on the R_stack instead of the C stack
+#else
+typedef struct {
+    SEXP loc;
+    Immediate idx;
```
o-: drive-by comment: this should not be needed anymore. if we have one fix location per <env,name> pair, then there is just no need to check if it is the correct name

rir/src/interpreter/cache.h
```diff
@@ -0,0 +1,125 @@
+#ifndef RIR_INTERPRETER_CACHE_H
+#define RIR_INTERPRETER_CACHE_H
+
+#include "instance.h"
+
+namespace rir {
+
+// must be a power of 2 for modulus using & CACHE_MASK to work
+#define MAX_CACHE_SIZE 128
+#define CACHE_MASK (MAX_CACHE_SIZE - 1)
```
o-, line 10: ok, but now I really don't see why this is still needed. just don't create more than MAX_CACHE_SIZE worth of cached loads/stores

charig: What if a function has more bindings? Is there a precise limit?

o-: as I said, just don't cache more. you can decide to emit cached/non-cached loads. if you have say more than 200 env/symbol pairs, then just stop emiting cached ones. this automatically ensures that the index is never bigger than 200.

o-: @charig the immediateIdx should not be needed anymore, that was my point for this change...

charig: As I've been saying, we still use the idx to distinguish from uninitialized. I do not understand precisely what do you have in mind for optimizing here.

PR #477, opened 2019-05-01
Unsound opts in release mode

charig:


o-:
how about not emiting the guards in the first place?

rir/src/interpreter/interp.cpp
```diff
@@ -3435,7 +3435,9 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
         }
 
         INSTRUCTION(guard_fun_) {
+#ifndef UNSOUND_OPTS
```
o-: why is this here? an env flag and a define flag with exactly the same name is very confusing. also, with the other change this is probably not needed anymore.

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -122,7 +124,9 @@ std::unordered_set<Opcode*> findMergepoints(rir::Code* srcCode) {
     std::unordered_set<Opcode*> mergepoints;
     // Create mergepoints
     for (auto m : incom)
-        if (std::get<1>(m).size() > 1)
+        // The first position must also be considered a mergepoint in case it
+        // has only one incoming (a jump)
+        if (std::get<0>(m) == first || std::get<1>(m).size() > 1)
```
o-, line 129: It would be really useful if you could also figure out why this caused the vm to grab all the memory insted of just throwing an assert or something else, which is easier to debug. (and fix that too)

o- (approved):

rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
charig: Let's finish this PR that already made me suffer a lot :). I open #488

PR #478, opened 2019-05-02
Loop detection code

mhyee:
```
Does the standard textbook approach of identifying loops. A loop is
defined by a header h, and a back edge n->h where h dominates n. If
there are multiple back edges to the same header, the body is the union
of all the nodes computed for each back edge.

Once back edges are identified, we do a reverse depth-first search from
the "tail" until we reach the header.

There is an option to compute the loop-nesting forest, but it disabled
by default because it is slow.
```

I'm splitting this off from my loop peeling work, to make things more manageable and because #362 needs this.

However, as this PR stands, the loop detection code is never reached, and won't be until we have loop-invariant hoisting and/or loop peeling.

### DIFF
--- a/rir/src/compiler/analysis/loop_detection.cpp
+++ b/rir/src/compiler/analysis/loop_detection.cpp
@@ -4,29 +4,20 @@
 namespace rir {
 namespace pir {
 
-bool LoopDetection::Loop::comesBefore(Instruction* a, Instruction* b) const {
-    if (a->bb() == b->bb()) {
-        return a->id().idx() < b->id().idx();
-    } else {
-        // ordering vector is reversed: larger means precedes
-        return ordering_.at(a->bb()) > ordering_.at(b->bb());
-    }
-}
-
 LoopDetection::LoopDetection(Code* code, bool determineNesting) {
     CFG cfg(code);
     DominanceGraph dom(code);
     // map of header nodes to tail nodes
     std::unordered_map<BB*, BBList> tailNodes;
 
-    // find back edges, i.e. edges n->h where h dominates n
+    // find back edges, i.e. edges tail->header where header dominates tail
     Visitor::run(code->entry, [&](BB* maybeHeader) {
-        for (const auto& n : cfg.immediatePredecessors(maybeHeader)) {
-            if (dom.dominates(maybeHeader, n)) {
+        for (const auto& maybeTail : cfg.immediatePredecessors(maybeHeader)) {
+            if (dom.dominates(maybeHeader, maybeTail)) {
                 if (tailNodes.count(maybeHeader)) {
-                    tailNodes[maybeHeader].push_back(n);
+                    tailNodes[maybeHeader].push_back(maybeTail);
                 } else {
-                    tailNodes.emplace(maybeHeader, BBList({n}));
+                    tailNodes.emplace(maybeHeader, BBList({maybeTail}));
                 }
             }
         }
@@ -36,8 +27,6 @@ LoopDetection::LoopDetection(Code* code, bool determineNesting) {
     for (auto& l : tailNodes) {
         const auto& header = l.first;
         auto& todo = l.second;
-        std::unordered_map<BB*, size_t> ordering;
-        size_t counter = 0;
         BBSet body = {header}; // header is part of loop body
 
 #ifdef DEBUG_LOOP_DETECTION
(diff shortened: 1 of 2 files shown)


rir/src/compiler/analysis/loop_detection.cpp
```diff
@@ -0,0 +1,106 @@
+#include "loop_detection.h"
+#include "../util/visitor.h"
+
+namespace rir {
+namespace pir {
+
+bool LoopDetection::Loop::comesBefore(Instruction* a, Instruction* b) const {
+    if (a->bb() == b->bb()) {
+        return a->id().idx() < b->id().idx();
```
o-: just a drive-by comment.

this is very (!) slow... since this scans the instruction array for both instructions. I think we should probably remove idx() from the api. (I think we only use it for printing, where I guess we don't care that it is slow).

Can you pass iterators instead of instructions? because then it is just `a < b`.

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -0,0 +1,105 @@
+#ifndef PIR_LOOP_DETECTION_H
+#define PIR_LOOP_DETECTION_H
+
+#include "../pir/pir.h"
+#include "../util/cfg.h"
```
charig: needed?

rir/src/compiler/analysis/loop_detection.h
```diff
@@ -0,0 +1,105 @@
...
+             const std::unordered_map<BB*, size_t>& ordering)
+            : header_(header), body_(std::move(body)), outer_(nullptr),
+              isInnermost_(true), ordering_(std::move(ordering)) {}
+
+        Loop(const Loop&) = delete;
+        Loop& operator=(const Loop&) = delete;
+        Loop(Loop&&) = default;
+        Loop& operator=(Loop&&) = default;
+
+        RIR_INLINE BB* header() const { return header_; }
```
charig: Do we need the RIR_INLINE? I'd rely on gcc for inlining here.

rir/src/compiler/analysis/loop_detection.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+LoopDetection::LoopDetection(Code* code, bool determineNesting) {
+    CFG cfg(code);
+    DominanceGraph dom(code);
+    // map of header nodes to tail nodes
+    std::unordered_map<BB*, BBList> tailNodes;
+
+    // find back edges, i.e. edges n->h where h dominates n
+    Visitor::run(code->entry, [&](BB* maybeHeader) {
+        for (const auto& n : cfg.immediatePredecessors(maybeHeader)) {
```
charig: Is it necessary the const&? 

n -> bb? or maybeTail?

rir/src/compiler/analysis/loop_detection.cpp
```diff
@@ -0,0 +1,106 @@
...
+        // depth-first search starting from the loop tail(s)
+        while (!todo.empty()) {
+            BB* cur = todo.back();
+#ifdef DEBUG_LOOP_DETECTION
+            std::cerr << " " << cur->id;
+#endif
+            todo.pop_back();
+            if (!body.count(cur)) {
+                body.insert(cur);
+                ordering[cur] = counter++;
```
charig: If we have multiple tails the ordering works because we know our visitor for including the tails is bfs? If that is the case maybe a comment?

charig (approved):

rir/src/compiler/analysis/loop_detection.cpp
mhyee: Thanks for the catch -- this would explain why it's so slow.

For now, I might just rip it out of this PR, and only add it in the loop peeling PR. The thing is, I find this `comesBefore` and `ordering` vector to be really hacky, and I'd like to find a better solution.

rir/src/compiler/analysis/loop_detection.h
mhyee: Bad habit; I think I wrote this before we discovered forcing inline was bad.

Now, `RIR_INLINE` is defined as `inline` which the compiler can decide to ignore. Interestingly, if you write an inline definition in a class definition like this, it's implicitly marked as `inline`.

In any case, I removed `RIR_INLINE`.

rir/src/compiler/analysis/loop_detection.cpp
mhyee: I think you're right; this might be broken:

image

In blue, node 5 should have been given a number less than 2.

I should have done my DFS/toplogical sort forwards, starting from the header, rather than going backwards from the tails. (See red numbering.)

In any case, I removed it from this PR, and I'll put a TODO in my loop-peeling branch.

charig (approved):

PR #479, opened 2019-05-02
Add a BBTransform helper to remove dead instructions.

mhyee:
The GVN pass can sometimes generate dead phis that are also invalid;
normally, these phis would be removed by the cleanup pass. However, this
could trip the verifier.


o-:
thanks

PR #480, opened 2019-05-02
Full verifier for CI

mhyee:
The verifier has a fast and a slow (or full) version that can run. Currently, we call the verifier in the following situations:

1. [The first time we compile from RIR to PIR.](https://github.com/reactorlabs/rir/blob/28aeb2084962ed2521ba0f54d370025cdbb116a6/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp#L159) (Fast verifier, in all builds.)
2. [After each optimization pass.](https://github.com/reactorlabs/rir/blob/28aeb2084962ed2521ba0f54d370025cdbb116a6/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp#L205) (Fast verifier, in debug builds, where `ENABLE_SLOWASSERT` is defined.) 
3. After all optimizations have completed.
    * [In the debug builds](https://github.com/reactorlabs/rir/blob/28aeb2084962ed2521ba0f54d370025cdbb116a6/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp#L217) (with `ENABLE_SLOWSSERT`) this will be the full verifier.
    * [In the release build](https://github.com/reactorlabs/rir/blob/28aeb2084962ed2521ba0f54d370025cdbb116a6/rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp#L220), this will be the fast verifier.
    * In the releasenoassert (with `NDEBUG`), this will be skipped.
4. [Just before compiling from PIR to RIR.](https://github.com/reactorlabs/rir/blob/28aeb2084962ed2521ba0f54d370025cdbb116a6/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp#L676) (Full verifier, in debug builds, where `ENABLE_SLOWASSERT` is defined.)

---

To summarize, in debug builds, we run the verifier:
  * When we compile from RIR to PIR (fast)
  * After each optimization pass (fast)
  * After all optimizations have completed (slow)
  * Just before compiling from PIR to RIR (slow)

In release builds, we run the verifier:
  * When we compile from RIR to PIR (fast) (always)
  * After all optimizations have completed (fast) (unless we disabled assertions)

---

There is (at least) one omission: we never run the full verifier after _every_ optimization pass. This is understandable because we have many passes (probably in the hundreds by now, because we run them multiple times). On the other hand, sometimes bugs slip in; this is why we have #479, because the GVN pass trips up the verifier.

Therefore, I added a build config that always runs the full verifier after each optimization pass, but in release mode. The intent is to only run this on CI.

For now, I want to see how long that particular CI config takes to run. Having summarized where we run the verifier, I'm wondering if there are other cases we should consider running the full verifier. I'm also unhappy with the mess of `ifdefs` in `rir_2_pir_compiler`.

Any thoughts or suggestions? For now, I'm going to leave this as a draft PR.


Jakobeha:
If you want to cleanup the macros, my suggestion would be to add an enum `VerifyStrength` and parse it as an env variable. You could also add an option to run the verifier at random after some passes, so we catch cases like the GVN without causing as much slowdown.

rir/src/compiler/translations/rir_2_pir/rir_2_pir_compiler.cpp
```diff
@@ -203,6 +203,12 @@ void Rir2PirCompiler::optimizeModule() {
 
 #ifdef ENABLE_SLOWASSERT
                 Verify::apply(v);
+#else
+#ifdef FULLVERIFIER
```
o-, line 210: this might be a bit confusing, since SLOWASSERT which is less strict than FULLVERIFIER actually takes precedence.

I would reorder the cases.

o- (approved):
lgtm modulo comment. imho no need to overthink this.

mhyee:
@o-, thanks. Just requesting a quick sanity check, you can merge if you think it's OK.

I reorganized when the verifier is run. Here is the summary:

 . | FULLVERIFIER | ENABLE_SLOWASSERT (Debug) | Release | NDEBUG (ReleaseNoAssert)
----|------------------------|---------------------------------|---------|---------------------------
**After RIR to PIR** | full | fast | fast | x
**After each PIR pass** | full | fast | x | x
**After all PIR opts** | [1] | full | fast | x
**Before PIR to RIR** | full | fast | x | x

[1] Not run, because it's subsumed by running the full verifier after each PIR phase.

---

> If you want to cleanup the macros, my suggestion would be to add an enum VerifyStrength and parse it as an env variable.

I personally feel this kind of thing should be a compile-time option. We can revisit this idea if we start adding more "levels" of verifier strength.

> You could also add an option to run the verifier at random after some passes, so we catch cases like the GVN without causing as much slowdown.

We have enough CI time to run the full verifier always. I'd also prefer these kinds of checks to be deterministic, so it's easier to debug.

o-:
awesome

PR #484, opened 2019-05-08
Serialize and Deserialize RIR expressions

Jakobeha:
- Call `rir.serialize(SEXP, "path")` or `SEXP <- rir.deserialize("path")` to serialize/deserialize an `SEXP`, and preserve compiled closures. In the future this will probably be more aggressive than R's `serialize`/`deserialize`, e.g. it will serialize package environments
- If `RIR_PRESERVE=1`, R will serialize compiled functions on exit, and reload them so they're still compiled
- If `RIR_SERIALIZE_CHAOS=n`, on every `n`th `rirCall`, R will serialize/deserialize the dispatch table before invoking the function, also serializing compiled closures. This tests the serialization, e.g. `RIR_SERIALIZE_CHAOS=1 ./bin/gnur-make-tests check-all`


rir/src/ir/BC.cpp
```diff
@@ -156,6 +157,282 @@ SEXP BC::immediateConst() const {
...
+            break;
+        case Opcode::deopt_: {
+            DeoptMetadata* meta =
+                DeoptMetadata::deserialize(code, refTable, inp);
+            size_t size =
+                sizeof(DeoptMetadata) + meta->numFrames * sizeof(FrameInfo);
+            SEXP store = Rf_allocVector(RAWSXP, size);
+            memcpy(DATAPTR(store), meta, size);
+            i.pool = Pool::insert(store);
+            delete meta;
```
JanJecmen, line 262: This was allocated by `malloc`, so it shouldn't be deallocated by `delete`

rir/src/runtime/Code.cpp
```diff
@@ -61,6 +76,71 @@ unsigned Code::getSrcIdxAt(const Opcode* pc, bool allowMissing) const {
...
+    for (unsigned i = 0; i < code->srcLength; i++) {
+        code->srclist()[i].pcOffset = InInteger(inp);
+        code->srclist()[i].srcIdx =
+            src_pool_add(globalContext(), ReadItem(refTable, inp));
+    }
+    SEXP store = Rf_allocVector(EXTERNALSXP, size);
+    memcpy(DATAPTR(store), code, size);
+    Code* old = code;
+    code = (Code*)DATAPTR(store);
+    delete old;
```
JanJecmen, line 107: Same here, `malloc` and `delete` mismatch

JanJecmen (commented):
not a full review yet, just putting it here so I don't forget

rir/src/runtime/Code.cpp
Jakobeha: `free` caused segfaults where the code body got corrupted, I have no idea how

JanJecmen: Well but this is UB: https://isocpp.org/wiki/faq/freestore-mgmt#mixing-malloc-and-delete

rir/src/ir/BC.cpp
```diff
@@ -156,6 +157,282 @@ SEXP BC::immediateConst() const {
...
+            i.staticCallFixedArgs.versionHint =
+                Pool::insert(ReadItem(refTable, inp));
+            break;
+        case Opcode::deopt_: {
+            DeoptMetadata* meta =
+                DeoptMetadata::deserialize(code, refTable, inp);
+            size_t size =
+                sizeof(DeoptMetadata) + meta->numFrames * sizeof(FrameInfo);
+            SEXP store = Rf_allocVector(RAWSXP, size);
+            memcpy(DATAPTR(store), meta, size);
```
JanJecmen, line 260: I was thinking, could you do this without the copying? Something like have the `deserialize` function return the already finished sexp? Then you would get rid of the `malloc`ing and `free`ing completely

rir/src/ir/BC.cpp
```diff
@@ -156,6 +157,282 @@ SEXP BC::immediateConst() const {
...
+            i.pool = Pool::insert(store);
+            free(meta);
```
JanJecmen: One way is to use the function `operator new` instead of the operator `new`:
In `DeoptMetadata::deserialize`:
```
DeoptMetadata* res = (DeoptMetadata*) ::operator new(size);
```
And here:
```
::operator delete(meta);
```

JanJecmen: And for `Code` as well

JanJecmen (commented):
I think there is some more memory leaking, there are vectors in the extra info for calls... But not sure yet how it gets leaked, in gdb the destructors get actually called

rir/src/ir/BC.cpp
Alice761: I tried before but it crashes. The difficulty is that the R GC can run while the code body is deserialized, but if the code is in an SEXP the GC will try to look through it, including looking through the body. I'll try to set the gc length to 0

Jakobeha:
I forgot I was logged in the wrong account (for a class project) for the above comments.

rir/src/ir/BC.cpp
Jakobeha: I forgot I was logged in the wrong account for the above comment.

PR #487, opened 2019-05-10
Hoist LdVars only and fix unsoundness when the loop may change the environment

charig:


rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -9,19 +9,29 @@ namespace pir {
...
+                        safe = false;
+                });
+                return safe;
+            }
+        }
 
         if (CallSafeBuiltin::Cast(i))
             return true;
 
         if (CallInstruction::CastCall(i))
```
o-: not needed (included in check below)

rir/src/compiler/opt/loop_invariant.cpp
```diff
@@ -9,19 +9,29 @@ namespace pir {
 
 bool isSafeToHoistLoads(const LoopDetection::Loop& loop) {
     auto noEnvironmentTainting = [](Instruction* i) {
-        if (LdFun::Cast(i))
+        // For this instructions we test later they not change the particular
+        // binding
+        if (StVar::Cast(i) || StVarSuper::Cast(i) || MkEnv::Cast(i))
```
o-, line 14: you could combine those two checks into one, by just putting the code from below here

rir/tests/pir_check.R
```diff
@@ -9,10 +9,10 @@ stopifnot(pir.check(
   f <- function(){
     j <- 0
     while (j < 2) {
-    c(j)
+    vector("integer",0)
```
o-, line 12: why doesn't it work with `c`? this should be a builtin safe on non-objects, not?

rir/src/compiler/test/PirCheck.cpp
```diff
@@ -135,9 +135,9 @@ static bool testOneNot(ClosureVersion* f) {
     return numNots == 1;
 }
 
-static bool testLdFunInFirstBB(ClosureVersion* f) {
+static bool testLdVarInFirstBB(ClosureVersion* f) {
     for (auto instruction : *f->entry) {
-        if (LdFun::Cast(instruction)) {
+        if (LdVar::Cast(instruction)) {
```
o-: I am afraid that this is not a very solid check, since it will trigger for any ldvar (which there's almost guaranteed to be one). Can you do sth like `testLdVarCInFirstBB` which tests for `ldvar "c"`

rir/src/compiler/opt/loop_invariant.cpp
o-: actually CallSafeBuiltin::Cast(i) is also included in check below

rir/tests/pir_check.R
charig: `c` uses ldFun

rir/src/compiler/opt/loop_invariant.cpp
charig: At least at the moment most of the loops will not be safe to hoist. So I prefer to do a general test first, and then eventually if there are `ldVars` to hoist a particular test regarding only the binding.

rir/src/compiler/test/PirCheck.cpp
```diff
@@ -135,10 +135,11 @@ static bool testOneNot(ClosureVersion* f) {
     return numNots == 1;
 }
 
-static bool testLdVarInFirstBB(ClosureVersion* f) {
+static bool testLdVarVectorInFirstBB(ClosureVersion* f) {
     for (auto instruction : *f->entry) {
-        if (LdVar::Cast(instruction)) {
-            return true;
+        if (auto ldvar = LdVar::Cast(instruction)) {
+            return std::string(CHAR(PRINTNAME(ldvar->varName)))
```
o-: variable names are symbols. should be compared by:
`ldvar->varName == Rf_install("vector")`

o- (approved):
lgtm modulo symbol comparison as string

PR #489, opened 2019-05-10
Type Assertions + Fixes

Jakobeha:
When RIR_CHECK_PIR_TYPES=1 (on debug), after every PIR instruction with a non-void type is converted into RIR bytecode, an assert_type_ opcode will be emitted with the instruction's type. This will assert that the top of stack SEXP actually has the instruction's type.

This makes an ugly dependency on PIR, since the RIR interpreter accesses PIR types. It doesn't create cycles, because `PirType` itself doesn't have much dependencies. A better solution might be to move PirType outside of PIR, maybe into the ir folder.

---

Also includes 1 fix, which were causing the assertions to fail:
- Some builtins `is.xxx` actually return logical vectors on vector args

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -143,5 +143,6 @@ cache:
   ccache: true
   apt: true
   directories:
+    - external
     # cache homebrew formulas
     - /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core
--- a/.travis.yml
+++ b/.travis.yml
@@ -143,6 +143,5 @@ cache:
   ccache: true
   apt: true
   directories:
-    - external
     # cache homebrew formulas
     - /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core
--- a/rir/src/common.h
+++ b/rir/src/common.h
@@ -12,11 +12,9 @@ extern void printBacktrace();
 
 #ifdef ENABLE_SLOWASSERT
 #define SLOWASSERT(what) assert(what)
-__attribute__((unused)) static bool IS_SLOWASSERT = true;
 #else
 #define SLOWASSERT(what)                                                       \
     {}
-__attribute__((unused)) static bool IS_SLOWASSERT = false;
 #endif
 
 // from boost
--- a/rir/src/common.h
+++ b/rir/src/common.h
@@ -12,9 +12,11 @@ extern void printBacktrace();
 
 #ifdef ENABLE_SLOWASSERT
 #define SLOWASSERT(what) assert(what)
+__attribute__((unused)) static bool IS_SLOWASSERT = true;
 #else
 #define SLOWASSERT(what)                                                       \
     {}
+__attribute__((unused)) static bool IS_SLOWASSERT = false;
 #endif
 
 // from boost
--- a/rir/src/compiler/opt/type_speculation.cpp
+++ b/rir/src/compiler/opt/type_speculation.cpp
@@ -47,17 +47,9 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
                                                 type.isA(RType::real))
                                                    ? type
                                                    : i->type.notObject();
-                            bool redundant = !Visitor::check(
-                                function->entry, [&](Instruction* i2) {
-                                    if (auto c2 = CastType::Cast(i2)) {
-                                        if (c2->arg<0>().val() == i &&
-                                            c2->type == specType)
-                                            return false;
-                                    }
-                                    return true;
-                                });
-                            if (!redundant)
-                                speculate[cp][i] = specType;
+                            speculate[cp][i] = specType;
+                            // Prevent redundant speculation
+                            i->typeFeedback = i->type;
                         }
                     }
                 }
--- a/rir/src/compiler/opt/type_speculation.cpp
+++ b/rir/src/compiler/opt/type_speculation.cpp
@@ -43,12 +43,21 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
                 if (!type.isVoid() && !i->type.isA(type)) {
                     if (auto cp = checkpoint.next(i)) {
                         if (!type.maybeObj()) {
-                            if (type.isA(RType::integer) ||
-                                type.isA(RType::real)) {
-                                speculate[cp][i] = type;
-                            } else {
-                                speculate[cp][i] = i->type.notObject();
-                            }
+                            PirType specType = (type.isA(RType::integer) ||
+                                                type.isA(RType::real))
+                                                   ? type
+                                                   : i->type.notObject();
+                            bool redundant = !Visitor::check(
+                                function->entry, [&](Instruction* i2) {
+                                    if (auto c2 = CastType::Cast(i2)) {
+                                        if (c2->arg<0>().val() == i &&
+                                            c2->type == specType)
+                                            return false;
+                                    }
+                                    return true;
+                                });
+                            if (!redundant)
+                                speculate[cp][i] = specType;
                         }
                     }
                 }
--- a/rir/src/compiler/opt/types.cpp
+++ b/rir/src/compiler/opt/types.cpp
@@ -43,7 +43,7 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                     i->eachArg([&](Value* v) {
                         if (i->mayHaveEnv() && v == i->env())
                             return;
-                        inferred = inferred | getType(v).orNotPromise();
+                        inferred = inferred | getType(v).notPromise();
                     });
                     return inferred;
                 };
--- a/rir/src/compiler/opt/types.cpp
+++ b/rir/src/compiler/opt/types.cpp
@@ -43,7 +43,7 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                     i->eachArg([&](Value* v) {
                         if (i->mayHaveEnv() && v == i->env())
                             return;
-                        inferred = inferred | getType(v);
+                        inferred = inferred | getType(v).orNotPromise();
                     });
                     return inferred;
                 };
--- a/rir/src/compiler/opt/types.cpp
+++ b/rir/src/compiler/opt/types.cpp
@@ -65,6 +65,10 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                     if (i->tag == Tag::Div && inferred.isA(RType::integer)) {
                         inferred = inferred | RType::real;
                     }
+                    if (i->tag == Tag::Add && inferred.isA(RType::logical)) {
+                        // TRUE + TRUE = 2
+                        inferred = inferred | RType::integer;
+                    }
                     break;
                 }
                 case Tag::CallSafeBuiltin: {
--- a/rir/src/compiler/opt/types.cpp
+++ b/rir/src/compiler/opt/types.cpp
@@ -43,7 +43,7 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                     i->eachArg([&](Value* v) {
                         if (i->mayHaveEnv() && v == i->env())
                             return;
-                        inferred = inferred | getType(v).notPromise();
+                        inferred = inferred | getType(v);
                     });
                     return inferred;
                 };
--- a/rir/src/compiler/opt/types.cpp
+++ b/rir/src/compiler/opt/types.cpp
@@ -87,6 +87,15 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                         break;
                     }
 
+                    static const std::unordered_set<std::string> vecTests = {
+                        "is.na", "is.nan", "is.finite", "is.infinite"};
+                    if (vecTests.count(name)) {
+                        inferred = PirType(RType::logical);
+                        if (getType(c->arg(0).val()).isScalar())
+                            inferred.setScalar();
+                        break;
+                    }
+
                     static const std::unordered_set<std::string> tests = {
                         "is.vector",   "is.null",      "is.integer",
                         "is.double",   "is.complex",   "is.character",
@@ -95,9 +104,7 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                         "is.raw",      "is.object",    "isS4",
                         "is.numeric",  "is.matrix",    "is.array",
                         "is.atomic",   "is.recursive", "is.call",
-                        "is.language", "is.function",  "is.single",
-                        "is.na",       "is.nan",       "is.finite",
-                        "is.infinite"};
+                        "is.language", "is.function",  "is.single"};
                     if (tests.count(name)) {
                         inferred = PirType(RType::logical).scalar();
                         break;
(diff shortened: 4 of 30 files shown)


.travis.yml
```diff
@@ -143,6 +143,5 @@ cache:
   ccache: true
   apt: true
   directories:
-    - external
```
o-: this should not be permanently removed

rir/src/common.h
```diff
@@ -12,9 +12,11 @@ extern void printBacktrace();
 
 #ifdef ENABLE_SLOWASSERT
 #define SLOWASSERT(what) assert(what)
+__attribute__((unused)) static bool IS_SLOWASSERT = true;
```
o-: that's not a good idea, to basically duplicate this. either we make it a static variable or a macro, but not both! also I think it's not needed, see comment below

rir/src/compiler/analysis/scope.cpp
```diff
@@ -136,6 +136,8 @@ AbstractResult ScopeAnalysis::doCompute(ScopeAnalysisState& state,
     } else if (Deopt::Cast(i)) {
         // who knows what the deopt target will return...
         state.returnValue.taint();
+        state.mayUseReflection = true;
```
o-: please document the dependencies between your PR's. this is from #499 right?

rir/src/compiler/opt/types.cpp
```diff
@@ -104,9 +104,8 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                     }
 
                     if ("c" == name) {
-                        inferred = mergedArgumentType();
-                        if (c->nCallArgs() > 1)
-                            inferred.setNotScalar();
+                        inferred =
+                            mergedArgumentType().collectionType(c->nCallArgs());
```
o-: again, this is from another PR.

rir/src/compiler/pir/type.cpp
```diff
@@ -1,11 +1,17 @@
...
 #include "R/r.h"
 
 extern "C" Rboolean(Rf_isObject)(SEXP s);
 
 namespace rir {
 namespace pir {
 
+bool Parameter::RIR_CHECK_PIR_TYPES =
+    getenv("RIR_CHECK_PIR_TYPES") ? (bool)atoi(getenv("RIR_CHECK_PIR_TYPES"))
+                                  : IS_SLOWASSERT;
```
o-: I find this very unexpected. please only enable it when RIR_CHECK_PIR_TYPES is set.

rir/src/compiler/pir/type.cpp
```diff
@@ -75,16 +80,14 @@ void PirType::merge(SEXPTYPE sexptype) {
...
     case S4SXP:
         t_.r = val().t_.r;
+        break;
+    default:
+        break;
     }
 }
 
 PirType::PirType(SEXP e) : flags_(defaultRTypeFlags()), t_(RTypeSet()) {
-    if (e == R_MissingArg)
```
o-: why did you remove this. it is definetly needed. don't remove it!! we have special types for missing and unbound! they are typed as symbol in pir.

rir/src/compiler/pir/type.cpp
```diff
@@ -123,5 +131,22 @@ void PirType::merge(const ObservedValues& other) {
         merge(record.sexptype);
     }
 }
+
+bool PirType::hasInstance(SEXP val) const {
```
o-: isInstance?

rir/src/compiler/pir/type.cpp
```diff
@@ -123,5 +131,22 @@ void PirType::merge(const ObservedValues& other) {
...
+    if (isRType()) {
+        if (TYPEOF(val) == PROMSXP)
+            return maybePromiseWrapped() || maybeLazy() ||
+                   PirType(RType::prom).isA(*this);
+        if (LazyEnvironment::cast(val))
+            return PirType(RType::env).isA(*this);
+        return PirType(val).isA(*this | RType::missing);
```
o-: why `| missing`? doesn't make sense. if val is the missing value, then the type should allow for missingness, otherwise it is an error.

o-: can you please also check scalar and objectness?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -919,6 +919,19 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                       needsEnsureNamed.count(instr)))
                 cb.add(BC::ensureNamed());
 
+            // Check the return type
```
o-, line 944: why do you only check the return type?

can you have different check levels, and on level 2 you also check let's say binops, or call builtin arguments/results

o- (commented):
hey, this is going into a good direction. I added some comments that still need to be worked out. 

please in the future: if you have dependencies between PRs, but want them to be reviewed in parallel, then document the dependencies. and make it such that it is very clear, which commits belong to which PR, e.g. by rewriting the branch history such that each PR only has one commit. Reviewing PR takes a lot of time. If you request one, it is polite to try to minimize the amount of work needed for the reviewer.

I would like to not couple these checks with SLOWASSERT, as mentioned in the inline comments. if you remove that, please make sure to add the checks to at least one testsuite run and at least one `gnur-make-tests check` run in .gitlab-ci.yaml, such that the tests are still being run.

Jakobeha:
For your first comment, should I do similar to before where I added a commit `#499` (302f2at), but then make sure future changes (47907d8) are also in that commit?

rir/src/compiler/analysis/scope.cpp
Jakobeha: yes

rir/src/compiler/pir/type.cpp
Jakobeha: Wouldn't these already be handled though? e.g. `PirType(<object>).isA(<non-object type>) => <object type>.isA(<non-object type>) => false`

o-:
> For your first comment, should I do similar to before where I added a commit `#499` (302f2at), but then make sure future changes ([47907d8](https://github.com/reactorlabs/rir/commit/47907d8742dd8e87a3490d119309080a47bd11b9)) are also in that commit?

sorry I don't understand.

what I would would like is the following:

```
PR1
  description: "This PR does blablabla"
  * commit x

PR2
  description: "This PR does blablabal. It is based on PR1, only commit y needs to be reviewed"
  * commit x
  * commit y
```

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
Jakobeha: I think arg types are already checked, because the verifier checks after each pass that each arg's return type conforms to its instructions' expected arg type, and when the arg instruction is executed it must conform to its return type. I'm not sure what you mean by builtin results

Jakobeha:
Is this OK?

```
PR123
  description: "..."
  * commit a
  * commit b
  * commit c
PR456
  description: "..."
  * commit a (#123)
  * commit d
  * commit b (#123)
  * commit c (#123)
  * commit e
```

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
o-: > each arg's return type conforms to its instructions' expected arg type, 

but it could still be wrong. that's just the static type of the arguments. this pr is all about verifying that the static type actually corresponds to the dynamic type.

so e.g. if you have `a+b` and we think `a` is an int, but it isn't actually, we might not notice. so it would be useful to check it.

> I'm not sure what you mean by builtin results

if you have a callBuiltin and we somehow think it returns a certain type, it would be nice to check that it actually does. same with binops and so on

o-:
> Is this OK?

I don't know how to review that on github..

you can use `git rebase -i` to reorder and combine commit into one

Jakobeha:
Ok, I was using `git rebase` before but sometimes it messes up the history (with merges), so I used `git cherry-pick`. I’ll go back to `rebase -i`

o-:
you can't just pile new stuff on top of an already reviewed huge PR. this takes me forever to get through, since I don't know what is existing and what is new. 

I am skeptical of your new modeling of missing in the type system. if you want to merge that, I want to review it in isolation. not gonna merge it like this

Jakobeha:
Ok, I rolled back to before I implemented the missing change, and added a "short circuit" which makes the assertions pass even on missing values.

I still have a different type fix (6145867 `some tests return vectors`), but that's only a single file and very straightforward. The only changes are that fix, the ability to print out bad instructions with `RIR_CHECK_PIR_TYPES=2`, and your feedback.

o-:
but didn't we agree that the proper workaround before fixing the missing flag is, to change `PirType::forced()` to add `RType::missing`?

Jakobeha:
Ok I did that. I thought it caused some tests to fail before, but it seems to be working now (maybe the verifier?).

o-:
@Jakobeha can you rebase this ontop of the current master? because there was some problem that not all tests would be run before

rir/src/compiler/opt/type_speculation.cpp
```diff
@@ -43,12 +43,21 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
...
-                                type.isA(RType::real)) {
-                                speculate[cp][i] = type;
-                            } else {
-                                speculate[cp][i] = i->type.notObject();
-                            }
+                            PirType specType = (type.isA(RType::integer) ||
+                                                type.isA(RType::real))
+                                                   ? type
+                                                   : i->type.notObject();
+                            bool redundant = !Visitor::check(
```
o-: we absolutely want to address the actual bug and not just put some ad-hoc test there that prevents it from popping up.

I am not quite sure I understand the problem.

Is it that we speculate twice on the same value? then maybe the fix is to just remove the typefeedback once we speculated. e.g. on line 87 say `t->typefeedback = i->type`

rir/src/compiler/opt/types.cpp
```diff
@@ -87,6 +87,15 @@ void TypeInference::apply(RirCompiler&, ClosureVersion* function,
                         break;
                     }
 
+                    static const std::unordered_set<std::string> vecTests = {
+                        "is.na", "is.nan", "is.finite", "is.infinite"};
+                    if (vecTests.count(name)) {
+                        inferred = PirType(RType::logical);
+                        if (getType(c->arg(0).val()).isScalar())
+                            inferred.setScalar();
```
o-, line 99: hah, good catch

rir/src/compiler/pir/type.cpp
```diff
@@ -75,6 +80,13 @@ void PirType::merge(SEXPTYPE sexptype) {
     case WEAKREFSXP:
     case S4SXP:
         t_.r = val().t_.r;
+        t_.r.reset(RType::missing);
+        t_.r.reset(RType::unbound);
```
o-: this makes no sense at all.

If you look at the constructor `PirType(SEXP)`, then it does a number of additional things already, including that.

rir/src/compiler/pir/type.h
```diff
@@ -326,11 +327,15 @@ struct PirType {
...
+    PirType forced() const {
         assert(isRType());
-        FlagSet notPromised =
-            ~(FlagSet() | TypeFlags::promiseWrapped | TypeFlags::lazy);
-        return PirType(t_.r, flags_ & notPromised);
+        PirType res = *this;
+        if (res.maybePromiseWrapped()) {
+            res.flags_.reset(TypeFlags::promiseWrapped);
+            res.flags_.reset(TypeFlags::lazy);
+            res.t_.r.set(RType::missing);
```
o-: can you put a small comment explaining why we set the missing flag

rir/src/compiler/pir/type.h
```diff
@@ -326,11 +327,15 @@ struct PirType {
         return PirType(t_.r, flags_ & ~FlagSet(TypeFlags::lazy));
     }
 
-    PirType constexpr forced() const {
+    PirType forced() const {
```
o-: please leave this constexpr. you need to use a constexpr to create the type. to add the missing flag it should be enough to just add `...(t_.r | RType::missing, ...)` to the expression that was there

rir/src/ir/BC_inc.h
```diff
@@ -108,6 +109,19 @@ class BC {
         Immediate expected;
         Immediate id;
     };
+    struct AssertTypeArgs {
+        static_assert(
+            sizeof(pir::PirType) == sizeof(Immediate) * 2,
+            "PirType must fit in 2 immediates, or change assert_type_ size");
+        Immediate typeData1;
+        Immediate typeData2;
+        const char* instr;
```
o-: please no pointers to the C heap in the bytecode stream! you should know why, when you try to serialize RIR code, this will be a pain.

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -919,6 +919,29 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+                    return true;
+                })) {
+                const char* instrStr;
+                if (pir::Parameter::RIR_CHECK_PIR_TYPES > 1) {
+                    std::stringstream instrPrint;
+                    instr->print(instrPrint, false);
+                    // WARNING: Leaks memory
+                    instrStr = (const char*)strdup(instrPrint.str().c_str());
+                } else {
+                    instrStr = "not generated, set RIR_CHECK_PIR_TYPES=2";
```
o-: you realize that this will pass a pointer to a temporary value to the bytecode?

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -919,6 +919,29 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
...
+                            return false;
+                    }
+                    return true;
+                })) {
+                const char* instrStr;
+                if (pir::Parameter::RIR_CHECK_PIR_TYPES > 1) {
+                    std::stringstream instrPrint;
+                    instr->print(instrPrint, false);
+                    // WARNING: Leaks memory
+                    instrStr = (const char*)strdup(instrPrint.str().c_str());
```
o-: let's not do that. you can use an R string here and put it into the constant pool.

```
SEXP msg = Rf_mkString(debugMessage.c_str());
unsigned idx = Pool::insert(msg);
```

o-: then store that pool index in the bytecode stream

rir/src/compiler/opt/type_speculation.cpp
Jakobeha: There are 2 bugs. One is a problem in the CFG, where we replace one of the instructions with the speculated type in a BB that isn't preceded by the cast, so the verifier fails with "instruction used before it's defined". That isn't addressed here.

There's also redundant type-speculation, which is fixed by this. This doesn't cause any unsoundness itself, just redundancy. Although if I fix this bug it sidesteps the first bug.

I definitely think we need to address the first bug, and I can look at it later, just not in this PR.

o-: ok, but please solve it by not scanning the whole instruction stream. this has like a O(n^2) complexity. removing the feedback should be ok. or, since you are iterating the whole instruction stream here anyways, it could also be just removed again, once we arrive at the redundancy anyway.

PR #490, opened 2019-05-10
Liveness analysis: handle liveness at bb start

mhyee:
Also add print debugging (behind a compile-time switch)

---

This is another thing I need for loop peeling, but I'm splitting it off because it can stand on its own (even if nothing uses it), and it'll make merging/reviewing the loop peeling branch easier when it's ready.

Per-instruction output looks like this:
```
========== Liveness info for instr:
real$'          %0.0  = LdConst                [1] 1
        Live in BB0: [0, 4)
        Live in BB1: ![0, 8)
        Live in BB2: ![0, 7)
        Live in BB6: ![0, 11)
```

### DIFF
--- a/rir/src/compiler/analysis/liveness.cpp
+++ b/rir/src/compiler/analysis/liveness.cpp
@@ -91,10 +91,13 @@ LivenessIntervals::LivenessIntervals(unsigned bbsSize, CFG const& cfg) {
         assert(pos == 0);
 
         // Mark everything that is live at the beginning of the BB.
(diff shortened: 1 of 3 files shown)


charig (approved):
Why we debug to std::cerr?

mhyee:
> Why we debug to std::cerr?

Don't remember exactly... I guess I wanted to keep it separate from the actual program output, and there may have been issues with flushing the output buffer when I was using gdb...

Issue #488, opened 2019-05-10
RIR 2 PIR ungracefull crash 


charig:
When compiling RIR to PIR, in case the compiler fails when computing the `mergepoints`, the compiler end up in a kind of infinite loop and consumes huges amounts of memory finally crashing.

It would be great if the compilary failed gracefully instead.

To repro just run: `tools:::.check_code_usage_in_package(package = "compiler")`.

To ensure there is a missing mergepoint under that code just remove the condition: `std::get<0>(m) == first ||` in `findMergepoints` (rir_2_pir.cpp).

See #477.

o-:
can't reproduce anymore

PR #491, opened 2019-05-10
Dirty fix for #488

Jakobeha:


charig (approved):
I was thinking 2 things:
* if `hangCounter++;` is inside an assert that may help in terms of documentation for the future (it is a debugging check)
* Isn't 1000000 to high?

Jakobeha:
I tested on my local machine and 1000000 triggered the error when there was still vm left.

mhyee:
How long did it take to trigger the error? One reason for a smaller limit is to catch the error earlier.

Jakobeha:
Not more than about 5secs, I dropped it to 100000

o-:
static variables are shared between all invocations. this method is used for compiling promises, while compiling the actual function. thus the counter is always reset when you compile a promise. i verified that it fails to catch the original error. please test your changes. also I don't like the approach, something like a hang counter should be really last resort, after we properly analyzed the problem and convinced ourselves that an actual solution is infeasible.

PR #493, opened 2019-05-13
Revert "Dirty fix for #488"

o-:
Reverts reactorlabs/rir#491


o-:
see https://github.com/reactorlabs/rir/pull/491#issuecomment-491706096

PR #496, opened 2019-05-13
package rir as a container

o-:
this changes how we run ci jobs on gitlab. the strategy is:

1. have a base image with all apt packages preinstalled
2. build a pir container based on (1)
3. push (2) tagged with the commit id to the registry
4. run all tests in parallel with the container from (3)
5. (on master) push the container with :master tag, if all tests succeed

this should shave roughly 12' from each ci job since we only need to compile once. And additionally this gives us a container we can use as a base for other things. We can use it on the benchmark machine, we can build artifacts on top, and we can use it for distributing rir.


baseContainer/README.mk
```diff
@@ -0,0 +1,15 @@
+## What
```
mhyee: This file should be using the `.md` extension.

mhyee (commented):
I don't understand Docker, so do we need to do anything about this?

```
$ echo "$CI_BUILD_TOKEN" | docker login -u "$CI_BUILD_USER" --password-stdin registry.gitlab.com
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

Also, what's the retention policy on containers? It seems like every CI build will produce a new container, but are there any limits? Do we want some sort of garbage collection run?

mhyee:
Another thing: the containers are still over 1GB in size...

o-:
@mhyee sorry for changing this while you are already reviwing. I added a second commit that should build a benchmark container, which contains all the infra for running benchmarks.

the advantage is that the benchmark container is guaranteed to work...

o-:
and yes the container is still big. I know. but since it is now layered on top of a base image, that contains all the apt dependencies, the actual data you need to pull when you already have the base container is actually quite small. something like 200mb

o-:
> I don't understand Docker, so do we need to do anything about this?

we build the container inside a container on our runner. so this file just vanishes when the jobs ends, no problem.

> Also, what's the retention policy on containers? 

Idk if github will just remove old ones at some point.

> It seems like every CI build will produce a new container

yes every job creates a new container. then the tests use that container.

> , but are there any limits? Do we want some sort of garbage collection run?

we could delete upon completion, but then the failed runs would still be lying around. And that is also super convenient, because in the future you can just do

    docker run ...:commit-id

and have locally the exact same container as on the ci.

I looked a bit at the solutions, but there is nothing really convenient. I will probably add a script we can run manually, that just deletes everything but master.

PR #497, opened 2019-05-14
Pass blacklist option

o-:
add an option to blacklist compiler passes. also move the pass
scheduling from utils/configurations (that name never made sense to me)
to compiler/opt/pass_scheduler and remove the ini file configuration,
which we never used and is not in sync with the actual passes anymore.


mhyee:
Yeah, the configurations and ini file were bugging me too. I'm also glad to see modern C++ goodness like `unique_ptr`.

PR #498, opened 2019-05-14
Deopt must change the return type

Jakobeha:
Scope analysis infers call return types (same as #492, it was already implemented). When it encounters a `Deopt` it taints the local state, the problem is this doesn't always affect the final state.


o-:
did you find an example where this breaks?

Jakobeha:
Yes, via #489, `.getGenericImplicit`’s return type was inferred to `nil’`, but it actually deoptimized and returned another closure. I  think it only happens in “atypical” cases though, `.getGenericImplicit` was very weird because it deoptimized on a true branch and the last BB (after the deopt) contained the return

o-:
ok, thanks, lets merge it for now then and figure out if we can improve later.

PR #500, opened 2019-05-16
add dockerfile and scripts to create a benchmark-baseline container with vanilla-r

o-:
just a dockerfile to create a container for what used to be vanilla-r.
this container is identical to the benchmark kind of containers, except that rir is not built and the r is built without any rir modifications.


o-:
I canceled the CI, since this is for manually building and publishing anyway. But I already built and published the container at: registry.gitlab.com/rirvm/rir_mirror/benchmark-baseline

charig (approved):

PR #501, opened 2019-05-17
Change in order of paramenters to run.sh (first mandatory then optional)

charig:


charig:
I disabled all the ci jobs but the benchmarks

PR #502, opened 2019-05-20
Cleanup, speedup and fix bugs in the bindings cache

o-:
looked at the slowdowns and found a lot to cleanup

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -676,7 +676,6 @@ class FLIE(LdFun, 2, Effects::Any()) {
 class FLIE(LdVar, 1, Effects() | Effect::Error | Effect::ReadsEnv) {
   public:
     SEXP varName;
-    bool usesCache = true;
 
     LdVar(const char* name, Value* env)
         : FixedLenInstructionWithEnvSlot(PirType::any(), env),
@@ -788,7 +787,6 @@ class FLIE(LdVarSuper, 1, Effects() | Effect::Error | Effect::ReadsEnv) {
 class FLIE(StVar, 2, Effect::WritesEnv) {
   public:
     bool isStArg = false;
-    bool usesCache = true;
 
     StVar(SEXP name, Value* val, Value* env)
         : FixedLenInstructionWithEnvSlot(PirType::voyd(), {{PirType::val()}},
--- a/rir/src/interpreter/cache.h
+++ b/rir/src/interpreter/cache.h
@@ -1,56 +1,104 @@
 #ifndef RIR_INTERPRETER_CACHE_H
 #define RIR_INTERPRETER_CACHE_H
 
+#include "R/r.h"
 #include "instance.h"
 
 namespace rir {
 
-// must be a power of 2 for modulus using & CACHE_MASK to work
-#define MAX_CACHE_SIZE 128
-#define CACHE_MASK (MAX_CACHE_SIZE - 1)
+// makes max cache size fit into 64bytes (with size)
+#define MAX_CACHE_SIZE 63
 #define ACTIVE_BINDING_MASK (1 << 15)
 #define BINDING_LOCK_MASK (1 << 14)
 #define IS_ACTIVE_BINDING(b) ((b)->sxpinfo.gp & ACTIVE_BINDING_MASK)
 #define BINDING_IS_LOCKED(b) ((b)->sxpinfo.gp & BINDING_LOCK_MASK)
+#define FRAME_LOCK_MASK (1 << 14)
+#define FRAME_IS_LOCKED(e) (ENVFLAGS(e) & FRAME_LOCK_MASK)
 
 #ifdef CACHE_ON_R_STACK
 // TODO: Create a version with a cache on the R_stack instead of the C stack
 #else
-typedef struct {
-    SEXP loc;
-} BindingCache;
-typedef BindingCache Cache;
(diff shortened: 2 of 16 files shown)


o-:
this fixes the regression in storage benchmark (surpassing the version before cache changes). not sure about the rest, I guess we'll have to see.

the delayed clearing is a *big* slowdown in this benchmark (not exactly sure why, though I have theories).

rir/src/compiler/pir/instruction.h
```diff
@@ -676,7 +676,6 @@ class FLIE(LdFun, 2, Effects::Any()) {
 class FLIE(LdVar, 1, Effects() | Effect::Error | Effect::ReadsEnv) {
   public:
     SEXP varName;
-    bool usesCache = true;
```
charig, line 679: The problem with this change and the related in the allocator is that these fields were here for later usage by PIR. 
We can have an analyses that decides that some ldVars and stVars do not require all the cache handling (for instance single ldvars/stvars outside loops, etc)

rir/src/interpreter/cache.h
```diff
@@ -1,56 +1,104 @@
...
 
+#include "R/r.h"
 #include "instance.h"
 
 namespace rir {
 
-// must be a power of 2 for modulus using & CACHE_MASK to work
-#define MAX_CACHE_SIZE 128
-#define CACHE_MASK (MAX_CACHE_SIZE - 1)
+// makes max cache size fit into 64bytes (with size)
```
charig: I do not get the comment?

rir/src/interpreter/cache.h
```diff
@@ -1,56 +1,104 @@
...
-static RIR_INLINE void cachedSetBindingCell(Immediate cacheIdx, Cache* cache,
-                                            bool smallCache, R_varloc_t loc) {
-    cache[cacheIdx].loc = loc.cell;
+static RIR_INLINE void
+cachedSetBindingCell(Immediate cacheIdx, BindingCache* cache, R_varloc_t loc) {
+    SLOWASSERT(cacheIdx < cache->length);
+    cache->entry[cacheIdx] = loc.cell;
+}
+
+static void rirDefineVarWrapper(SEXP symbol, SEXP value, SEXP rho) {
```
charig, line 47: Why do we need this wrapper?

rir/src/interpreter/interp.cpp
```diff
@@ -1472,25 +1472,17 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
...
         bindingCache = cache;
     } else {
-        bindingCache =
-            (BindingCache*)(alloca(sizeof(BindingCache) * cacheSize));
-        /*
-         * In case we are entering the executing of a PIR optimized function
-         * we delay the cache cleaning until we know we are actually requiring
-         * an environment (mk_env).
-         */
-        if (env != symbol::delayedEnv)
```
charig, line 1490: This saves one cache clearing in every pir function activation. Why remove it?

rir/src/compiler/pir/instruction.h
o-: yeah, but this is perfectly possible externally. I don't like unused code for later use, really. half of what i removed here was for your predecessor PR but then not removed.

rir/src/interpreter/cache.h
o-: the cache datastructure now also contains the size. because reading out the size from c was fishy (think about the case where c changes but the cache not...)

rir/src/interpreter/cache.h
o-: the main slowdown in storage is from the fact that you use INCREMENT_NAMED on things which don't need them

rir/src/interpreter/interp.cpp
o-: because it is faster (see my comment)

o-: my guess is clearing right after alloca is faster than later

rir/src/interpreter/cache.h
charig: Yes, that I got it. But what max cache size fit into 64 bytes means in this context?

o-: 64bytes = L1 cacheline

rir/src/interpreter/interp.cpp
charig: This is a sensible guess. But the thing is that we are not cleaning now and not cleaning later. We are doing both. The only difference is when a PIR function, before creating an env changes the environment. Maybe we can just do something with those functions so that we do the cleaning at startup.

rir/src/interpreter/cache.h
charig: My concern is, we were executing the `INCREMENT_NAMED` previously. I did not change the semantics of the variable set, did I?

o-: yes, so in the cached lookups we had case that would check if the thing that we are overriding is the same. in this case you do not need to increment.

with your changes all the non-caching implementations lost that trick and they started incrementing named counts like crazy.

you probably didn't notice since you set the cache limit so high, it would never trigger (also seen by the fact that both the normal compilar and pir optimizer had several bugs, when the cache size would be reached...)

rir/src/interpreter/interp.cpp
o-: I know that we are cleaning twice. was still faster.

PR #503, opened 2019-05-21
Selective cache clear

o-:
based on #502 , only clear cache idx which belong to a particular mkenv.

### DIFF
--- a/rir/src/compiler/translations/pir_2_rir/allocators.h
+++ b/rir/src/compiler/translations/pir_2_rir/allocators.h
@@ -473,6 +473,9 @@ class CachePosition {
             auto e = Env::Cast(env.first);
             if (e && e->rho != R_BaseEnv && e->rho != R_BaseNamespace) {
                 for (const auto& key : env.second) {
+                    // If a binding is used only once, then it does not pay off
+                    // to cache (yes, it could be used in a loop, this is just a
+                    // static approximation).
                     if (key.second <= 2)
                         continue;
                     if (uniqueNumbers.size() == MAX_CACHE_SIZE)
@@ -483,9 +486,11 @@ class CachePosition {
         }
         globalEnvsCacheSize_ = uniqueNumbers.size();
 
-        // Heuristic: after how many loads does it pay off to cache a variable?
-        // It depends on the size of the environment, since the smaller the
-        // environment, the faster the lookup.
+        // At runtime looking up a binding in a local environment is a linear
+        // search. Therefore, the smaller then environment, the faster the
+        // lookup. E.g. looking up in an env of size one is not much slower than
+        // going through the cache. Therefore we will scale the limit on the
+        // size of the environment.
         auto minAccessEnvSize = [](size_t s) -> unsigned {
             if (s == 0)
                 return 99999; // This env seems empty, caching is just a waste.
@@ -497,7 +502,7 @@ class CachePosition {
         };
         for (const auto& env : found) {
             if (!Env::Cast(env.first)) {
-                envCachePositions[env.first].first = uniqueNumbers.size();
+                envCacheRanges[env.first].first = uniqueNumbers.size();
                 auto limit = minAccessEnvSize(env.second.size());
                 for (const auto& key : env.second) {
                     if (key.second <= limit)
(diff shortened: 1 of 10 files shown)


rir/src/ir/BC_inc.h
```diff
@@ -121,6 +121,10 @@ class BC {
         PoolIdx poolIndex;
         CacheIdx cacheIndex;
     };
+    struct CacheIdxs {
```
charig: CacheRange?

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -451,26 +451,88 @@ class CachePosition {
...
+                        break;
+                    uniqueNumbers.emplace(key.first, uniqueNumbers.size());
+                }
+            }
+        }
+        globalEnvsCacheSize_ = uniqueNumbers.size();
+
+        // Heuristic: after how many loads does it pay off to cache a variable?
+        // It depends on the size of the environment, since the smaller the
+        // environment, the faster the lookup.
```
charig: Why the smaller the faster? It is a hash table...
Oth, the code defines bigger limits for smaller environment
And then, why for non-global environments we cache even if there is only one load while for global we do not?
Also how the heuristic treat loop lds is unclear...

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -451,26 +451,88 @@ class CachePosition {
...
+                return 99999; // This env seems empty, caching is just a waste.
+            else if (s == 1)
+                return 3;
+            else if (s < 5)
+                return 2;
+            return 1;
+        };
+        for (const auto& env : found) {
+            if (!Env::Cast(env.first)) {
+                envCachePositions[env.first].first = uniqueNumbers.size();
```
charig: envCacheRanges?

rir/src/compiler/translations/pir_2_rir/allocators.h
```diff
@@ -451,26 +451,88 @@ class CachePosition {
...
+            }
+        });
+
+        // First all global envs
+        for (const auto& env : found) {
+            auto e = Env::Cast(env.first);
+            if (e && e->rho != R_BaseEnv && e->rho != R_BaseNamespace) {
+                for (const auto& key : env.second) {
+                    if (key.second <= 2)
+                        continue;
```
charig, line 480: What if it is inside a loop?

rir/src/compiler/translations/pir_2_rir/allocators.h
o-: > Why the smaller the faster? It is a hash table...

no, its a list. lookup is a linear search

> Oth, the code defines bigger limits for smaller environment

yes, the smaller the env, the more lookups are needed to win anything by caching

> And then, why for non-global environments we cache even if there is only one load while for 
> global we do not?

well if you think about it, for local envs the number of bindings we statically see is roughly equal to the size of the environment. but for a global env, that is certainly not true, as the global env will have many more unrelated bindings. so there this heuristic does not work.

> Also how the heuristic treat loop lds is unclear...

well it doesn't, but in practice it still seems to work quite nicely

rir/src/compiler/translations/pir_2_rir/allocators.h
o-: i know, still a somewhat useful proxy.

rir/src/compiler/translations/pir_2_rir/allocators.h
charig: I didn't get it. You were talking about the standard lookup. I understood it as the binding cache lookup.
I would add to the comment ...the smaller the env, the faster the (standard|noncached) lookup. 
Oth, this is not taking into account that the lookup could need to go through a hierarchy of envs.

charig (approved):

rir/src/compiler/translations/pir_2_rir/allocators.h
o-: > I would add to the comment ...the smaller the env, the faster the (standard|noncached) lookup.

ok, I'll make it more clear

> Oth, this is not taking into account that the lookup could need to go through a hierarchy of envs.

sure, but then we don't cache anyway.

PR #506, opened 2019-05-24
elide env when forcing promises

o-:
The gnu R implementation of forcePromise sets the environment to nil
after it has been evaluated.

In case we statically evaluate a promise we should therefore be able to
elide the environment from the start.


o-:
tbr

PR #507, opened 2019-05-24
use stub envs for calls, as well as local loads and stores 

o-:


o-:
tbr

rir/src/interpreter/LazyEnvironment.h
```diff
@@ -18,37 +19,29 @@ namespace rir {
  * environment lazily.
  */
 struct LazyEnvironment
```
charig, line 21: We discussed this when the `LazyEnvironment` was introduced. The `RuntimeObject` may be tidier but it involves creating a SEXP which is not much different than just creating an environment.

o-: the current approach

1. ... was broken, since it leaked memory. all envs created where collected in a vector but then never deleted, which I already mentioned the last time we talked about it.
2. ... does not work for calls, because env stub might outlive the function call.
3. ... adds overhead for the tracking of envstubs

so it's not a matter of taste, but a matter of necessity. the current use of stubs was super limited.

> creating a SEXP which is not much different than just creating an environment.

no. full env means allocate env, and linked list for bindings (ie. 1+n allocations) whereas this is just 1 allocation. I am pretty sure that the current pushing of all env stubs onto a vector (which in term allocates and grows some backing store) is slower than the allocation of the one heap object here.

also the main advantage is in the assumptions the compiler can make about stub envs (bindings are stable) and not the runtime overhead for creating them.

charig: If there was a leak that was solvable. 
For the rest, I disagree with the necessity. 
Anyway, I guess with this approach we can rollback some of these changes in gnu-r: https://github.com/reactorlabs/gnur/commit/c575dedb863481e6b6e904525a44a0fc2d65e7a6#diff-d564936611d9b4c8e598252d86c13559R1714

PR #512, opened 2019-05-29
Fix gte/lte mixup.

mhyee:
Somehow we made the same mistake in rir_2_pir and pir_2_rir, so the bug
cancelled itself out.

However, Constantfold did not have the bug, so some branches got
incorrectly optimized out.

(This was originally fixed by @o- but not merged yet, but I need this
fix.)

---

I'm merging this once CI passes.


o-:
sorry for not backporting this, I thought I had, but probably forgot

PR #513, opened 2019-06-12
Uses fixed-len array instead of vector to store arg types

Jakobeha:
In `FunctionSignature`, we create vectors but don't explicitly `delete` them (since the signatures are destroyed by the R GC). This PR uses a fixed-length array of 4 arguments, and ignores the rest (currently signature arg types seem to only be used for printing)

### DIFF
--- a/rir/src/runtime/FunctionSignature.h
+++ b/rir/src/runtime/FunctionSignature.h
@@ -102,6 +102,9 @@ struct FunctionSignature {
         return numArguments - assumptions.numMissing();
     }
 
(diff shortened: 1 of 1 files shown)


JanJecmen (commented):
It should be easier to use `std::array`, then you get pretty much the same api as `std::vector`

Other than that, I think this is ok for now (otherwise the signatures and the arguments would have to be exposed to the collector). Maybe add a comment to explain that?

Jakobeha:
subsumed by serialization

PR #516, opened 2019-06-25
cleanup and unify type inference

o-:
unify all the type inference logic in Intruction, let the type inference
pass reuse the logic from the instruction.

cleanup and disentangle effect and type inference.

fix type inference for binops (did cast away object flag). Fix
regressions by adding binops to speculated on instructions.

### DIFF
--- a/rir/src/compiler/test/PirCheck.cpp
+++ b/rir/src/compiler/test/PirCheck.cpp
@@ -41,7 +41,6 @@ static ClosureVersion* recompilePir(SEXP f, Module* m) {
         []() { Rf_warning("pir check failed: couldn't compile"); });
 
     cmp.optimizeModule();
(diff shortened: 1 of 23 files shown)


rir/src/compiler/test/PirCheck.cpp
```diff
@@ -41,7 +41,6 @@ static ClosureVersion* recompilePir(SEXP f, Module* m) {
         []() { Rf_warning("pir check failed: couldn't compile"); });
 
     cmp.optimizeModule();
-    cmp.optimizeModule(); // TODO: Why is this needed twice?
```
mhyee, line 44: Note that this was copied from `PirTests.cpp`, which still has two calls to `optimizeModule`.

PR #523, opened 2019-06-27
Less serialization tests

Jakobeha:
Also see if this makes `test_big_inline` faster

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -75,8 +75,8 @@ test_types_and_serialize:
     - RIR_CHECK_PIR_TYPES=1 bin/gnur-make-tests check-all
     - ../../tools/check-gnur-make-tests-error
     - RIR_SERIALIZE_CHAOS=1 bin/tests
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -75,8 +75,8 @@ test_types_and_serialize:
     - RIR_CHECK_PIR_TYPES=1 bin/gnur-make-tests check-all
     - ../../tools/check-gnur-make-tests-error
     - RIR_SERIALIZE_CHAOS=1 bin/tests
-    - RIR_SERIALIZE_CHAOS=1 bin/gnur-make-tests test-all-basics
     - RIR_SERIALIZE_CHAOS=3 bin/tests
+    - RIR_SERIALIZE_CHAOS=$(shuf -i 55-75 -n 1) bin/gnur-make-tests check-all
```
o-: if you do that, please also print the value to the console, so it is possible to reproduce a failure. see PIR_DEOPT_CHAOS_SEED for an example

PR #528, opened 2019-07-01
forced promises does not need the env anymore

o-:


o-:
actually the constructor does that already

PR #531, opened 2019-07-02
test serialization only on ./bin/tests, and fix memory leaks

Jakobeha:
this should use much less RAM than gnur-tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -98,7 +98,7 @@ tests_fullverify:
     - bin/tests
     - bin/gnur-make-tests check-devel
 
-# Test particular features, like deoptimization
+# Test particular features, like deoptimization and serialization
 test_features:
   image: registry.gitlab.com/rirvm/rir_mirror:$CI_COMMIT_SHA
   variables:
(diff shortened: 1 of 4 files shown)


o-:
last time I checked bin/tests used massive ammounts of ram. please investigate the source or confirm that it does not!

And do that before pushing, because this clogs our testrunner machine!

Jakobeha:
I checked on prl5, `RIR_SERIALIZE_CHAOS=1 ./bin/tests` and `RIR_SERIALIZE_CHAOS=10 ./bin/tests` both take about 5mins. `RIR_SERIALIZE_CHAOS=3` seemed to be taking longer so I changed it to `=10`. On GitLab the entire `test_features` took 36 minutes. I haven't checked RAM, but I think if it isn't taking too long and the other GitLab tests aren't taking longer RAM is OK too. 

I was running `RIR_SERIALIZE_CHAOS=1 ./bin/gnur-make-tests` before, which meant that every function call in the entire GNUR test suite, the function would be copied by serialization/deserialization. Also, some data in the copied `RirRuntimeObjects` is probably leaking, so that could explain the high RAM. This runs `./bin/tests`, so there are still a lot of copies and leaks, but much less (this could explain why it takes 5mins while `./bin/tests` without serialization takes 1min).

o-:
So this `RIR_SERIALIZE_CHAOS=1 ./bin/tests` uses 1.5 Gb of ram.

There are big leaks here, and I would like to see them addressed. You can use ` -d valgrind --debugger-args=--leak-check=full` as arguments to R to find them.

o-:
see https://github.com/reactorlabs/rir/pull/535, that should have fixed the issue from decodeInstruction

PR #540, opened 2019-07-09
fix printing abstract unbound values

Jakobeha:


rir/src/compiler/analysis/abstract_value.cpp
```diff
@@ -80,8 +80,12 @@ void AbstractPirValue::print(std::ostream& out, bool tty) const {
     for (auto it = vals.begin(); it != vals.end();) {
         auto vo = *it;
         vo.val->printRef(out);
-        out << "@";
-        vo.origin->printRef(out);
+        if (vo.origin != NULL) {
```
o-: this is a bit dangerous, since we might at some point have other abstract values that have no origin instruction. a more conservative approach would be to check for `val == Unbound::instance()`

Jakobeha: What about in `abstract_value.h` `assert(val == Unbound::instance() || origin != NULL);`

o-: sounds reasonable. thanks

PR #567, opened 2019-07-24
Loop peeling (in RIR)

mhyee:
The idea of loop peeling is to unroll a loop once, so that the unrolled loop (sometimes called a "peel") is a sequence of instructions with no backwards jump. Then it executes the real loop, which has backwards jump.

The idea is that the first iteration of the loop allows things to initialize (e.g. types, values, etc.) without any control flow merges, so that static analysis has an easier time with the loop body.

I implemented this at the AST level, in the RIR compiler, and it works for `for`, `repeat`, and `while` loops. It also works with the "simple `for` loops" which avoid vector allocations. (The previous attempt at the PIR level was horrendously complicated, because it's lower level and in SSA.) I have a check to only peel innermost loops.

There are two drive-by commits: e760e823326e1bc7bb22272738c6cf5ff573d94b is a minor thing to clean up error messages, and ed7d037d83888014e31e50b89fdccf3eb5685c7b is pretty substantial.

---

I'm opening this PR to get CI to run, and I plan to merge it tomorrow, maybe after people have had a chance to look at it. (I'm not asking for a full code review.) That way, I can get the benchmarks to run and see if there's any regression.

I'm interested in whether loop peeling helps with optimizations, and whether compilation times have increased. (If compilation times increase, then I'll probably have a condition to only peel loops that are smaller than a certain threshold.)

---

If anything goes wrong after merging, we can set `Compiler::loopPeelingEnabled` to false and disable a few tests.


mhyee:
I waited too long and now the CI backlog is very long, so I'll just merge this now and get a spot in the queue...

PR #577, opened 2019-07-26
Force speculation

JanJecmen:


rir/src/compiler/pir/type.cpp
```diff
@@ -121,13 +121,6 @@ void PirType::merge(const ObservedValues& other) {
         return;
     }
 
-    if (other.numTypes == ObservedValues::MaxTypes) {
```
o-, line 124: I don't understand why this is removed?

o-: I am merging as is, but I am pretty sure that this needs to come back. The reason is we stop recurding after MaxTypes, therefore we don't know what kind of other types are flowing through and we have to assume the worst. otherwise this might lead to deopt loops.

JanJecmen: Oh, I was confused for a bit why I removed this, too.. but this block is a duplicate of the one just above. So it's only a cleanup

o-: ah, yeah, that was hidden by the diff view. thanks

PR #580, opened 2019-07-29
Fix serialize and re-add tests

Jakobeha:
Some changes to the native backend broke serialization (we don't serialize native code). I also stopped testing serialization for `pir_regression_check_code` because that by far uses the most memory.

Serialization doesn't "technically" leak a lot of memory anymore, but it creates extra constant pool entries which are never destroyed. But not testing `pir_regression_check_code`, I believe it won't use more than 1GB of memory at once (the second longest test, `pir_regression7`, used 150MB with `RIR_SERIALIZE_CHAOS=1`. `pir_regression_check_code` was using almost 4GB)


o-:
@Jakobeha I wanted to merge this, but after rebasing it now fails

PR #581, opened 2019-07-29
fix dead store anaylsis for promises

Jakobeha:
A promise can load a variable in the parent environment, so that the last store isn't actually "dead".

### DIFF
--- a/rir/src/compiler/analysis/dead_store.h
+++ b/rir/src/compiler/analysis/dead_store.h
@@ -15,23 +15,25 @@ class DeadStoreAnalysis {
       protected:
         SubAnalysis() {}
 
-        AbstractResult applyCommon(State& state, Instruction* i) const {
-            AbstractResult effect;
+        bool applyRecurse(AbstractResult& effect, State& state,
+                          Instruction* i) const {
             if (auto force = Force::Cast(i)) {
                 if (auto mk = MkArg::Cast(force->input()->followCasts())) {
                     if (mk->usesPromEnv()) {
                         // The promise could get forced here and
                         // use variables in the parent environment
                         effect.max(
-                            recurse(state, i, mk->prom(), mk->promEnv()));
+                            handleRecurse(state, i, mk->prom(), mk->promEnv()));
+                        return true;
                     }
                 }
             }
-            return effect;
+            return false;
         }
 
-        virtual AbstractResult recurse(State& state, Instruction* i,
-                                       Promise* prom, Value* env) const = 0;
+        virtual AbstractResult handleRecurse(State& state, Instruction* i,
+                                             Promise* prom,
+                                             Value* env) const = 0;
     };
 
     // 1. keep track of LdFunctionEnv aliases.
@@ -65,11 +67,13 @@ class DeadStoreAnalysis {
         }
 
         AbstractResult apply(AliasSet& state, Instruction* i) const override {
-            return applyCommon(state, i);
+            AbstractResult effect;
+            applyRecurse(effect, state, i);
+            return effect;
         }
 
-        AbstractResult recurse(AliasSet& state, Instruction* i, Promise* prom,
-                               Value* env) const override {
+        AbstractResult handleRecurse(AliasSet& state, Instruction* i,
+                                     Promise* prom, Value* env) const override {
             AbstractResult res;
 
             auto ld = LdFunctionEnv::Cast(*prom->entry->begin());
@@ -139,6 +143,7 @@ class DeadStoreAnalysis {
       protected:
         AbstractResult apply(EnvSet& state, Instruction* i) const override {
             AbstractResult effect;
+            applyRecurse(effect, state, i);
             if (i->leaksEnv()) {
                 Value* env = aliases.resolveEnv(i->env());
                 if (auto mk = MkEnv::Cast(env)) {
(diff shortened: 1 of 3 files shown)


o-:
@Jakobeha do you have a repro for this issue?

It seems to me that the root cause must be somewhere else. Because if you look at the analysis, there is a case for `i->readsEnv()` which covers both `Force` and `CallInstruction`. If an instruction reads its env, then the whole env and all parents should be marked as observed. So don't see how your change could get more environments marked as observed?

Jakobeha:
The problem is, the `Force` has it's env elided even though it uses `LdFunctionEnv`, so `readsEnv` is false.

If the forced promise uses `LdFunctionEnv`, does it perform reflection, and can its env be elided? The problem is that if the force still has an env, PIR fails to inline and do other optimizations in some of the tests.

The repro occurs in a branch where I prevent pir from inserting any assumptions. The problematic code is the function in:

```r
stopifnot(pir.check(function() {
  a <- function() 41L
  b <- function() 1L
  f <- function(x, y) x() + y
  f(a, b())
}, Returns42L, warmup=function(f)f()))
```

If the force has an environment, the above function works but the check fails.

o-:
ok, I see. thanks for the explanation.

so basically the problem is: if mkarg and force are in the same function, then it can happen that the env of mkarg is marked as not leaked, and therefore the force with elided environment will not do anything to the state, even if it has `i->effects.contains(Effect::ExecuteCode)`.

In the case of CallInstruction I don't think this does anything, since anyway the env of the call instruction will be marked as leaked and then all stores are observable by the call (since it executes more code than just the promises that are passed).

So basically what you want is that in the case of Force, where you statically see the MkArg, you want to recurse the analysis into the promise code.

1. in the recursive analysis of the promise you only handle the `auto ld = LdVar::Cast(i)`. this might fix your particular issue, but for the general case you must handle all case, ie. you must recursively call the full analysis loop on the promise. (e.g. the promise could also have a force instruction that does this and so on).
2. Even if you now specially handle the Force instruction still triggers `(i->exits() || i->effects.contains(Effect::ExecuteCode))` in your PR, which means it is a lot less precise than it needs to be. basically observing all leaked envs, even though we should be more precise now.

also a comment describing the issue would be nice and a testcase.

o-:
please don't merge this. What does this have to do with the original bug you found and why didn't you fix that one instead?

Jakobeha:
I thought this was what you meant by “in the case of force when you statically know the MkArg recurse into the promise.” I don’t know what the right solution is, the original was too narrow but of course this is too broad and prevents further optimization.

I guess we need to move inlining MkArg/Force into scope resolution. There’s a cycle where the force can’t be inlined because it needs the env, and the env can’t be removed unless the force is inlined. Should I try this?

o-:
> I thought this was what you meant by “in the case of force when you statically know the MkArg recurse into the promise.”

I guess we didn't really understand each other.  What I meant to say was that in your initial commit here: https://github.com/reactorlabs/rir/commit/c950d82bf303d899fbf44083189f38dbfa226d83#diff-e88bc7b1c611f9591b665127cfa4ec71R191 you analyze the promise code. but in the promise code you only apply one part of the dead store analysis (namely the case for LdVar). An my point was, shouldn't you apply all cases of the dead store analysis to the promise code (like the suff we do in case of stvar, readsEnv, and so on).

> I guess we need to move inlining MkArg/Force into scope resolution. 

I don't think this has anything to do with scope analysis

o-:
ping me on slack if it's not clear

Jakobeha:
I think I understand now, see my new changes. I made it so we run entire sub-analyses on the promise code though, I don't know if this is necessary or if I coded it right.

o-:
that looks great! I have one suggestions, that you might want to add here or in a later PR

rir/src/compiler/analysis/dead_store.h
```diff
@@ -177,15 +293,28 @@ class DeadStoreAnalysis {
                     effect.update();
                 }
             } else if (i->readsEnv()) {
-                observeFullEnv(i->env());
+                observeFullEnv(aliases.resolveEnv(i->env()));
             } else if (i->exits() || i->effects.contains(Effect::ExecuteCode)) {
```
o-, line 299: 1. I think there is an existing bug here, because if an instruction has both both ExecuteCode and readsEnv, then probably the ExecuteCode case should apply first, since it is the one with more effects.
2. The `Force` instruction you handle in applyCommon has both readsEnv and ExecuteCode, but since you now handle it with a sub-analysis, we don't need to be pessimistic anymore. It would be nice if you can somehow not doe the observeFullEnv iff the sub analysis analyzes the promise.

Jakobeha: For 2. I think if a `Force` forces a statically known promise, instead we can just remove the ExecuteCode.

o-: oh no. Granted we know what we are executing, still we are executing more code... I think of this flag as sort of a stand-in for any effects the code that gets executed could have...

o-: and still it also does not fix the fact that we are still doing the readsEnv case to a Force, that we actually handle in applyCommon instead

PR #592, opened 2019-08-01
Fix force speculation

JanJecmen:
Attempt at fixing #577

### DIFF
--- a/rir/src/compiler/opt/type_speculation.cpp
+++ b/rir/src/compiler/opt/type_speculation.cpp
@@ -31,9 +31,9 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
                 !i->typeFeedback.type.isVoid() &&
                 !i->type.isA(i->typeFeedback.type) &&
                 !(i->type & i->typeFeedback.type).isVoid()) {
-                switch (i->tag) {
-                case Tag::Force: {
-                    auto arg = i->arg(0).val()->followCasts();
+
+                if (auto force = Force::Cast(i)) {
+                    auto arg = force->input()->followCasts();
                     // Blacklist of where it is not worthwhile
                     if (!LdConst::Cast(arg) &&
                         // leave this to the promise inliner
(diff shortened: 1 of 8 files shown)


JanJecmen:
@o- This doesn't propagate the feedback to the argument of force, it instead marks the force with a flag and during type speculation temporarily adds the force's type feedback to its argument's. This should avoid the issue completely, if I understand it correctly?

As a side note, I remember you saying that the checkpoint selection in type speculation was wrong (shouldn't use `next()`) but I cannot recall what was the conclusion (and why it ended up working when that part wasn't changed)

rir/src/compiler/opt/type_speculation.cpp
```diff
@@ -45,13 +45,18 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
...
-                        i = Instruction::Cast(i->arg(0).val());
-                    break;
-                }
-                default: {}
+                    if (force->observed != Force::ArgumentKind::unknown &&
+                        force->observed != Force::ArgumentKind::promise &&
+                        Instruction::Cast(force->input())) {
+                        i = Instruction::Cast(force->input());
+                        if (force->observed == Force::ArgumentKind::value)
+                            i->typeFeedback = force->typeFeedback;
```
o-: drive by comment: wouldn't this do the exact same thing as the previous try, just a bit later? if multiple force have the same input then this still leads to wrong typefeeback.

JanJecmen:
Update: the `while vector` benchmark times out now, probably because the vector is copied in every iteration (we have `subassign` - named = 0, then `ensure named` - named = 1, then `store` - named = 2)

Didn't get further in investigating this

o-:
that means that the result of subassign is consumed twice in pir.

do you see speedups?

o-:
#607 might help. actually wanted to do that already a while ago

JanJecmen:
Looks like that could help
Didn't check performance yet, just worked to get it fixed

PR #595, opened 2019-08-05
flag to disable assumptions in compiled code

Jakobeha:
Adds a flag `RIR_UNSOUND_ASSUME`, which prevents PIR from emitting assumptions when closures are compiled. If we get to a point where PIR would deopt, instead we simply crash. However, this enables PIR to get extra optimizations.

The purpose is, you could run this on library code (e.g. internal functions) which you know won't receive unexpected arguments. We can also serialize this code then deserialize it in sound execution, so we have a library with internal functions which are faster than PIR could normally generate.

Supersedes / rebased on #581

### DIFF
--- a/rir/src/compiler/transform/bb.cpp
+++ b/rir/src/compiler/transform/bb.cpp
@@ -1,5 +1,6 @@
 #include "bb.h"
 #include "../analysis/dead.h"
+#include "../parameter.h"
 #include "../pir/pir_impl.h"
 #include "../util/visitor.h"
 #include "R/Funtab.h"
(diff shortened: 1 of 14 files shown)


rir/src/compiler/transform/bb.cpp
```diff
@@ -195,22 +196,35 @@ BB* BBTransform::lowerExpect(Code* code, BB* src, BB::Instrs::iterator position,
...
-
-void BBTransform::insertAssume(Value* condition, Checkpoint* cp,
-                               bool assumePositive) {
+Assume* BBTransform::insertAssume(Instruction* condition, Checkpoint* cp,
+                                  BB* bb, BB::Instrs::iterator& position,
+                                  bool assumePositive) {
+    if (Parameter::RIR_UNSOUND_ASSUME) {
+        // don't actually insert the assume
+        delete condition; // we don't actually need it
+        return NULL;
```
o-, line 205: but this might silently produce a wrong result, which is not the same as "crashing"

PR #598, opened 2019-08-06
Limit debug output

JanJecmen:
Only print passes that change the code.

This is super inefficient, it stores a string with the last pass output and then buffers the current pass and compares before writing it out.. But for debugging it should be fine.


o-:
how about patching a "changed" boolean through?

That would be useful since we could use this to early abort the optimization pipeline if certain passes don't change anything anymore.

(but not arguing against merging this one, if you need it now)

JanJecmen:
I actually started with that, but then decided for this version to get it quickly. It was annoying to go through optimizations and setting the flag - the one I did needed it at like 4 places

PR #599, opened 2019-08-07
Bk llvm subassing

charig:


rir/src/compiler/native/lower_llvm.cpp
```diff
@@ -2454,16 +2508,133 @@ bool LowerFunctionLLVM::tryCompile() {
...
+                                load(i, extract->idx(), Representation::Real),
+                                t::i64);
+                        } else {
+                            index = builder.CreateZExt(
+                                load(i, extract->idx(),
+                                     Representation::Integer),
+                                t::i64);
+                        }
+                    } else {
+                        // Maybe not needed if we know it is scalar?
```
o-, line 2548: yes, isScalar() corresponds to the R level IS_SIMPLE_SCALAR macro. so actually isScalar => !maybeObject(). (If you see a type that is scalar but also maybeObject, that sounds wrong. We should probably have this as an assert in the types).

PR #603, opened 2019-08-07
Load elision type fix

Jakobeha:
Something which was affecting my promise recording / sandboxing branch

### DIFF
--- a/rir/src/compiler/opt/load_elision.cpp
+++ b/rir/src/compiler/opt/load_elision.cpp
@@ -47,8 +47,13 @@ struct AvailableLoads : public StaticAnalysis<IntersectionSet<ALoad>> {
         AbstractResult res;
         if (auto ld = LdVar::Cast(i)) {
             for (auto& ald : state.available) {
-                if (ald.same(ld))
-                    return AbstractResult::None;
+                if (ald.same(ld)) {
+                    if (ld->type.isA(ald.origin->type)) {
+                        ald.origin->type = ld->type;
+                        res.update();
+                    }
+                    return res;
+                }
             }
             state.available.insert(ld);
             res.update();
--- a/rir/src/compiler/opt/load_elision.cpp
+++ b/rir/src/compiler/opt/load_elision.cpp
@@ -15,10 +15,13 @@ namespace pir {
 
 struct ALoad {
     // cppcheck-suppress noExplicitConstructor
-    ALoad(LdVar* ld) : origin(ld), env(ld->env()), name(ld->varName) {}
+    ALoad(LdVar* ld)
+        : origin(ld), type(ld->type), env(ld->env()), name(ld->varName) {}
     // cppcheck-suppress noExplicitConstructor
-    ALoad(LdFun* ld) : origin(ld), env(ld->env()), name(ld->varName) {}
+    ALoad(LdFun* ld)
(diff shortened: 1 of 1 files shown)


o-:
can you please explain?

it seems like you are changing the type of the actual load instruction in an analysis?

analysis must not change the code. only analyze

Jakobeha:
Say we have:

```
val %1 = LdVar a, e0
%1.1 = RequiresForced %1
...
val~ %2 = LdVar a, e0
```

and `%2` dominates `%1`. If the analysis finds `val~ %2` first, then `val %1`, `%1.1`'s arg will be replaced with `val~`, and it will get a type error. If all loads load the same variable, then we know the variable has the most narrow type we could infer.

You're right that the analysis shouldn't modify the type. It actually could've changed the instruction even if the analysis didn't find anything.

o-:
ok, but now you just replace it with a random type (whichever happens to be picked last)...

Jakobeha:
yes that was another typo

o-:
please don't expect me to look so carefully at all the PR's

rir/src/compiler/opt/load_elision.cpp
```diff
@@ -64,12 +67,14 @@ struct AvailableLoads : public StaticAnalysis<IntersectionSet<ALoad>> {
         return res;
     }
 
-    Instruction* get(Instruction* i) const {
+    Instruction* get(Instruction* i, PirType& type) const {
```
o-: ugh, output arguments -- that's really C style -- and then also mixed with normal return values. that not nice. why not return the ALoad?

rir/src/compiler/opt/load_elision.cpp
```diff
@@ -47,8 +56,13 @@ struct AvailableLoads : public StaticAnalysis<IntersectionSet<ALoad>> {
         AbstractResult res;
         if (auto ld = LdVar::Cast(i)) {
             for (auto& ald : state.available) {
-                if (ald.same(ld))
-                    return AbstractResult::None;
+                if (ald.same(ld)) {
+                    if (ld->type.isA(ald.type)) {
+                        ald.type = ld->type;
```
o-, line 61: maybe I am stupid now. But to me this looks like you are just storing the type of whatever instruction the analysis saw last. given the analysis does not have a fixed order this could be fairly random. In the end you might pick up a more precise type or also a less precise one.

If I understand you correctly you wanted to prevent this pass from degrading to a less precise type. Which seems like a reasonable thing to do. So why don't you do just that on the replacement side instead of in the analysis?

Jakobeha: You’re right that I’m trying to store the most precise type. What this should do is, only store the type of the last-seen instruction if it’s a subtype of the current type.

Say the analysis finds `int %1 = Ldvar a` and then `val %2 = LdVar a` and both loads are the same. The `ALoad` will have type `int` when it sees `val %2`, and since `val->isA(int)` is false, the type will stay `int`.

Of course, if one type is `int` and the other is `real` (neither is a subtype), the final type will be that of the first instruction, since neither type is more precise. But I don’t really know how to fix this since we don’t have true intersection types, and this is already what would happen.

o-: i see, I missed the test. ok

PR #604, opened 2019-08-08
Phi placement fix

JanJecmen:
@o- I think this should be fine, don't see why using the dominator tree order matters in this instance. (explanation in commit msg)

### DIFF
--- a/rir/src/compiler/util/cfg.cpp
+++ b/rir/src/compiler/util/cfg.cpp
@@ -200,7 +200,7 @@ const DominanceGraph::BBList& DominanceGraph::dominators(BB* bb) const {
 
 void DominanceGraph::dominatorTreeNext(
     BB* bb, const std::function<void(BB*)>& apply) const {
-    BreadthFirstVisitor::run(bb, [&](BB* b) {
+    DepthFirstVisitor::run(bb, [&](BB* b) {
         if (immediatelyDominates(bb, b))
             apply(b);
     });
--- a/rir/src/compiler/util/visitor.cpp
+++ b/rir/src/compiler/util/visitor.cpp
@@ -13,8 +13,8 @@ void DominatorTreeVisitor<VisitorHelpers::IDMarker>::run(
     todo.push_back(entry);
 
     while (!todo.empty()) {
-        BB* cur = todo.front();
-        todo.pop_front();
+        BB* cur = todo.back();
+        todo.pop_back();
         if (!done.check(cur)) {
             done.set(cur);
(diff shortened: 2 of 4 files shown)


o-:
the implementation is following http://ssabook.gforge.inria.fr/latest/book.pdf (as mentioned in the comment)

what you are changing is the code from algorithm 3.3 which says "foreach BB: basic Block in depth-first search preorder traversal of the dom. tree do"...

rir/src/compiler/util/cfg.cpp
```diff
@@ -200,7 +200,7 @@ const DominanceGraph::BBList& DominanceGraph::dominators(BB* bb) const {
 
 void DominanceGraph::dominatorTreeNext(
     BB* bb, const std::function<void(BB*)>& apply) const {
-    BreadthFirstVisitor::run(bb, [&](BB* b) {
```
o-, line 203: this change is 100% wrong.

if we have
```
  A
 / \
B   C
 \ /
  D
```

then starting from A we want to visit in the order B,C,D. Which you discover by enumerating dominated nodes in breadth-first order of the cfg. This has nothing to do with the traversal order of the dominator tree, which is
```
A
|\ \
B C D
```

rir/src/compiler/util/visitor.cpp
```diff
@@ -13,8 +13,8 @@ void DominatorTreeVisitor<VisitorHelpers::IDMarker>::run(
     todo.push_back(entry);
 
     while (!todo.empty()) {
-        BB* cur = todo.front();
-        todo.pop_front();
+        BB* cur = todo.back();
```
o-, line 16: I think this change is also wrong. Let's take our favourite example:
```
  A
 / \
B   E
|   |
C   |
 \ /
  D
```

the `dominatorTreeNext` function will (before you changed it) give you: B,E,D for the node A.

next we want to visit B, so clearly fifo is what we want here and not lifo (like you changed it here).
But then, to have a depth first traversal we need to insert C to the front.

in short the `dominatorTreeNext` need to be inserted to the front of the todo list, but in the original order. So we need a temp list of the next nodes and then insert this list in the original order to the front of the todo list.

o-:
reverting, that was a bit of a random change, wasn't it...

rir/src/compiler/util/visitor.cpp
o-: or even shorter summary of the bug:
the problem is that using `dom.dominatorTreeNext(cur, [&](BB* bb) { todo.push_back(bb); });`, when you consume from the back, the immediately dominated nodes actually appear in reverse.

JanJecmen:
right, sorry about that

PR #612, opened 2019-08-13
fix deopt loop in pidigits

o-:
the main improvement is that we do not optimize while still
deoptimizing. then, also add some back-off counter that delays
reoptimization for frequently deoptimizing functions.


o-:
combined in #610

PR #646, opened 2019-09-23
avoid copying the abstract state in analysis

o-:


o-:
big inline now runs inside time budget. tested here: https://gitlab.com/rirvm/rir_mirror/-/jobs/301446150

PR #650, opened 2019-09-24
Improved no reflection flag

o-:
detect simple lookup promises to keep non-reflective arguments flag
fixes perf regression

### DIFF
--- a/rir/src/compiler/opt/scope_resolution.cpp
+++ b/rir/src/compiler/opt/scope_resolution.cpp
@@ -19,8 +19,8 @@ using namespace rir::pir;
 
 // Checks (using the local state of the caller) if forcing a promise can have
 // reflective effects
-static bool noReflection(Code* code, Value* callEnv, ScopeAnalysis& analysis,
-                         ScopeAnalysisState& state) {
+static bool noReflection(ClosureVersion* cls, Code* code, Value* callEnv,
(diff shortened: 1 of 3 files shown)


rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -36,41 +36,55 @@ static bool noReflection(Code* code, Value* callEnv, ScopeAnalysis& analysis,
...
-            auto e = ld->env() == funEnv ? callEnv : ld->env();
-            auto res = analysis.loadFun(state, ld->varName, e);
-            if (anyReflection(res.result))
-                return false;
-            return true;
-        }
+            if (auto ld = LdFun::Cast(i)) {
+                auto e = ld->env() == funEnv ? callEnv : ld->env();
+                auto res = analysis.loadFun(state, ld->varName, e);
+                if (anyReflection(res.result))
```
charig, line 72: Shouldn't all these similar `if` be the other way around?

o-: ugh, that's ugly, also doulbe negation. was not a bug per-se, since the caller also used it the wrong way round :) pushed a cleanup directly to master

o-: thanks

PR #657, opened 2019-10-01
refactor type test generation

o-:


rir/src/compiler/opt/type_speculation.cpp
```diff
@@ -16,106 +17,64 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
...
-                    }
-                }
-            }
-            ip = next;
+        }
+
+        if (!speculateOn)
+            return;
+
+        if (auto cp = checkpoint.next(i)) {
```
JanJecmen: For the case, this used to be `auto cp = checkpoint.at(force)` and not `next`

o-: I see. guess `checkpoint.next(speculateOn)` makes the most sense, doesn't it?

o-: the reason I am working on this is, that I saw in matrix-for we added a sub-par test and I wanted to add more types.

so I started with this refactoring, that maybe should have been done earlier...

JanJecmen: Good point.. My reasoning was to not have the checkpoint and the assume too far apart, but I guess this allows to hoist the assume further.

By more types you mean something like `(logical | real)'`?

charig: I wonder whether `checkpoint.next(speculateOn)` could have as a consequence that less checkpoints are available

PR #677, opened 2019-10-11
Elide environments in fastpaths even when a mkarg uses them in a deop…

charig:
…t branch

### DIFF
--- a/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
@@ -768,10 +768,6 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                 auto p = mk->prom();
                 unsigned id = ctx.cs().addPromise(getPromise(ctx, p));
                 promMap[p] = id;
-                if (auto env = MkEnv::Cast(mk->env())) {
-                    if (mk->bb()->isDeopt() && env->stub)
-                        cb.add(BC::materializeEnv());
-                }
                 if (mk->isEager()) {
                     cb.add(BC::mkEagerPromise(id));
                 } else {
@@ -1062,6 +1058,11 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                         cb.add(
                             BC::clearBindingCache(range.first, range.second));
                 });
+                break;
+            }
+
+            case Tag::MaterializeEnv: {
+                cb.add(BC::materializeEnv());
                 break;
             }
 
--- a/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
+++ b/rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
@@ -443,8 +443,6 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
         };
 
         // Only needed for deopt branches
-        bool seenMkArg = false;
-
         for (auto it = bb->begin(); it != bb->end(); ++it) {
             auto instr = *it;
 
@@ -680,8 +678,7 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                 auto key =
                     CachePosition::NameAndEnv(ldvar->varName, ldvar->env());
                 auto mkenv = MkEnv::Cast(ldvar->env());
-                if (mkenv && mkenv->stub &&
-                    (!ldvar->bb()->isDeopt() || !seenMkArg)) {
+                if (mkenv && mkenv->stub) {
                     cb.add(BC::ldvarNoForceStubbed(
                         mkenv->indexOf(ldvar->varName)));
                 } else if (needsLdVarForUpdate.count(instr)) {
@@ -737,8 +734,7 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                 auto mkenv = MkEnv::Cast(stvar->env());
                 assert(!MkEnv::Cast(stvar->arg(0).val()));
                 if (stvar->isStArg) {
-                    if (mkenv && mkenv->stub &&
-                        (!stvar->bb()->isDeopt() || !seenMkArg)) {
+                    if (mkenv && mkenv->stub) {
                         cb.add(
                             BC::stargStubbed(mkenv->indexOf(stvar->varName)));
                     } else if (cache.isCached(key)) {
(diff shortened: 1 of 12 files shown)


rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -733,7 +736,8 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
                 auto mkenv = MkEnv::Cast(stvar->env());
                 assert(!MkEnv::Cast(stvar->arg(0).val()));
                 if (stvar->isStArg) {
-                    if (mkenv && mkenv->stub) {
+                    if (mkenv && mkenv->stub &&
+                        (!stvar->bb()->isDeopt() || !seenMkArg)) {
```
o-: this can't possibly work. think about this case:

```
e = mkenv
checkpoint -> [deopt(mkarg(..., e))]
stvar(..., i)
```

if you visit the deopt branch before the continue branch of the cp, you will emit stvar. if you visit the continue branch before the deopt branch, then you will emit stvar_stubbed.

imho the prefered solution here is to also have this instruction in pir.

then the example I wrote you can express as:

```
e = mkenv
checkpoint -> [e1 = materialize(e); deopt(mkarg(..., e1))]
stvar(..., i)
```

and it will be clear which is which

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
```diff
@@ -757,10 +762,15 @@ rir::Code* Pir2Rir::compileCode(Context& ctx, Code* code) {
...
 
             case Tag::MkArg: {
+                seenMkArg = true;
                 auto mk = MkArg::Cast(instr);
                 auto p = mk->prom();
                 unsigned id = ctx.cs().addPromise(getPromise(ctx, p));
                 promMap[p] = id;
+                if (auto env = MkEnv::Cast(mk->env())) {
+                    if (mk->bb()->isDeopt() && env->stub)
+                        cb.add(BC::materializeEnv());
```
o-: missing in native

rir/src/compiler/translations/pir_2_rir/pir_2_rir.cpp
charig: I do not see how a change in the continue branch could happen. My change applies only on how `stVars` are emitted in deopt branches.

o-: be creative...

Assuming BB's are visited: S,A,B,C,D
```
S:
  e = mkenv
  checkpoint -> A B
A:
  mkarg(..., e)
  ...
  deopt
B:
  checkpoint -> C D
C:
  stvar(..., e)  # this will emit stvar, when it should emit stvar stubbed  
  ...
  deopt
D:
  ...
```

charig: I must be missing sth, but from what I see, the `LoweringVisitor` at some point will `visit C (after S,A,B)` and since `mkenv && mkenv->stub && !stvar->bb()->isDeopt()` it would emit a `stVarStubbed`.

o-: >  it would emit a stVarStubbed.

no it won't since seenMkarg is true

o-: details aside, you can't use a global variable to track a flow-sensitive property...

charig: But...
1. It is an `||` clause. 
2. I am not using a global. `seenMkArg` is a per BB var, it should be restarted on every `bb` as far as I understand the visitor. And the only idea is, if I am in a deopt block, in which an `mkArg` already happened, I should not emit store/load stubs, because the mkArg would have already materialized the env.

o-: >  seenMkArg is a per BB var, 

oh, I missed that.

well then it will break for mkarg in the fastpath (i know, currently not, but ugh, this is brittle) and deopt branches with more than one BB. All our deopts use at least 2 BBs after lower() in the pir2rir.

o-: ignoring if this particular implementation might work with the kind of code we generate, still:

you are tracking a flow sensitive property (is the stubenv materialized), with a condition that is not flow sensitive (one boolean per BB, instead of one boolean per flow).

don't you agree that there is a design flaw here?

charig: Definitely, my assumption was that `deopt` branches consist always on a single `basic block`. If that is not the case, then this will not work

charig: Oth, there are lot of places on the code that uses the `isDeoptBanch` which only looks for a `deopt` instruction in the branch. At least, that is how I've been using it. This opt, only elides an env in case the `mkArg` is inside a single `deoptBranch`, and then the change when lowering `ldVars and stVars` also only applies to that single `block`. 
The opt could be explain as: 
`We can elide environments even if there are `mkArgs` using it but those must only reside inside basic blocks but only if they are deopt branches (contain a deopt isntruction). In that case, in those deopt branches we must ensure that ldVars and stVars after the mkArg instruction (which now materializes de env silently) use the real materialized environment instead of the stub`

o-: currently the following is the case:

before Pir2Rir::lower  :  deopt branches are (probably) always 1BB
after Pir2Rir::lower : deopt branches are always at least 2 BBs (because assume gets converted to branch).

o-: it would be good if we don't rely too much on how exactly the deopt branch looks. I know about the issue with isDeoptBranch. we should fix that method to better deal with mulit-bb deopt branches.

o-: I don't want to include a transformation that relies on the fact that we do not split a BB into two. that is beyond silly.

o-: what would be the problem with adding a materialize env instruction at the pir level? sth. that takes a sutb env as input and returns an env as output. that's probably easier to implement than the time spent thinking about in which situations the current implementation might be correct...

PR #687, opened 2019-10-15
Support for moving burst of instructions only needed for deopt to mul…

charig:
…tiple deopt branches

### DIFF
--- a/rir/src/compiler/opt/delay_instr.cpp
+++ b/rir/src/compiler/opt/delay_instr.cpp
@@ -9,72 +9,87 @@ namespace pir {
 void DelayInstr::apply(RirCompiler&, ClosureVersion* function,
                        LogStream&) const {
 
-    std::unordered_map<Instruction*, BB*> usedOnlyInDeopt;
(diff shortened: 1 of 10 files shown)


rir/src/compiler/opt/delay_instr.cpp
```diff
@@ -9,72 +9,89 @@ namespace pir {
...
-                    if (i->bb()->isDeopt() || usedOnlyInDeopt[i]) {
-                        if (!u)
-                            u = i->bb();
-                        else
-                            u = (BB*)-1;
+    Visitor::run(function->entry, [&](Instruction* instruction) {
+        instruction->eachArg([&](Value* v) {
+            if (auto usage = Instruction::Cast(v)) {
+                if (isTarget(usage) && !usage->bb()->isDeopt())
+                    dataDependencies[usage].insert(instruction);
```
o-, line 23: usage confused me a bit (since to me that alludes more that we are talking about i). sometimes we call the value `arg` and then when we cast it to instruction `argi`

o- (commented):
merging this, so you already have the benchmark number tomorrow to look at...

PR #689, opened 2019-10-17
don't treat error as a deopt barrier

o-:
the reasoning is that errors terminate the function, so no rollback
before the error can happen, thus they can be ignored for assumes.

also this fixes a bug where the assume opt pass would remove too
many cp's since it did skip the first instr of the next bb when
serching for a next cp.

debugging and thinking by @charig

### DIFF
--- a/rir/src/compiler/opt/type_speculation.cpp
+++ b/rir/src/compiler/opt/type_speculation.cpp
@@ -91,25 +91,32 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
         if (!speculate.count(cp))
             return;
 
-        bb = bb->trueBranch();
-
         for (auto sp : speculate[cp]) {
-            auto i = sp.first;
-
-            auto ip = bb->begin();
-            if (i->bb() == bb)
-                ip = bb->atPosition(i) + 1;
-
+            auto speculationTarget = sp.first;
             TypeTest::Info& info = sp.second;
+            // The instruction requiring speculation could be before or after
+            // the cp, so we need to find the place that is after both to insert
+            // the typetest and the assume
+            BB* targetBB;
+            BB::Instrs::iterator insertPosition;
+            targetBB = bb->trueBranch();
+            if (speculationTarget->bb()->beforeInCfg(targetBB)) {
+                insertPosition = targetBB->begin();
+            } else {
+                if (targetBB->beforeInCfg(speculationTarget->bb()))
+                    targetBB = speculationTarget->bb();
+                insertPosition = targetBB->atPosition(speculationTarget) + 1;
+            }
 
-            BBTransform::insertAssume(info.test, cp, bb, ip, info.expectation,
-                                      info.srcCode, info.origin);
+            BBTransform::insertAssume(info.test, cp, targetBB, insertPosition,
+                                      info.expectation, info.srcCode,
+                                      info.origin);
 
-            auto cast = new CastType(i, CastType::Downcast, PirType::any(),
-                                     info.result);
+            auto cast = new CastType(speculationTarget, CastType::Downcast,
+                                     PirType::any(), info.result);
             cast->effects.set(Effect::DependsOnAssume);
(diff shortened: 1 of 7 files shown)


rir/src/compiler/pir/bb.cpp
```diff
@@ -182,7 +182,11 @@ bool BB::before(Instruction* a, Instruction* b) const {
     }
     assert(false);
     return false;
-};
+}
+
+// TODO: More robust version. Should be based on real cfg. But on some analysis
+// it is expensive to compute it only for this behavior.
+bool BB::beforeInCfg(BB* otherBB) const { return this->id < otherBB->id; }
```
o-: we currently have no such guarantees currently, that does not work

rir/src/compiler/opt/type_speculation.cpp
```diff
@@ -91,25 +91,32 @@ void TypeSpeculation::apply(RirCompiler&, ClosureVersion* function,
...
+            BB* targetBB;
+            BB::Instrs::iterator insertPosition;
+            targetBB = bb->trueBranch();
+            if (speculationTarget->bb()->beforeInCfg(targetBB)) {
+                insertPosition = targetBB->begin();
+            } else {
+                if (targetBB->beforeInCfg(speculationTarget->bb()))
+                    targetBB = speculationTarget->bb();
+                insertPosition = targetBB->atPosition(speculationTarget) + 1;
+            }
```
o-: this does not seem right. how do you know that the checkpoint is still available at that position?

rir/src/compiler/pir/bb.cpp
charig: Empirically works. The commment just expose that, to make it sound we need to calculate the `cfg` for this pass. I guess that will increase compilation time significantly. Clearly it is easy to do. I just was not sure it is worth to pay for the overhead at this point. We probably should have a cfg for compiled function that is mutable and accessed from anywhere?

rir/src/compiler/opt/type_speculation.cpp
charig: because the previous code calculates exactly which is the cp for that instruction?

rir/src/compiler/pir/bb.cpp
o-: > Empirically works

?? what does it even mean, e.g. in a loop

rir/src/compiler/opt/type_speculation.cpp
o-: ok, I think I get now what you are trying to do.

on line 66 you know if we have the  `next` or the `at` checkpoint. so no need to do this dance here to figure out again. if we did `checkpoint.at`, then we must insert the assume right after i (like we do now). if we did `checkpoint.next` then we must insert the assume at `cp->next()->begin()`. right?

so it seems like this was somehow completely broken, but never triggered any issues? do you understand why?

charig: I've seen similar things when working around the more aggressive delaying of instructions. Also, the environment elision had some issues. My impression is that we were performing all these speculative optimizations only in very few and simple places...

rir/src/compiler/pir/bb.cpp
charig: It is implemented as a `<`. So the semantics it is intending to implement are `cfg.strictilyPreceeds(a,b)`

o-: well, the point is that this is currently only correct after the cleanup phase.

I am open to making this the default, since it might help us in many places. but not like this.

we could extract the renumbering from the cleanup phase and then have a renumber pass, that you can schedule before every pass where you want to rely on this property. and then some boolean on the function that tells you if the bb's are ordered. then we can clear it whenever we change sth and we can assert that it is set when we need it.

o-:
hmm, basically this pr suggests that the speculation on promised values never really worked. and that probably all attempts at making it work and not regress perf were fake...

I wouldn't be surprised if this PR also regresses perf

PR #701, opened 2019-10-31
use llvm pass builder to build the pass schedule

o-:
as the only customization add a cold/hot splitting pass that should
help with deopt branches.


o-:
ran this on some examples and it seems slower to compile and to produce worse code. need to investigate first. for now lets try #702 instead

PR #703, opened 2019-11-01
Instead of returning the earliest avaialable checkpoint (which is not…

charig:
… possible using this lattice), we return the latest


rir/src/compiler/analysis/available_checkpoints.h
```diff
@@ -11,16 +11,13 @@ namespace pir {
...
-            if (auto cp = Checkpoint::Cast(i)) {
-                state.set(cp);
-                return AbstractResult::Updated;
-            }
+        if (state.get() && i->isDeoptBarrier()) {
+            state.clear();
+            return AbstractResult::Updated;
+        }
+        if (auto cp = Checkpoint::Cast(i)) {
+            state.set(cp);
```
o-: maybe add  `if (cp != state.get())` to avoid returning Updated when we don't change anything

PR #705, opened 2019-11-01
add a silly heuristic for local variables named `c`

o-:


rir/src/compiler/translations/rir_2_pir/rir_2_pir.cpp
```diff
@@ -222,6 +222,8 @@ bool Rir2Pir::compileBC(const BC& bc, Opcode* pos, Opcode* nextPos,
     case Opcode::ldvar_cached_:
     case Opcode::ldvar_for_update_:
     case Opcode::ldvar_for_update_cache_:
+        if (bc.immediateConst() == symbol::c)
```
charig, line 225: Why do we need to consider also loads?

o-: it's a heuristic... if I see a load for `c` then i assume that `c` is potentially used as a variable...

PR #739, opened 2019-11-12
Env spec fix

charig:


rir/src/compiler/opt/elide_env_spec.cpp
```diff
@@ -122,8 +123,8 @@ void ElideEnvSpec::apply(RirCompiler&, ClosureVersion* function,
     // If we only see these (and call instructions) then we stub an environment,
     // since it can only be accessed reflectively.
     static std::unordered_set<Tag> allowed{
-        Tag::Force,      Tag::PushContext, Tag::LdVar,     Tag::StVar,
-        Tag::StVarSuper, Tag::Call,        Tag::FrameState};
+        Tag::Force,      Tag::PushContext, Tag::LdVar,      Tag::StVar,
+        Tag::StVarSuper, Tag::Call,        Tag::FrameState, Tag::CallBuiltin};
```
o-: this is already whitelisted below by `bt && !supportsFastBuiltinCall(bt->blt)`.

this is a bit stupid, but our problem is that if we don't support the builtin either as a fast builtin in `interpreter/builtin.cpp` or as a native builtin in the compiler, then we will end up materializing the environment for sure. so we can only stub envs if we know that we compile a builtin that we have a fastcase for. we should extend that list to also allow builtins that we can handle directly in the native backend, I think that is still missing...

charig: It is not whilelist because `!allowed()` already is enough for banning

o-: got it!

o-: good catch

Issue #789, opened 2020-01-22
sequences starting with non-constant value


o-:
For some reason in https://github.com/reactorlabs/rir/blob/master/rir/src/ir/Compiler.cpp#L238 we only properly compile a simple for loop `for (i in a:b)` if `a` is a numerical constant.

This causes us to allocate lots of unnecessary vectors (e.g. in nbody_naive_2).

I suspect that this is a hack, exploiting the fact that `a < b` happens to error similarly to `a:b` if `a` is a number.

This should be fixable quite easily by emitting a check for the type of `a`  ensuring that it is compatible with `:` first.

Jakobeha:
Also we need to handle the situation if `a` is a promise which changes `:`.

Would it be better to just do what GNU-R does and override `colon_` in the interpreter, so that it creates an ALTREP.

Alternatively we could do promise speculation in RIR bytecode (with the original `for` loop as a fallback, so we copy the bytecode), and then if PIR is smart enough with constantfolding and we know `a` isn't a promise it would get rid of the check and dead branch.

o-:
> Also we need to handle the situation if a is a promise which changes :.

I don't see that we currently gracefully handle the situation where `:` would be dynamically overwritten at all. so that seems not to make things worse...

> Would it be better to just do what GNU-R does and override colon_ in the interpreter, so that it creates an ALTREP.

unfortunately we don't support altrep in the native backend and I am not sure we want to get into that. also for llvm to be able to optimize the loop it has to look roughly like a C style loop.

o-:
I think this should be addressed as a pir optimizer pass, since probably in rir it is difficult to do so...

PR #800, opened 2020-01-25
Variable simple range

Jakobeha:
Addresses #789. Note that this may slow performance because we still usually have to compile the regular for, unless we can statically resolve that the lower bound will behave like an integer (which is hard because usually it's a real, and it would need to be rounded to an integer if so)

### DIFF
--- a/rir/src/compiler/native/lower_llvm.cpp
+++ b/rir/src/compiler/native/lower_llvm.cpp
@@ -4688,7 +4688,10 @@ bool LowerFunctionLLVM::tryCompile() {
             }
 
             case Tag::ColonInputEffects:
-                // TODO: Handle via a native builtin
+            case Tag::ColonCastLhs:
+            case Tag::ColonCastRhs:
+            case Tag::ColonGetStep:
+                // TODO: Handle via native builtins
                 success = false;
                 break;
 
--- a/rir/src/compiler/native/lower_llvm.cpp
+++ b/rir/src/compiler/native/lower_llvm.cpp
@@ -3570,6 +3570,12 @@ bool LowerFunctionLLVM::tryCompile() {
                 }
 
                 auto t = IsType::Cast(i);
+                if (t->isIntegerCastable) {
+                    // TODO: Generate native code to check this
+                    success = false;
+                    break;
+                }
+
                 auto arg = i->arg(0).val();
                 if (representationOf(arg) == Representation::Sexp) {
                     auto a = loadSxp(arg);
--- a/rir/src/compiler/native/lower_llvm.cpp
+++ b/rir/src/compiler/native/lower_llvm.cpp
@@ -2933,50 +2933,12 @@ bool LowerFunctionLLVM::tryCompile() {
                 break;
             }
 
-            case Tag::AsInt: {
-                auto arg = i->arg(0).val();
-                auto asint = AsInt::Cast(i);
-                llvm::Value* res = nullptr;
-                if (representationOf(arg) == Representation::Integer) {
-                    res = load(arg, Representation::Integer);
-                } else if (representationOf(arg) == Representation::Real) {
-                    auto a = load(arg, Representation::Real);
-                    if (asint->ceil) {
-                        res = builder.CreateIntrinsic(Intrinsic::ceil,
-                                                      {a->getType()}, {a});
-                    } else {
-                        res = builder.CreateIntrinsic(Intrinsic::floor,
-                                                      {a->getType()}, {a});
-                    }
-                    res = builder.CreateFPToSI(res, t::Int);
-                } else {
-                    auto bt = asint->ceil ? NativeBuiltins::asIntCeil
-                                          : NativeBuiltins::asIntFloor;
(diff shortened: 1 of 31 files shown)


Jakobeha:
superseded by `debug-improvements`

rir/src/ir/Compiler.cpp
```diff
@@ -224,112 +224,117 @@ void emitGuardForNamePrimitive(CodeStream& cs, SEXP fun) {
...
+            cs << skipRegularForBranch;
+
+            // m' <- colonCastLhs(m')
+            cs << BC::swap() << BC::colonCastLhs() << BC::swap();
+
+            // n' <- colonCastRhs(m', n')
+            cs << BC::colonCastRhs();
+
+            // step <- colonGetStep(m', n')
+            cs << BC::colonGetStep();
```
o-: I think the old version of explicitly computing `step = m<n ? 1 : -1` was better, because pir can speculate on the branch (thanks to `recordTest`) and then constantfold step. this is not possible anymore here.

rir/src/compiler/native/lower_llvm.cpp
```diff
@@ -4725,6 +4687,14 @@ bool LowerFunctionLLVM::tryCompile() {
                 break;
             }
 
+            case Tag::ColonInputEffects:
+            case Tag::ColonCastLhs:
+            case Tag::ColonCastRhs:
+            case Tag::ColonGetStep:
+                // TODO: Handle via native builtins
```
o-: this needs to be implemented, or you disable the native backend for any function with a for loop, that's not good...

rir/src/compiler/opt/cleanup.cpp
```diff
@@ -130,13 +130,13 @@ class TheCleanup {
...
-                        asInt->replaceUsesWith(arg);
-                        removed = true;
-                        next = bb->remove(ip);
+                } else if (auto colonInputEffects =
+                               ColonInputEffects::Cast(i)) {
+                    auto lhs = colonInputEffects->arg<0>().val();
+                    auto rhs = colonInputEffects->arg<1>().val();
+                    if (!lhs->type.maybeHasAttrs() ||
+                        !rhs->type.maybeHasAttrs()) {
+                        colonInputEffects->replaceUsesWith(True::instance());
```
o-: move to constantfolding.

rir/src/ir/Compiler.cpp
```diff
@@ -224,112 +224,117 @@ void emitGuardForNamePrimitive(CodeStream& cs, SEXP fun) {
...
-            cs << BC::ceil() << BC::dec() << BC::ensureNamed() << BC::swap();
-            // diff' <- -1
-            cs << BC::push(-1);
-            // gt' <- TRUE
-            cs << BC::push(R_TrueValue);
-            cs << BC::put(3) << BC::put(2) << BC::br(startBranch) << fwdBranch;
+            cs << BC::force();
+            // if (!colonInputEffects(m, n)) {
+            cs << BC::colonInputEffects();
+            cs.addSrc(seq);
```
o-, line 279: putting a `recordTest` here allows the compiler to speculate on dead branches more easily

rir/src/compiler/pir/instruction.h
```diff
@@ -1134,17 +1134,79 @@ class FLI(AsTest, 1, Effects() | Effect::Error | Effect::Warn) {
     size_t gvnBase() const override { return tagHash(); }
 };
 
-class FLI(AsInt, 1, Effect::Error) {
+class FLI(ColonInputEffects, 2, Effect::Error) {
   public:
-    bool ceil;
+    explicit ColonInputEffects(Value* lhs, Value* rhs, unsigned srcIdx)
+        : FixedLenInstruction(NativeType::test,
+                              {{PirType::any(), PirType::any()}}, {{lhs, rhs}},
```
o-: this instruction expects evaluated values, not promises. so the input types should be `PirType::val()` and not `any()`

rir/src/compiler/util/type_test.h
```diff
@@ -33,9 +33,10 @@ class TypeTest {
 
         assert(feedback.origin);
         // First try to refine the type
-        if (expected.noAttribs().isA(RType::integer) ||
-            expected.noAttribs().isA(RType::real) ||
-            expected.noAttribs().isA(RType::logical)) {
+        if (!expected.maybeObj() && // TODO: Is this right?
```
o-, line 36: I think currently noAttrib implies noObj, but it's more robust, so I think it's ok.

o- (commented):
nice PR!
can you constantfold the castLhs/Rhs one too if the inputs are numerical? not sure, needs some thought.

rir/src/ir/Compiler.cpp
o-: to be clear, I am proposing to remove the colonGetStep instruction in favor if just a branch on `lt` that pushes either `1L` or `-1L`

rir/src/compiler/pir/instruction.h
o-: (same for the other instrs)

PR #801, opened 2020-01-27
perform range analysis just before its use

manasthakur:
range analysis need not be performed for instructions other than `Extract1_1D`


manasthakur:
Not needed as the analyses extending the generic version are performed lazily!

PR #804, opened 2020-01-31
Track specific instructions in log

Jakobeha:
Adds 2 debug flags:

- `PrintPassesIntoFolders`: Prints each optimization pass into a separate file under the folder `<name>-<id>` (for reference, `PrintIntoFiles` prints into `<name>-<id>.log`). So instead of a stream of passes in 1 file there are many files with 1 pass.
- `PrintInstructionIds`: Makes instructions print their uids (memory addresses)

Both of these are used to better track an instruction across passes. Specifically, I'm trying to find which pass an instruction is deleted. Using features of this PR, one can:

- Run R with `PIR_DEBUG=PrintPassesIntoFolders,PrintOptimizationPasses,PrintInstructionIds`, compile the target closure where an instruction is incorrectly deleted
- find a pass where the instruction is present, remember its uid
- find the uid across multiple files (using e.g. in VSCode or "find" command), find the first file where the uid is missing
- The previous pass must have removed the instruction

e.g. in pir_regression8.R `ex-pir-function-00000`, the closure environment is removed. It's present in the first pass, all the way until pass 164. Pass 165 is a cleanup pass, which must remove it because it's no longer used. Once again, in the first pass it's used by 2 loads - the first (`LdFun`) is removed in pass 14 (another cleanup, because it's just never used) and the second (`LdVar`) gets shifted to the global env in pass 28 by scope resolution - and some framestates.

Also, printing each pass into a separate file makes it easier at least for me to jump through passes, since I don't need to scroll.

### DIFF
--- a/documentation/debugging.md
+++ b/documentation/debugging.md
@@ -26,7 +26,12 @@ completely disables the PIR optimizer. As follows are the different Options avai
 
     PIR_DEBUG=                     (only most important flags listed)
         help                       list all available flags
+        PrintPassesIntoFolders     print each pass into `<closure>-pir-function/<pass#>.log` (or `.out` if GraphViz)
+        PrintIntoFiles             print into `<closure>-pir-function.log` (or `.out` if GraphViz)
         PrintIntoStdout            print without buffering (useful for crashes during compilation)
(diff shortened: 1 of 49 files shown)


Jakobeha:
superseded by debug-improvements

PR #805, opened 2020-02-03
Debug improvements

Jakobeha:
Adds 3 debug flags:

- PrintPassesIntoFolders: Prints each optimization pass into a separate file under the folder <name>-<id> (for reference, PrintIntoFiles prints into <name>-<id>.log). So instead of a stream of passes in 1 file there are many files with 1 pass.
- PrintInstructionIds: Makes instructions print their uids (memory addresses)
- PrintSSAInRir: For each PIR instruction, adds a "comment" to the bytecode describing it and its stack before the corresponding RIR instructions it emits. *Note: this overrides srcref, so it may affect certain calls. I'm not sure if srcref affects anything other than log / error messages*

Also adds a macro `PRINT_STACK` to `interp.cpp`, which (when `PRINT_INSTR` is also defined) makes the interpreter print the RIR stack before each instruction.

Lastly, minor improvements to existing logging and documentation.


.vscode/c_cpp_properties.json
```diff
@@ -22,9 +23,9 @@
             "defines": [],
             "compilerPath": "/usr/bin/gcc",
             "cStandard": "c11",
-            "cppStandard": "c++17",
+            "cppStandard": "c++14",
```
o-: we are actually still using c++11

rir/src/utils/optional.h
```diff
@@ -0,0 +1,2160 @@
+
```
o-: is it really necessary to drag this in? our build times are long enough as it is...

rir/src/utils/filesystem.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
o-, line 1: a bit strange since we use the ifdef style everywhere else..

rir/src/utils/filesystem.cpp
```diff
@@ -0,0 +1,28 @@
+#include "filesystem.h"
+#include <errno.h>
+#include <ftw.h>
+#include <stdio.h>
+#include <unistd.h>
+
+// From
+// https://stackoverflow.com/questions/5467725/how-to-delete-a-directory-and-its-contents-in-posix-c
+static int unlink_cb(const char* fpath, const struct stat* sb, int typeflag,
```
o-: naming convention is camelCase

o- (commented):
> Note: this overrides srcref, so it may affect certain calls. I'm not sure if srcref affects anything other than log / error messages

yes it does, sorry that does not work. why do you want to add it to the code stream instead of provising some printing functionality?

as I said, I really like change 1 and 2. we need to talk about 3, so maybe if you want to proceed it would be better to split them.

.vscode/c_cpp_properties.json
Jakobeha: according to CMakeLists (line 42) it specifies c++14?

rir/src/utils/optional.h
Jakobeha: not if it takes too long. But I figure some sort of optional type would be useful, we already have some workarounds in existing code. Maybe a much simpler one, like

```cpp
template <typename T>
class Optional {
  T storage;
  
 public:
  const bool hasValue;

  Optional() : hasValue(false), storage() {}
  Optional(T value) : hasValue(true), storage(value) {}
  
  T value() { assert hasValue; return storage; }
}
```

is ok?

rir/src/utils/filesystem.h
Jakobeha: Not everywhere

I do think we should standardize some of our style guide though: we use clang-format but there is also absolute/relative imports, method names, etc.

rir/src/utils/filesystem.cpp
Jakobeha: yes that code was copy/pasted from StackOverflow. I admit I was lazy and didn't want to write the "delete directory" function manually :)

rir/src/utils/optional.h
Jakobeha: nevermind, without SSA printing at least for this branch I don't need optional.

.vscode/c_cpp_properties.json
o-: oh, i did not remember, alright then :)

PR #806, opened 2020-02-04
Update to R version 3.6.2

o-:


o-:
@Jakobeha as already hinted at, this partially reverts #799 since that PR broke travis. can you look into that? to get a clean travis run for a PR that updates the gnur module you unfortunately have to clear the travis caches first. let me know if you need some help with that.

PR #812, opened 2020-02-11
Refactor LLVM - Split `lower_llvm.cpp` into many files + other small changes

Jakobeha:
Main changes:

- Splits `lower_llvm.cpp` into many files:
  - `lower_function.h`: Interface for `LowerFunctionLLVM` (renamed to `LowerFunction`)
  - `lower_function/helpers.cpp`: All of the helpers for `LowerFunctionLLVM` except `tryInlineBuiltin` and `tryCompile`
  - `lower_function/builtins.cpp`: `tryInlineBuiltin`, with the implementation for inlining builtins
  - `lower_function/lower.cpp`: `tryCompile`
  - `phi_builder.h`: The `PhiBuilder` class
  - `variable.h`: The `Variable` class
  - `representation.h`: The `Representation` class
  - `constants.h`: The static helpers to create LLVM constants
  - `lower.h` / `lower.cpp` itself: the interface / implementation of `LowerLLVM`
- Moves `native` outside of `compiler`
- Creates `llvm_imports.h` which includes all of the LLVM libraries, makes other files explicitly include this file first, and documents why: R's macros break LLVM's headers, and these standards would help prevent import headaches.

There are other unrelated refactors (when I was rewriting certain functions to handle the NaN-box case), which I can revert if necessary.

---

The git diff isn't very helpful. It's better to copy the original `lower_llvm.cpp` and compare it against each of the above (it would be awesome if you could get git to do this automatically but I doubt it's possible).

Overall, I'm not sure what you think of this / how useful it is. The main reason I created the PR is because I was refactoring some of the LLVM code when experimenting with implementing NaN-boxed values (the code is here https://github.com/reactorlabs/rir/tree/unboxed-multirep-native), to help myself edit and VSCode's tools run faster. There aren't (supposed to be) any semantic changes, and I'm assuming this holds because/iff the tests pass...


o-:
@Jakobeha I am sorry, but this is just doing too much at the same time. Can we discuss smaller increments and plan them a bit better?

PR #813, opened 2020-02-12
variable simple range fix

Jakobeha:
Attempt to fix the crash on nbody_2 from #800. Also added a test so we can verify before merging.


o-:
I manually merged everything here except 5caf1e8. Please create a new PR for that feature

PR #822, opened 2020-02-24
changed ColonInputEffects to NativeType::test

Jakobeha:
The only thing I needed was to change `RecordDeoptReason` so that it takes the type of its input instead of `PirType::val`

### DIFF
--- a/rir/src/compiler/pir/instruction.h
+++ b/rir/src/compiler/pir/instruction.h
@@ -1137,9 +1137,8 @@ class FLI(AsTest, 1, Effects() | Effect::Error | Effect::Warn) {
 
 class FLI(ColonInputEffects, 2, Effect::Error) {
   public:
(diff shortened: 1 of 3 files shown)


rir/src/compiler/pir/instruction.h
```diff
@@ -1828,7 +1827,7 @@ class FLI(RecordDeoptReason, 1, Effects::Any()) {
   public:
     DeoptReason reason;
     RecordDeoptReason(const DeoptReason& r, Value* value)
-        : FixedLenInstruction(PirType::voyd(), {{PirType::any()}}, {{value}}),
+        : FixedLenInstruction(PirType::voyd(), {{value->type}}, {{value}}),
```
o-, line 1830: I like that change, if it works, as it should prevent us from wrongly replacing values with something of a different type..

o- (approved):

PR #825, opened 2020-02-26
Skip NA check in simple for increment

Jakobeha:
To implement this, I extended the type-system with a non-NA flag. I know it's a broad approach but I couldn't figure out any clean narrower alternative - I could add a property `bool isNa` to `pir::Value`, or make the native backend explicitly check if lhs and rhs are constants or colon cast instructions, but both approaches are hacky and probably would end up being more work.

I did explicitly prevent PIR from speculating on non-NA values though, because that was breaking `pir_check` regression tests. Maybe later we can re-enable the speculation and add more non-NA optimizations, e.g. I think you mentioned that subscripts have to be boxed in case of NA arguments.

### DIFF
--- a/rir/src/compiler/pir/instruction.cpp
+++ b/rir/src/compiler/pir/instruction.cpp
@@ -495,7 +495,7 @@ LdConst::LdConst(SEXP c, PirType t)
 LdConst::LdConst(SEXP c)
     : FixedLenInstruction(PirType(c)), idx(Pool::insert(c)) {}
 LdConst::LdConst(int num)
-    : FixedLenInstruction(PirType(RType::integer).scalar().noAttribs()),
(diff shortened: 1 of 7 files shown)


rir/src/compiler/pir/instruction.cpp
```diff
@@ -495,7 +495,7 @@ LdConst::LdConst(SEXP c, PirType t)
 LdConst::LdConst(SEXP c)
     : FixedLenInstruction(PirType(c)), idx(Pool::insert(c)) {}
 LdConst::LdConst(int num)
-    : FixedLenInstruction(PirType(RType::integer).scalar().noAttribs()),
```
o-, line 498: why drop the noAttrib?

PR #826, opened 2020-02-27
Ci configuration for running paper experiments (no speculation)

charig:


o-:
you need to update the benchmark repo ref too. see container/benchmark/Dockerfile

charig:
> you need to update the benchmark repo ref too. see container/benchmark/Dockerfile

Did it yesterday

o-:
> > you need to update the benchmark repo ref too. see container/benchmark/Dockerfile
> 
> Did it yesterday

there is a commit id in that file, you need to update it. otherwise the container will not include the newest commits on the benchmark repo

PR #827, opened 2020-03-02
Skip NA/NaN checks in LLVM + better NA/NaN inference

Jakobeha:
New
- Native backend doesn't emit NA/NaN checks for static non-NaN values
- ~Type speculation on NA/NaN values~ `IsType` and `CastType` for non-NaN values, but only on instructions whose static type is non-NaN
- Improved type inference for NA/NaN

Improvements / bugfixes
- There is a NaN value for reals which isn't NA, but functions very similar to NA (e.g. when coerced to an integer / logical it becomes NA). Here we treat it identical to NA in PIR, hence "maybeNan", "notNan", etc.
- RIR type assertions also check for NA/NaN values, so if an instruction with a static non-NaN type ever produces NA/NaN under `RIR_CHECK_PIR_TYPES`, we will get an error.


rir/src/compiler/native/lower_llvm.cpp
```diff
@@ -1497,19 +1501,29 @@ void LowerFunctionLLVM::incrementNamed(llvm::Value* v, int max) {
...
+    assert(type.isA(PirType::num().scalar()));
     if (!notNa)
         notNa = BasicBlock::Create(C, "", fun);
-    if (v->getType() == t::Double) {
-        auto isNotNa = builder.CreateFCmpUEQ(v, v);
-        builder.CreateCondBr(isNotNa, notNa, isNa, branchMostlyTrue);
+    if (!type.maybeNAOrNaN()) {
+        // Don't actually check na because we statically know it's not
+        // Was having trouble using just CreateBr. The conditional jump will get
+        // optimized away
```
o-: ok, fine, but maybe at least unify all the 3 `builder.CreateCondBr(isNotNa, notNa, isNa, branchMostlyTrue);` into one and then just set the `isNotNa` to true in this case. Btw the correct constant to pass is `builder.getTrue()`

rir/src/compiler/native/types_llvm.cpp
```diff
@@ -316,7 +316,7 @@ int initializeTypes(LLVMContext& context) {
         llvm::FunctionType::get(t::Double, {t::SEXP}, false);
 
     NativeBuiltins::colonInputEffects.llvmSignature =
-        llvm::FunctionType::get(t::Int, {t::SEXP, t::SEXP, t::Int}, false);
+        llvm::FunctionType::get(t::SEXP, {t::SEXP, t::SEXP, t::Int}, false);
```
o-: why did you change this? this is super inefficient, as it will end up going through `unboxIntLgl`

rir/src/compiler/opt/constantfold.cpp
```diff
@@ -531,7 +542,11 @@ void Constantfold::apply(RirCompiler& cmp, ClosureVersion* function,
             if (auto colonInputEffects = ColonInputEffects::Cast(i)) {
                 auto lhs = colonInputEffects->arg<0>().val();
                 auto rhs = colonInputEffects->arg<1>().val();
-                if (!lhs->type.maybeHasAttrs() || !rhs->type.maybeHasAttrs()) {
+                if ((!lhs->type.maybeHasAttrs() ||
+                     !rhs->type.maybeHasAttrs()) &&
+                    !((lhs->type.maybe(RType::integer) ||
```
o-: I don't understand this, can you explain?

rir/src/compiler/pir/instruction.cpp
```diff
@@ -218,6 +218,42 @@ void Instruction::printGraph(std::ostream& out, bool tty) const {
     printEnv(out, tty);
 }
 
+bool Instruction::willDefinitelyNotOverflow() const {
+    if (!Add::Cast(this) && !Sub::Cast(this)) {
+        return false;
+    }
+    UsesTree uses(bb()->owner);
```
o-: please no expensive analysis in instructions! if you want to run an analysis then put it into opt/types.

rir/src/compiler/pir/instruction.cpp
```diff
@@ -661,35 +697,53 @@ TypeChecks IsType::typeChecks() const {
     auto in = arg(0).val();
     assert(!t.isVoid() && !t.maybeLazy());
     if (t.isA(PirType(RType::logical).orAttribs())) {
-        if (t.isScalar() && !t.maybeHasAttrs() && !in->type.isScalar())
+        if (t.isScalar() && !t.maybeHasAttrs() && !in->type.isScalar()) {
+            assert(t.maybeNAOrNaN() || !in->type.maybeNAOrNaN());
             return TypeChecks::LogicalSimpleScalar;
-        else
+        } else {
+            assert(t.maybeNAOrNaN() || !in->type.maybeNAOrNaN());
```
o-: why not put this at the top? you seem to be doing it in all cases?

rir/src/compiler/pir/instruction.h
```diff
@@ -355,24 +359,34 @@ class Instruction : public Value {
...
             if (m.maybe(RType::logical))
                 t = t | RType::integer;
+            // the binop result becomes NA if it can't be represented in a
+            // fixpoint integer (e.g. INT_MAX + 1 == NA)
+            // * the condition checks iff at least one of the arguments is an
+            // integer (doesn't happen with only logicals), and the result is an
+            // integer (doesn't happen with real coercion)
+            if (m.maybe(RType::integer) && t.maybe(RType::integer) &&
+                willDefinitelyNotOverflow())
+                t.setMaybeNAOrNaN();
```
o-, line 365: isn't that exactly the opposite of what you want?

rir/src/interpreter/interp.cpp
```diff
@@ -1635,15 +1637,26 @@ size_t expandDotDotDotCallArgs(InterpreterInstance* ctx, size_t n,
...
 
-bool colonInputEffects(SEXP lhs, SEXP rhs, unsigned srcIdx) {
+SEXP colonInputEffects(SEXP lhs, SEXP rhs, unsigned srcIdx) {
     auto getSrc = [&]() { return src_pool_at(globalContext(), srcIdx); };
 
     // 1. decide fastcase
-    bool fastcase = !inherits(lhs, "factor") || !inherits(rhs, "factor");
+    bool fastcase =
+        (!inherits(lhs, "factor") || !inherits(rhs, "factor")) &&
+        !((TYPEOF(lhs) == INTSXP || TYPEOF(lhs) == LGLSXP ||
```
o-: I don't understand this condition. can you comment?

rir/src/interpreter/interp.cpp
```diff
@@ -1635,15 +1637,26 @@ size_t expandDotDotDotCallArgs(InterpreterInstance* ctx, size_t n,
...
+           (TYPEOF(lhs) == REALSXP && doubleCanBeCastedToInteger(*REAL(lhs))) ||
```
o-, line 1661: this will segv for empty sized reals

rir/src/ir/Compiler.cpp
```diff
@@ -277,8 +278,21 @@ bool compileSimpleFor(CompilerContext& ctx, SEXP fullAst, SEXP sym, SEXP seq,
             cs << BC::colonInputEffects();
             cs.addSrc(seq);
             bool staticFastcase =
-                (isConstant(start) && !inherits(start, "factor")) ||
-                (isConstant(end) && !inherits(end, "factor"));
+                (isConstant(start) && !inherits(start, "factor") &&
```
o-, line 282: ok,I am reading this code for the 3rd time, still no clue what it does, and each time it looks a bit different...

rir/src/compiler/pir/type.cpp
```diff
@@ -90,34 +90,46 @@ void PirType::merge(SEXPTYPE sexptype) {
...
-            if (R_IsNA(COMPLEX(vector)[i].i))
-                return true;
-            break;
-        case STRSXP:
-            if (STRING_ELT(vector, i) == NA_STRING)
-                return true;
-            break;
-        default:
-            break;
+static bool containsNan(SEXP vector) {
```
o-: shoud we call this maybeContainsNaOrNan?

rir/src/compiler/native/types_llvm.cpp
Jakobeha: It was necessary because `ColonInputEffects` uses `NativeType::test` and gets passed to a `Branch` instruction. There was a bug where the native backend was boxing the returned integer into a different value, which was messing up the `Branch`. It's weird too because the native backend does use integers for `NativeType::test`, so the bug doesn't usually occur, it was the one that only occurred in the release build in Docker containers and I couldn't reproduce locally or in PRL5.

o-: But test boxing is handled in https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/native/lower_llvm.cpp#L847

o-:
can you remove commit 63ca39a  from this PR?

PR #832, opened 2020-03-04
Branch elim

skrynski:
Performs branch elimination by placing Phi's only in simple scenarios. Next step would be to incorporate proper PhiPlacement

### DIFF
--- a/rir/src/compiler/test/PirTests.cpp
+++ b/rir/src/compiler/test/PirTests.cpp
@@ -724,16 +724,12 @@ static Test tests[] = {
                  "theFun <- function() {a<-FALSE; if (a) {q <-1} else "
                  "{if (a) q <- 3 else q <- 2}; q}");
          }),
-    /*
-          // it was removed because with the 'branch elim' optimization
-          // the LoadVar is not present anymore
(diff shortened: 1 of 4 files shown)


rir/src/compiler/test/PirTests.cpp
```diff
@@ -724,11 +724,16 @@ static Test tests[] = {
...
-         []() {
-             return hasLoadVar("theFun <- function(a) {if (a) {q <-1} else {if "
-                               "(a) 3 else q <- 2}; q}");
-         }),
+    /*
+          // it was removed because with the 'branch elim' optimization
+          // the LoadVar is not present anymore
+          Test("merge_missing",
+           []() {
+               return hasLoadVar("theFun <- function(a) {if (a) {q <-1} else {if
```
o-: try changing the test to `function (a,b) if (a) ... else if (b) ...`

o-:
cool, thanks for this PR!

rir/src/compiler/test/PirTests.cpp
skrynski: Yes, it's working this way. I will commit the changes

PR #843, opened 2020-03-27
count rir/llvm compilation/evaluation using the old event counter api

Jakobeha:
To enable, set the CMake configuration to `debug`, `debugopt` or `measure`. Then run with `ENABLE_EVENT_COUNTERS=1`. The output will be in `rir_statistics.csv`. The specific events counted are:

- LLVM code evaled (in `evalRirCode`)
- RIR (not LLVM) evaled
- PIR optimized (went through `Rir2Pir` into either RIR or LLVM)
- Unoptimizable (tried to PIR optimize but gave up, e.g. because the function was too big)
- LLVM lowered (went through `Pir2Rir` and generated native code)
- RIR lowered (went through `Pir2Rir` *and didn't generate native code* (technically we always compile a RIR version, even if we also compile the LLVM version. This doesn't count)))
- deopt


o-:
other useful information: where do we call legacyArgListFromStackValues, which R builtins or specials do we end up calling through doCall (i think there is something already called BUILTIN_SLOWCASE_COUNTER)

Jakobeha:
There is already SLOWCASE_COUNTER which tracks which builtins we call through doCall, so do you want me to do anything with that?

Jakobeha:
Hello, I have added counters to track:
- \# of invocations per closure version or promise
- total execution time per closure version or promise (including nested closures / promises)
- \# of call sites per closure version
- \# of closure versions per dispatch table
- \# of deoptimized closure versions per dispatch table

Note that when we call closures in non-RIR code (e.g. in the REPL) or in the native backend, we might merge call-sites, because we can't get the specific address of the call and (in non-RIR code) the caller.

Jakobeha:
Renamed branch to `statistics`

PR #844, opened 2020-03-31
Revert "Skip NA/NaN checks in LLVM + better NA/NaN inference"

o-:
Reverts reactorlabs/rir#827


o-:
@Jakobeha  the PR broke this test: https://gitlab.com/rirvm/rir_mirror/-/jobs/492719388

PR #850, opened 2020-04-07
Fix liveness2

o-:


rir/src/compiler/analysis/liveness.cpp
```diff
@@ -146,6 +149,15 @@ LivenessIntervals::LivenessIntervals(Code* code, unsigned bbsSize) {
...
     }
 
+    Visitor::run(code->entry, [&](BB* bb) {
+        for (auto n : bb->succsessors()) {
+            if (!liveAtEnd.count(n))
+                todo.insert(n);
+        }
+    });
+    if (!todo.empty())
+        goto restart;
```
charig, line 159: Do we really need a `goto` to solve this?

skrynski: No, we don't . This was actually a quick and dirty thing we did with oli to validate an idea. I'm getting rid of it on my next commit.

o-: goto's are animals too

PR #854, opened 2020-04-13
Speed up dominators algorithm

mhyee:
I rewrote the dominators computation to use the Lenguaer-Tarjan algorithm. There were a few drive-by commits I included:

- Typo in BB successors method.
- LT algorithm assumes the start node has no predecessors, so I made some changes to check/enforce that.
- Previously, the `dominates` method only checked for strict dominance, i.e. a node does not dominate itself. I made `dominates` check for non-strict dominance and added a `strictlyDominates` method. I also went and changed other places in the codebase to use `strictlyDominates` if it made sense. @skrynski reviewed that commit for me.

The implementation is thoroughly documented, but it is a fairly close translation of the pseudocode from [_Modern Compiler Implementation_](https://www.cs.princeton.edu/~appel/modern/).

If you have any branches that use dominance, please double-check after merging in this branch. Check whether you really want strict or non-strict dominanace.

### Performance numbers

On the flexclust benchmark, this speeds up the compiler overhead by about 2 seconds. I ran flexclust with our benchmark framework, but also inserted timing/counters into the dominators code. These timers/counters are across all runs of the benchmark.

**Before this PR**
```
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 16346475us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 14905355us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 21660213us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 21436819us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 14856454us
flexclust_no_s4: iterations=5; average: 17841063 us; total: 89205316us
=== DominanceGraph perf:
                DomGraph ctor    4988 ms (called     61940 times)
               DomGraph merge    4434 ms (called   1732656 times)
          DomGraph dominates?      42 ms (called    911066 times)
         DomGraph idominates?     186 ms (called   7798074 times)
```

We can see that there is some overhead on the first run, and significant overhead on the third and fourth runs. The time for `DomGraph merge` is included inside `DomGraph ctor`. Note that while we spend 5 seconds inside `DomGraph ctor`, that time is mainly split over the third and fourth runs, i.e. we spend 2.5 seconds constructing the dominance graph.

**With this PR**
```
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 15604257us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 14444395us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 18667478us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 18996542us
Flexclust::kcca  5  352.0658   74
flexclust_no_s4: iterations=1 runtime: 14712814us
flexclust_no_s4: iterations=5; average: 16485097 us; total: 82425486us
=== DominanceGraph perf:
                DomGraph ctor      90 ms (called     61950 times)
          DomGraph dominates?     169 ms (called    928404 times)
         DomGraph idominates?     153 ms (called   8083418 times)
```

There is a similar pattern for when we see the overheads. However, the time for the slow runs (numbers three and four) is less, about 2 seconds faster than before. We also see that we spend significantly less time inside `DomGraph ctor`.

`DomGraph dominates` got a bit slower -- before, it was looking up a value in an array. But now, we need to walk up the dominator tree. We could precompute this, but I don't think it's worth the trouble.

### DIFF
--- a/rir/src/compiler/util/cfg.cpp
+++ b/rir/src/compiler/util/cfg.cpp
@@ -49,7 +49,6 @@ bool CFG::isPredecessor(BB* a, BB* b) const {
         std::bind(std::equal_to<BB*>(), std::placeholders::_1, a));
 }
 
-// TODO: fix bug, algorithm assumes start node has no predecessor...
 DominanceGraph::DominanceGraph(Code* start) : idom(start->nextBBId) {
     int size = start->nextBBId;
     int N = 0;
--- a/rir/src/compiler/util/cfg.cpp
+++ b/rir/src/compiler/util/cfg.cpp
@@ -20,7 +20,7 @@ CFG::CFG(Code* start)
                 transitivePredecessors[next->id].push_back(bb);
             }
         };
-        auto succs = bb->succsessors();
+        auto succs = bb->successors();
         for (auto suc : succs)
             apply(suc);
         if (succs.size() == 0)
@@ -90,7 +90,7 @@ DominanceGraph::DominanceGraph(Code* start) : dominating(start->nextBBId) {
             if (d.merge(curState))
                 todo.push(bb);
         };
-        for (auto suc : cur->succsessors())
+        for (auto suc : cur->successors())
             apply(suc);
     }
 }
@@ -147,7 +147,7 @@ DominanceGraph::BBSet DominanceGraph::dominatedSet(Code* start,
                 }
             }
         };
-        for (auto suc : cur->succsessors())
+        for (auto suc : cur->successors())
             apply(suc);
     }
 
--- a/rir/src/compiler/util/cfg.cpp
+++ b/rir/src/compiler/util/cfg.cpp
@@ -49,49 +49,114 @@ bool CFG::isPredecessor(BB* a, BB* b) const {
         std::bind(std::equal_to<BB*>(), std::placeholders::_1, a));
 }
 
-bool DominanceGraph::DomTree::merge(const DomTree& other) {
-    bool changed = false;
-    auto it = container.begin();
-    while (it != container.end()) {
-        if (std::find(other.container.begin(), other.container.end(), *it) ==
-            other.container.end()) {
-            it = container.erase(it);
-            changed = true;
-        } else {
-            it++;
-        }
-    }
-    return changed;
-}
+// TODO: fix bug, algorithm assumes start node has no predecessor...
+DominanceGraph::DominanceGraph(Code* start) : idom(start->nextBBId) {
+    int size = start->nextBBId;
+    int N = 0;
 
-DominanceGraph::DominanceGraph(Code* start) : dominating(start->nextBBId) {
-    // Static Analysis computes the set of all dominating bb's, for every bb
-    // reachable from start. Runs until none of the sets grow anymore.
+    std::vector<SmallSet<BB*>> bucket(size);
+    std::vector<int> dfnum(size, -1);
+    BBList semi(size);
+    BBList ancestor(size);
+    BBList samedom(size);
+    BBList parent(size);
+    BBList best(size);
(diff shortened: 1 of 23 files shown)


rir/src/compiler/util/cfg.cpp
```diff
@@ -49,49 +49,303 @@ bool CFG::isPredecessor(BB* a, BB* b) const {
...
+
+    // Note that this is an overapproximation of the actual number of BBs,
+    // because some BBs may have been deleted.
+    int size = start->nextBBId;
+
+    // Counter for numbering BBs, by depth-first search order.
+    int N = 0;
+
+    // Indexed by BB id. `dfnum[n->id]` is the dfnum of node `n`.
+    std::vector<int> dfnum(size);
```
o-, line 165: @mhyee why do you keep a number on the side, when you can use the bb->id to store it? The compiler is designed such that it is always allowed to renumber the bbs between passes, the only invariant is that the max is is smaller than nextBBId (you can even reset nextBBId after renumbering). The id is mostly used to speed up lookups in analysis and visitors (by using the id as an array index). Therefore the only caveat is to not use the default visitor while you renumber (see BBTransform::renumber for an example).

You could make the BBTransform::renumber apply this order. We could even exploit this in passes where we need to detect back-edges. Currently we use the dominance analysis for that, but if the id's are such that back-edges are from higher to lower numbers, then in those cases renumbering would be enough.

UPDATE: ok, I see how this could be very confusing, when you compute some analysis, then do dominance analysis and suddenly your analysis is broken, because dominance analysis invalidated all your bb id's... Let's think first how we could have an api to deal with this. For example we could hide the actual dominance implementation from users and rather have a declarative way where passes can request analysis and the pass manager would take care of ordering.

o-:
most beautiful PR award :joy: 
thanks

rir/src/compiler/util/cfg.cpp
mhyee: Honestly, I just didn't think of this. Doing it on-the-fly means I need some other way of tracking which BBs have already been seen, so the code might be slightly messier.

But you're right -- if some other analysis requires dominance, it might break if the BBs are renumbered...

mhyee:
> most beautiful PR award 😂
> thanks

Haha, this was the only way I had a chance of understanding what the algorithm was doing. I think I understand _what_ it's doing, but _how_ and _why_ are different matters...

Also, completely unscientific "benchmarking": the CI run for this merge took 4 hours, but previous runs on `master` would take 4.5 to 4.75 hours.

rir/src/compiler/util/cfg.cpp
o-: > But you're right -- if some other analysis requires dominance, it might break if the BBs are renumbered...

ah, i did not realize that you are already using the id permanently. I think that is an ok assumption, that id's are mostly stable. but it would be nice if your analysis could somehow rely on the fact that they are already correctly ordered. maybe we could even have this as an invariant?

mhyee: But there's quite a few places where we insert or merge blocks, and that would ruin the depth-first order. So I don't think we can guarantee that we always have depth-first order, unless we do a renumber after every BB transformation.

o-: right. just throwing around ideas...

PR #865, opened 2020-04-21
Skip passes if no change happened since last running them

JanJecmen:


rir/src/compiler/opt/pass_scheduler.h
```diff
@@ -29,21 +30,34 @@ class PassScheduler {
...
+                        if (budget <= pass->cost()) {
                             budget = 0;
                             break;
                         }
+                        // We assume that if no change occured since last
+                        // running this pass, we may safely skip it (if the last
+                        // change was done by this pass, we skip too - the pass
+                        // should do everything in a single invocation)
+                        if (lastChange <= lastPassRun[pass->getName()]) {
+                            continue;
```
o-, line 53: why not break?

rir/src/compiler/opt/pass_scheduler.h
```diff
@@ -29,21 +30,34 @@ class PassScheduler {
...
         for (auto& phase : schedule_.phases) {
             auto budget = phase.budget;
             bool changed = false;
             do {
                 changed = false;
                 for (auto& pass : phase.passes) {
+                    counter++;
                     if (!phase.once) {
-                        if (budget < pass->cost()) {
+                        if (budget <= pass->cost()) {
```
o-, line 44: off by one. if there is enough budget we should run the pass

rir/src/compiler/opt/pass_scheduler.h
```diff
@@ -56,6 +70,8 @@ class PassScheduler {
     Schedule schedule_;
     Schedule::Phases::iterator currentPhase;
 
+    mutable std::unordered_map<std::string, size_t> lastPassRun;
```
o-, line 73: ugh

o- (commented):
I am not sure if I understand the intention. my guess is you want to abort the phase if there is no change after running every pass once (whereas now the condition is also *starting from the first of the phase*).

this seems unnecessarily complex. you could keep a counter of passes run without a change and then: `if (passesWithoutChange >= phase.passes.size()) break;`

PR #870, opened 2020-04-24
Value profiler: keep track of TypeFeedback from previous compilation

ksmonkey123:


o-:
@ksmonkey123 the trick is to mark `mdEntries` as const

o-:
@ksmonkey123  Just two nitpicks that you could maybe address in a follow up PR: 1. there is some commented out code that should be removed. 2. if the profiler is disabled there is no need to create and store the mapping.

PR #916, opened 2020-09-23
Fixes

JanJecmen:
Some cleanup and fixes. Notable:
* printing PIR with flag OnlyChanges now uses the pass return value instead of comparing the output strings
* fixed problems with promargs materialization in gnur - forgotten case plus updating the contexts in case the result is empty
* fixed condition in llvm for is.vector where it would test the presence of attributes instead of absence in the fastcase


rir/src/api.cpp
```diff
@@ -66,7 +66,7 @@ REXPORT SEXP rir_disassemble(SEXP what, SEXP verbose) {
     return R_NilValue;
 }
 
-REXPORT SEXP rir_compile(SEXP what, SEXP env) {
+REXPORT SEXP rirCompile(SEXP what, SEXP env) {
```
o-, line 69: Currently all the symbols that we call from R (i.e. all functions in api.cpp) have this `rir_` prefix. it seems strange only to change one...

rir/src/compiler/native/lower_llvm.cpp
```diff
@@ -3222,7 +3222,7 @@ bool LowerFunctionLLVM::tryCompile() {
                     }
                     case blt("is.vector"):
                         if (auto cnst = LdConst::Cast(b->arg(1).val())) {
-                            if (b->arg(0).val()->type.maybeHasAttrs()) {
+                            if (!b->arg(0).val()->type.maybeHasAttrs()) {
```
o-, line 3225: whooops :)

o- (approved):

rir/src/interpreter/runtime.cpp
```diff
@@ -28,15 +24,9 @@ bool isValidClosureSEXP(SEXP closure) {
...
-    envSymbol = Rf_install("environment");
-    callSymbol = Rf_install(".Call");
-    execName = Rf_mkString("rir_executeWrapper");
-    R_PreserveObject(execName);
-    promExecName = Rf_mkString("rir_executePromiseWrapper");
-    R_PreserveObject(promExecName);
     // initialize the global context
     globalContext_ = context_create();
-    registerExternalCode(rirEval, rirApplyClosure, rirForcePromise, rir_compile,
+    registerExternalCode(rirEval, rirApplyClosure, rirForcePromise, rirCompile,
```
JanJecmen, line 29: > Currently all the symbols that we call from R (i.e. all functions in api.cpp) have this `rir_` prefix. it seems strange only to change one...

It was to be consistent with this but I agree :)

PR #917, opened 2020-09-23
Arg order

JanJecmen:


JanJecmen:
This PR adds a vector with the original order of arguments that were matched statically. Originally this was added to all call instructions except builtin calls, now only static calls have it ~(not sure if normal calls won't need it too, since StaticCall in some cases produces them).~ This is indeed what fails
The reordering is stored in the immediates of the calls and is used when gnur needs to materialize lazy promargs or if we need to do a legacy call.
There are now two functions to spawn an arglist, one for producing the promargs, ie. list after dots are expanded but before matching to formals, and one for creating the version after matching. The third place that creates an arglist is mk_env_ of inlined functions, and here the dots are also handled now.

rir/src/interpreter/interp.cpp
```diff
@@ -833,13 +920,12 @@ static unsigned serializeCounter = 0;
 // Call a RIR function. Arguments are still untouched.
 RIR_INLINE SEXP rirCall(CallContext& call, InterpreterInstance* ctx) {
     SEXP body = BODY(call.callee);
-    bool bodyPreserved = false;
+    Protect p;
```
o-, line 923: that thing is probably a bit slow and heavy to use in one of the hottest-path functions in the interpreter. afaik it creates some stl container....

JanJecmen: It just keeps an integer and calls unprotect in the destructor

rir/src/compiler/native/builtins.cpp
```diff
@@ -71,41 +71,37 @@ SEXP createEnvironmentImpl(SEXP parent, SEXP arglist, int contextPos) {
...
     SEXP res = Rf_NewEnvironment(R_NilValue, arglist, parent);
+    PROTECT(res);
 
     if (contextPos > 0) {
         if (auto cptr = getFunctionContext(contextPos - 1)) {
             cptr->cloenv = res;
             if (cptr->promargs == symbol::delayedArglist) {
-                auto promargs = arglist;
-                bool hasMissing = false;
+                auto promargs = R_NilValue;
```
o-, line 80: it seems to me that previously we would share the promargs list with the environment in most cases, whereas now they are typically two lists. so this leads to massively more allocations, not?

rir/src/compiler/pir/instruction.h
```diff
@@ -1138,6 +1139,21 @@ class FLIE(MkArg, 2, Effects::None()) {
     bool usesPromEnv() const;
 };
 
+class FLIE(FakeMkArg, 2, Effects::None()) {
```
o-, line 1142: I still claim that this is not a good idea. They should be normal MkArg's and I don't understand why we have these. In the worst case just compile them again...

rir/src/compiler/pir/instruction.cpp
```diff
@@ -1040,25 +1055,53 @@ ClosureVersion* StaticCall::tryOptimisticDispatch() const {
...
     }
+    while (expandedArgs.size() < argOrderOrig.size())
+        expandedArgs.push_back(MissingArg::instance());
+
+    size_t j = 0;
+    for (auto i : argOrderOrig) {
+        i = BC::decodeArgOrder(i);
+        assert(i < expandedArgs.size());
+        if (auto mk = MkArg::Cast(expandedArgs[i])) {
+            promises[j] = mk->prom()->rirSrc();
```
o-, line 1099: wow, so promises has a different order than args_, slightly confusing...

rir/src/interpreter/interp.cpp
o-: I see, still potentially slow due to the C++ obj being created...

o-:
I think there is an issue here that the original order is lost when we inline. When inlining the call instruction is removed and instead we put a `push_context` there. This instruction creates a context with a `delayedArglist` symbol. Then finally when we encounter `mk_env`, we update the arglist on the context with the same list of arguments like the mkenv has. Which is wrong in the case of reordered args.

I think the clean solution would be to remove all this `delayedArglist` nonsense. Instead the `push_context` should take a variable number of args, receive all the elements of the arglist as parameters and finally create a `ArgsLazyData` and set that on the context.

rir/src/compiler/native/builtins.cpp
o-: actually the `delayedArglist` stuff should probably be removed. see comment below

o-:
though maybe we should do the above separately and for now just ensure we don't inline if the args are reordered?

o-:
> though maybe we should do the above separately and for now just ensure we don't inline if the args are reordered?

I can do that part, I think I know how.

PR #918, opened 2020-09-26
Check if inputs are Numbers for logical operators

StrongerXi:
In the R implementations of `||` and `&&`, the arguments are checked by `isNumber` before being applied to `asLogical`.
To enable a similar checking behavior in rir, we add in the check during interpretation of `aslogical_` instruction, since it's only used in compiling `||` and `&&`.

The added integration test cases demonstrate the expected behavior, which is currently missing.

### DIFF
--- a/rir/src/interpreter/interp.cpp
+++ b/rir/src/interpreter/interp.cpp
@@ -3248,6 +3248,10 @@ SEXP evalRirCode(Code* c, InterpreterInstance* ctx, SEXP env,
 
         INSTRUCTION(aslogical_) {
             SEXP val = ostack_top(ctx);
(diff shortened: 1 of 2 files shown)


o-:
that looks good. now this bytecode also gets translated to the pir instruction `AsLogical` (see rir2pir.cpp) and then lowered to native in `compiler/native/lower_llvm.cpp`.  There are 3 cases: if the input is an unboxed integer or real nothing needs to be done. otherwise the native code relies on a native builtin called `asLogicalBlt`, which you find in `native/builtins.cpp`, which needs the same patch.

StrongerXi:
Migrating this PR to another one with branch created in upstream repo instead of forked repo.

Issue #906, opened 2020-09-22
support for `switch`


o-:
See #872 for a hacky first try. This special should be supported in the rir compiler already.

PR #920, opened 2020-10-02
Inline `switch` special function call in rir

StrongerXi:
This PR attempts to solve #906 
Currently `switch` special function call is dispatched to the built-in implementation in R.
Inlining it in rir enables more analysis and optimizations in pir.


rir/src/ir/Compiler.cpp
```diff
@@ -1072,6 +1072,150 @@ bool compileSpecialCall(CompilerContext& ctx, SEXP ast, SEXP fun, SEXP args_,
...
+                argMissing[argIdx] = false;
+            }
+        }
+        printf("compiling switch\n");
+        Rf_PrintValue(ast);
+
+        /******************* INSTRUCTIONS START HERE *********************/
+        compileExpr(ctx, args[0]);
+        // the inlined list of types is based on definition of `isVector` in `Rinlinedfuns.h`
+        for (auto type : {LGLSXP, INTSXP, REALSXP, CPLXSXP, STRSXP,
```
StrongerXi: I tried adding support for `BC::isType(TypeChecks::Vector)` (also for `String` and `Factor`), and handle it accordingly in `interp.cpp`. However I found that the `isType` instruction currently only comes from pir, so I decided to refrain from that and use `BC::is` instead.

StrongerXi:
Notice the `TODO`s in comments point out several places where we need to insert an `errorcall` instruction, which is currently not supported. (the current behavior is to either not inline the call, or return `Nil`. Maybe I should open up another PR to work on that, and then finish up this one?

o-:
this looks reasonable to me on the first glance. I didn't check the instructions you emit in detail, let me know if you get stuck on fixing some of the broken tests, or need help in reproducing them locally. it can be tricky at times.

for the error cases: you could use the same trick as on https://github.com/reactorlabs/rir/pull/920/files#diff-e7dea8cf4dd72c5c6e070769bc4adc22R1172 and use callBuiltin to call the error builtin, that should do the trick without adding any bytecodes.

StrongerXi:
I should've mentioned. I did try using `BC::callBuiltin`, but the `errorcall` function is not a builtin apparently, i.e. it's not in `R_FunTab`, and in R they only invoke it as a helper function in the implementations of other builtin.

o-:
> I should've mentioned. I did try using `BC::callBuiltin`, but the `errorcall` function is not a builtin apparently, i.e. it's not in `R_FunTab`, and in R they only invoke it as a helper function in the implementations of other builtin.

ah, you are way ahead of me :) try the `stop` internal

StrongerXi:
Rebasing onto master to resolve travis ci build failure.

o-:
@StrongerXi hey, what is the status? I see there is still a todo in there. can I help with something? would be cool if we can merge it soon!

StrongerXi:
Hi @o- . 

There are 2 todos:
1. A corner case that involves coercing `RAWSXP` to `INTSXP`. Handling this would probably require adding an `asInteger` instruction. The corner case hasn't shown up in any test, but I can try adding that instruction if you deem it necessary.
2. I couldn't simulate the `isFactor` check. I tried (a) using the builtin `inherit` function, (b) adding `TypeCHecks::Factor` and handle it in `BC::isType`. Both fail with different degree of errors, as documented in the code.

Currently only Travis fails with a time out error, but I'm not sure how to resolve that.

o-:
the test that we have are really not that extensive. so if you know about issue, we should fix them and if possible add tests that check them. it's unfortunately easy to pass the tests with broken things...

if you give me examples of switches that still fail I can help to debug, I have some free time again...

PR #930, opened 2020-11-10
Fix force dominance

JanJecmen:
Simply removing an UpdatePromise would lose information about the argument being forced, which could lead to inlining it multiple times. This commit instead replaced the UpdatePromise with a fresh eager MkArg.


o-:
https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids%5B%5D=841733658&job_ids%5B%5D=840894685&kind=diff&warmup=5

JanJecmen:
Ah, I see, if an eager promise gets to a force through a phi, this doesn't get rid of it

o-:
oh, that's annoying. I could imagine this being a common problem for such "replacement" kind of optimizations...

JanJecmen:
superseeded by #933 and #934

PR #933, opened 2020-11-12
force dominance eager mkarg replacement fix

JanJecmen:
can only use the known eager value of a mkarg to replace a force (instead of the casttype, which could lead to the eager value escaping where the promise was expected, eg. into substitute)

### DIFF
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -440,10 +440,8 @@ bool ForceDominance::apply(Compiler&, ClosureVersion* cls, Code* code,
             if (auto f = Force::Cast(*ip)) {
                 if (auto mkarg = MkArg::Cast(f->followCastsAndForce())) {
                     if (mkarg->isEager()) {
-                        anyChange = true;
-                        Value* eager = mkarg->eagerArg();
-                        f->replaceUsesWith(eager);
-                        next = bb->remove(ip);
+                        assert(false &&
+                               "followCastsAndForce returned an eager MkArg");
                     } else if (toInline.count(f)) {
                         anyChange = true;
                         Promise* prom = mkarg->prom();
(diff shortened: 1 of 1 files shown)


rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -578,15 +576,29 @@ bool ForceDominance::apply(Compiler&, ClosureVersion* cls, Code* code,
...
                     if (mk->isEager()) {
                         anyChange = true;
-                        auto eager = mk->eagerArg();
-                        cast->replaceUsesWith(eager);
+                        Value* eager = mk->eagerArg();
+                        f->replaceUsesWith(eager);
                         next = bb->remove(ip);
                     }
+                } else if (auto phi = Phi::Cast(f->followCastsAndForce())) {
+                    phi->eachArg([&](BB* bb, InstrArg& v) {
```
o-: this is not correct. think about:
```
p = phi(a,b)
lazyUse(p)
force(p)
```

here you will break the lazy use

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -440,10 +440,8 @@ bool ForceDominance::apply(Compiler&, ClosureVersion* cls, Code* code,
             if (auto f = Force::Cast(*ip)) {
                 if (auto mkarg = MkArg::Cast(f->followCastsAndForce())) {
                     if (mkarg->isEager()) {
-                        anyChange = true;
-                        Value* eager = mkarg->eagerArg();
-                        f->replaceUsesWith(eager);
-                        next = bb->remove(ip);
+                        assert(false &&
```
o-, line 443: I don't see why we would want this invariant?

rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -578,15 +576,29 @@ bool ForceDominance::apply(Compiler&, ClosureVersion* cls, Code* code,
                             dead.insert(split);
                         break;
                     }
-                }
-            } else if (auto cast = CastType::Cast(*ip)) {
-                if (auto mk = MkArg::Cast(cast->arg<0>().val())) {
+                } else if (auto mk = MkArg::Cast(f->input()->followCasts())) {
```
o-, line 579: so this is
```
if (auto mkarg = MkArg::Cast(f->followCastsAndForce())) {
...
} else if (auto mk = MkArg::Cast(f->input()->followCasts())) {
...
}
```
In other words this happens *never*....

o-: imho this is (a) unreadable and (b) should be moved outside that if. any `force(cast*(mkarg(eager)))` should be replaced by `eager`

JanJecmen: the problem is that `followCastsAndForce` doesnt't stop at `MkArg` if it's eager but returns the eager value, whereas the `followCasts` returns the `MkArg`

JanJecmen: so the first invariant above is something that never happens with the current implementation but if that changes we would want to fix it

o-: I see. it is still unreadable

o-: also unfortunately not quite correct. if you have a downcast there it would not be correct to replace it, wouldn't it?

o-: imho what we want is:
```
if (auto f = Force::Cast(i)) {
  if (!f->input()->maybePromiseWrapped()) {
    f->replaceUsesWith(f->input()));
  } else {
    if (auto c = CastType::Cast(f->input())) {
      // An upcast from a value to a lazyValue is the cue that we can unwrap mkarg
      if (c->type == CastType::UpCast && !c->input()->type.maybePromiseWrapped()) {
        if (auto mk = MkArg::Cast(f->input()->followCasts())) {
          if (mk->isEager()) {
            f->replaceUsesWith(mk->eager());
          }
        }
      }
    }
  }
}
```

o-:
I believe our intuition on how the force dominance pass works here was totally off.

I added a summary [here](https://github.com/reactorlabs/rir/pull/946/files#diff-e25f9c095374b9ee89c34c7b4095ffdada4e70334d2b107d12abeda879a5d7bbR385). This is mostly how it already works on master. As you can see, most of the cases that you are trying to fix here seem to be already addressed.

I think what you saw was either push contexts being updated, when they shouldn't ([fixed here](https://github.com/reactorlabs/rir/pull/946/files#diff-e25f9c095374b9ee89c34c7b4095ffdada4e70334d2b107d12abeda879a5d7bbR678)) or the scope analysis not dealing with updatePromise ([fixed here](https://github.com/reactorlabs/rir/pull/946/files#diff-38fbf4a898fe672036ee14f50bd3d1149bef72684660711b5a4cceb9cae2ea5fR227))

PR #934, opened 2020-11-12
force dominance update promise removal fix

JanJecmen:
even if an update promise doesn't escape, we still need to know that the promise has been forced / inlined, so we either don't remove the update promise or create a fresh mkarg

### DIFF
--- a/rir/src/compiler/opt/force_dominance.cpp
+++ b/rir/src/compiler/opt/force_dominance.cpp
@@ -47,6 +47,7 @@ using namespace rir::pir;
 
 struct ForcedBy {
     rir::SmallMap<Value*, Force*> forcedBy;
+    rir::SmallMap<MkArg*, rir::SmallSet<CastType*>> promToValueCast;
     rir::SmallSet<Value*> inScope;
     rir::SmallSet<Value*> escaped;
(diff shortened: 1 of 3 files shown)


rir/src/compiler/opt/force_dominance.cpp
```diff
@@ -177,6 +190,20 @@ struct ForcedBy {
             }
         }
 
+        for (auto& mc : other.promToValueCast) {
+            if (promToValueCast.contains(mc.first)) {
```
o-: this is slow and causes analysis times to explode.

please use `p = promToValueCast.find(mc.first)`, then you can check with `p == promToValueCast.end()` and access with `p->second`

rir/src/compiler/pir/instruction.h
```diff
@@ -1256,6 +1256,9 @@ class FLI(CastType, 1, Effects::None()) {
         }
         return type;
     }
+    bool promToValue() const {
+        return kind == Kind::Upcast && MkArg::Cast(arg(0).val());
```
o-: you could return `Mkarg*` (or null for false), that might be cleaner for the client...

PR #938, opened 2020-11-26
builtins and fixes

o-:
* support fast builtins when they involve trivial promise args
* add/extend builtins
* fix length builtin


rir/src/interpreter/builtins.cpp
```diff
@@ -519,13 +550,49 @@ SEXP tryFastBuiltinCall(const CallContext& call, InterpreterInstance* ctx) {
     }
 
     case blt("is.vector"): {
-        if (nargs != 1)
+        if (nargs < 1 && nargs > 2)
```
JanJecmen: `||`?

o-: pretty cool that cppcheck also finds it: https://gitlab.com/rirvm/rir_mirror/-/jobs/874735614

PR #951, opened 2021-01-12
Argument reordering

JanJecmen:


JanJecmen:
I did some measuring, this is for the `Simple.R` test from the `Matrix` package. Create promargs, create environment frame and create lazy arglist are the entrypoints to createLegacyArglist, so these add up to be the same. The reordered entry is how many createLegacyArglist calls were reordering the arguments.
```
                                          this branch          master

                               time       20.2388 m            8.64827 m
                    compileBC calls       1M (1182730)         756K (756544)
                     new StaticCall       21K (21453)          15K (15872)
                    create promargs       129K (129138)        130K (130376)
           create environment frame       19K (19561)          22K (22212)
                create lazy arglist       3K (3581)            1K (1244)
          createLegacyArglist calls       152K (152280)        153K (153832)
           recreateOriginalPromargs       3K (3581)            1K (1244)
                      reordered           2K (2341)            NA
```

rir/src/runtime/ArglistOrder.h
```diff
@@ -0,0 +1,86 @@
...
+constexpr static size_t ARGLIST_ORDER_MAGIC = 0xaabbccdd;
+
+struct Code;
+
+struct ArglistOrder
+    : public RirRuntimeObject<ArglistOrder, ARGLIST_ORDER_MAGIC> {
+
+    using CallId = size_t;
+    using ArgIdx = size_t;
+    using CallArglistOrder = std::vector<ArgIdx>;
```
o-, line 23: unfortunately this will leak, since RirRuntimeObjects do not have their distructors called when the R GC collects them...

o-: ah, gotcha :)

PR #952, opened 2021-01-13
[wip] User Annotated Contexts

o-:


rir/R/rir.R
```diff
@@ -6,14 +6,16 @@ rir.markFunction <- function(what, which,
...
+                             Annotated=NA) {
     doIt = function(n) {
         .Call("rirMarkFunction", what, n,
               Reopt,
               ForceInline, DisableInline,
               DisableAllSpecialization,
               DisableArgumentTypeSpecialization,
-              DisableNumArgumentsSepcialization);
+              DisableNumArgumentsSepcialization,
+              Annotated);
```
o-: Can we rename this to use something meaningful? DepromisedArgs, e.g.?

rir/src/api.cpp
```diff
@@ -119,6 +120,7 @@ REXPORT SEXP rirMarkFunction(SEXP what, SEXP which, SEXP reopt_,
         getBool(disableNumArgumentSpecialization_);
     auto disableArgumentTypeSpecialization =
         getBool(disableArgumentTypeSpecialization_);
+    auto annotated = getBool(annotated_);
```
o-: dito

rir/src/interpreter/call_context.h
```diff
@@ -58,7 +58,7 @@ struct CallContext {
     const Code* caller;
     const size_t suppliedArgs;
     size_t passedArgs;
-    const R_bcstack_t* stackArgs;
+    /*const */ R_bcstack_t* stackArgs;
```
o-: I don't understand what the commented out const means

o- (commented):
so what about using the depromised flag for optimizations? that seems to be missing

rir/R/rir.R
skrynski: sure

rir/src/interpreter/call_context.h
skrynski: I removed the 'const' to be able to replace forced promises on the stack with the unwrapped values. Should I remove this for good or add a comment explaining why it is being commented out?

rir/src/compiler/opt/inline_force_prom.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+            if (auto call = CallInstruction::CastCall(*ip)) {
+
+                auto cls = call->tryGetCls();
+
+                if (cls) {
+
+                    auto functionVersion = cls->rirFunction();
+                    if (functionVersion->flags.contains(
+                            rir::Function::Flag::Annotated)) {
```
skrynski: @o-  here is the use of the flag for optimizations. Is this what you were referring to in your comment?

rir/src/interpreter/call_context.h
o-: I mean if it is not const, then it is not const... I guess I don't understand the reason for commenting out instead of deleting. or what kind of explanation would you write in the coment, other than "it is not const"...

rir/src/compiler/opt/inline_force_prom.cpp
o-: no. I mean that the type of function arguments should be adjusted for functions with the depromised annotation

rir/src/interpreter/call_context.h
skrynski: got it

.vscode/tasks.json
```diff
@@ -4,14 +4,16 @@
     "version": "2.0.0",
     "tasks": [
         {
-            
-            "label": "compile debugopt",
+            "label": "compile debug",
             "type": "shell",
-            "command": "cmake --build .",
```
o-, line 10: I am not using vscode, so I don't really care. but this will fail when you run it on an empty repo. because cmake by default generates Makefiles, but then you use ninja and not make....

if you want to use ninja, you need `-GNinja` added to cmake. and anyway `cmake --build` is better than `ninja` because it automatically picks the build system you have chosen.

also debug build is really slow. maybe debugopt?

PR #957, opened 2021-01-20
Measuring

JanJecmen:
I have been using this recently. It's a global event / time counter extracted from `compiler/compiler.cpp` and extended a bit, with an easy to use api: just `#include "utils/measuring.h"` and count events by `Measuring::countEvent("what")` etc. Runing with `PIR_MEASURING=1` will then dump the counters to `std::cerr` when the program ends.

Do we want it? There is already the code I based this on, plus some similar functionality in `event_counters.h` and `interp.cpp` (the `SlowcaseCounter`)...


o-:
> Do we want it? There is already the code I based this on, plus some similar functionality

maybe unify all of these?

PR #1011, opened 2021-04-19
speculate on call targets later in the pipeline

o-:
currently we only speculate on call targets in rir2pir. this change
inserts guards late in the pipeline, that should allow the existing
match arg pass to turn them into static calls.


o-:

* WARNING: the longest CI job tests_debug2 took 2.41h
* The real_thing suite improved by 1.06
* The real_thing_annotations suite regressed by 0.97
* The shootout_annotations suite improved by 1.05

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1223221392&job_ids[]=1229753391

PR #1016, opened 2021-04-21
Missing bug

JanJecmen:


rir/src/compiler/opt/scope_resolution.cpp
```diff
@@ -339,18 +338,19 @@ bool ScopeResolution::apply(Compiler&, ClosureVersion* cls, Code* code,
                         auto v =
                             res.result.singleValue().val->followCastsAndForce();
                         if (!v->type.maybePromiseWrapped() &&
-                            !v->type.maybeMissing()) {
+                            !v->type.maybeMissing() &&
+                            /* Warning: Forcing a (non-missing) promise can
+                                still return missing... */
+                            !MkArg::Cast(v)) {
```
o-, line 344: there seems to be a regression in reversecomplement_2 https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids%5B%5D=1199929943&job_ids%5B%5D=1191721056&kind=diff&selection=main&warmup=5

don't know if it is a flake or real...

an idea: if the mkarg is eager, then you should be able to tell if it is missing or not...

PR #1021, opened 2021-04-28
Revert "speedup loading from baseEnv"

o-:
Reverts reactorlabs/rir#1019

trying to see if that is what made debug_2 ci test so slow


o-:
the perf regression looks like a flake after all

PR #1026, opened 2021-05-06
towards more rir -> pir calls

o-:
this PR takes another step towards being able to call more optimized
functions from unoptimized ones. Here we mark calls, where `...` is a
formal argument, but the caller does not passes not enough args for any
of them to end up in the `...` as statically matched, ie. when `...` is
missing.

### DIFF
--- a/rir/src/compiler/native/builtins.cpp
+++ b/rir/src/compiler/native/builtins.cpp
@@ -169,6 +169,8 @@ void stargImpl(SEXP sym, SEXP val, SEXP env) {
             } else {
                 ENSURE_NAMED(val);
             }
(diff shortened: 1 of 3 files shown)


o-:

* WARNING: the longest CI job test_features_1 took 3.36h
* The real_thing suite regressed by 0.97
* The real_thing_annotations suite improved by 1.05
* The shootout_annotations suite improved by 1.03
* The simple suite regressed by 0.98

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1283456818&job_ids[]=1282826854

PR #1052, opened 2021-06-09
Prevent segfault in rir compilation when an undefined function is called as Internal

vogr:
Currently, the RIR compilation of `Sys.timezone` fails with a segfault. This is due to the line

```
if(.Platform$OS.type == "windows") return(.Internal(tzone_name()))
```
in its definition: the function `tzone_name` is only mapped to the C function `do_tzone_name` on win32 platforms.

More generally, any .Internal call to an undefined function results to a segfault during the compilation to rir
```
# fails to compile
f <- function()
    .Internal(undefined_function())
```

Such a call should result in a runtime error and not in a segfault during compilation.


o-:

* WARNING: the longest CI job tests_debug2 took 2.82h
* The real_thing_annotations suite improved by 1.02

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1330856389&job_ids[]=1327742939

PR #1065, opened 2021-06-18
Try avoiding unnecessary analysis state recomputation

JanJecmen:
* The state would be recomputed even with the keepSnapshot flag
* (experimental) Disable the lookup cache

### DIFF
--- a/rir/src/compiler/analysis/available_checkpoints.h
+++ b/rir/src/compiler/analysis/available_checkpoints.h
@@ -37,9 +37,7 @@ class FwdAvailableCheckpoints
         return AvailableCheckpointsApply::apply(state, i);
     }
 
-    Checkpoint* reaching(Instruction* i) const {
-        return StaticAnalysis::at<PositioningStyle::BeforeInstruction>(i).get();
-    }
+    Checkpoint* reaching(Instruction* i) const { return before(i).get(); }
 };
 
 class RwdAvailableCheckpoints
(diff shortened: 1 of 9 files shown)


o-:
this is massively faster. so let's merge it!

JanJecmen:
Is it? The benchmarks see no change, that's expected.. but the couple of CI jobs I checked seem slower. Just timing the check code regression test seems faster, but then again it wildly fluctuates so it's hard to say

o-:
really the granularity of comparing CI jobs is hours. they are just too noisy to compare that. you need independent experiments.

I was playing with MEASURE_COMPILER on regression7 and it straight up decreased the time spent in scope resolution by 1/3. that's when I decided to merge since my other small improvements just didn't make sense without that huge baseline shift.

re. the cache, i really don't know if it helps.... for the sake of reducing complexity, maybe we could just remove it....

I implemented this: https://github.com/reactorlabs/rir/pull/1070/files?diff=split&w=1#diff-2d24e9abbd080ba4f08ad46e3c25218624ebd6871239c84315d3ff72c73c2578R314 and I think its the better approach. Here you can pass the state of the previous instruction to the analysis query and it will reuse it if it makes sense. it's a bit tricky because you have to not do it when you changed the instructions in the bb or you won't get good results... but anyway it is simple and fits many usecases, where we go linearly through the instructions.

PR #1096, opened 2021-09-03
better change tracking in constantfold and gvn

o-:
With LdConst removed there is less noise in our optimization passes and
we see aborted opt schedules due to passes not reporting when they
change stuff.

This should fix the vector-for regression.


o-:

* WARNING: the longest CI job test_release_2 took 2.66h
* vectorWhile regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1562359582&job_ids[]=1562350929&selection=all

PR #1103, opened 2021-09-10
Prototype for deoptimizationless

o-:
This prototype implements the deoptimizationless idea. It is restricted
to the situation where we have only one frame, and a lazy environment.
Then it compiles a continuation for the deoptimization context. This
context consists of: the pc where the continuation starts, the types of
the values on the stack, and the types and names of the values in the
lazy env.

Many things are still missing here:
* continuations are only used once and then thrown away. we need some
  sort of table for them and a dispatching mechanism such that we can
  use them multiple times. The deopt context should be usable as the
  index in such a table.
* The continuation compilation uses all the type-feedback normally. This
  leads to "follow-up" deoptimizations, because only the type-feedback
  for the immediate deoptimization reason is updated. We should have a
  type-feedback propagation pass, that tries to invalidate feedaback
  that is likely affected by the original deopt reason. E.g. if we deopt
  because the type of `a` in `a+b` changed, then we must also generalize
  the expected type of the expression `a+b` and not just `a`.
* No clue how to compile continuations for inlined functions yet. that
  would be fancy, but I have to think about it on paper first.
* We pass the contents of the deopt-env on the stack and the
  continuation creates a new stub-env with the values from the stack.
  this is kinda inefficient and it would be better to re-use the
  stub-env that we already have. The reason I did it this way is that
  the continuation has a normal `MkEnv` instruction that all our
  optimization passes know about. We would have to refactor quite a lot
  to make it work properly. Or we could also remove that `MkEnv` with a
  new `loadDeoptEnv` instruction while lowering.


o-:

* WARNING: the longest CI job test_release_2 took 2.98h
* are-we-fast-r suite regressed by 0.98
* simple suite improved by 1.07
* Bounce_nonames regressed by 0.93
* Bounce regressed by 0.94
* spectralnorm_alt_3 regressed by 0.93
* matrixFor regressed by 0.9

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1631199938&job_ids[]=1631185608&selection=all

rir/src/compiler/native/builtins.cpp
```diff
@@ -884,18 +889,25 @@ void deoptImpl(rir::Code* c, SEXP cls, DeoptMetadata* m, R_bcstack_t* args,
                         cmp.optimizeModule();
 
                         fun = backend.getOrCompile(cnt);
-                        Protect p(fun->container());
+                        PROTECT(fun->container());
```
JanJecmen: Missing `UNPROTECT(1)`?

JanJecmen: oh nevermind, findcontext takes care of it..

o-: yeah, indeed. but it is bound to cause issues at some point down the line... So I'll at least add a comment or something...

o-:

* WARNING: the longest CI job test_big_inline took 2.67h
* reversecomplement_2 regressed by 0.91
* reversecomplement regressed by 0.95
* lapply regressed by 0.92

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1662127495&job_ids[]=1661399097&selection=all

rir/src/compiler/opt/typefeedback_cleanup.cpp
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                }
+            }
+            i->eachArg([&](Value* v) {
+                if (auto vi = Instruction::Cast(v)) {
+                    if (affected.count(vi)) {
+                        affected.insert(i);
+                        if (i->hasTypeFeedback()) {
+                            auto inferred = i->inferType([&](Value* v) {
+                                if (auto vi = Instruction::Cast(v)) {
```
skrynski: silly comment: vi has same name as outer variable.

rir/src/compiler/opt/typefeedback_cleanup.cpp
```diff
@@ -0,0 +1,97 @@
...
+                        return;
+                    }
+                }
+            }
+            i->eachArg([&](Value* v) {
+                if (auto vi = Instruction::Cast(v)) {
+                    if (affected.count(vi)) {
+                        affected.insert(i);
+                        if (i->hasTypeFeedback()) {
+                            auto inferred = i->inferType([&](Value* v) {
```
skrynski: wouldn't it be faster to first adjust the types  on each arg (i->eachArg([&](Value* v) ) and then call i->inferType() only once  ? Maybe there is some kind of closed feedback loop I am missing

o-:

* WARNING: the longest CI job test_big_inline took 2.63h
* simple suite improved by 1.07
* reversecomplement_2 regressed by 0.94
* binarytrees_2 regressed by 0.94
* emptyFor regressed by 0.88
* Overall benchmarks improved by 1.02

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1663089861&job_ids[]=1661399097&selection=all

PR #1105, opened 2021-09-29
missing protect

o-:
The BC::push(args[0]) can cause the constantpool to grow which causes
the function to be collected before it can be added to the pool.


o-:

* WARNING: the longest CI job test_big_inline took 2.29h
* are-we-fast-r suite regressed by 0.96
* Bounce_nonames_simple regressed by 0.87
* pidigits regressed by 0.94
* scalarFor regressed by 0.89

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1632771953&job_ids[]=1631185608&selection=all

PR #1115, opened 2021-10-13
Fix naked dots miscompilation

JanJecmen:
In rir, we use `R_DotsSymbol` for marking an argument that should be expanded at runtime, but also when we need to push the actual symbol (in `quote(...)`).
In pir we don't know anymore which case this was, and so try to always expand the dots.
This adds a symbol placeholder `.expandDotsTrigger.` to be used in the arglists instead.

### DIFF
--- a/rir/src/R/Printing.cpp
+++ b/rir/src/R/Printing.cpp
@@ -1,6 +1,7 @@
-#include "Printing.h"
-
+#include "R/Printing.h"
 #include "R/Funtab.h"
+#include "R/Symbols.h"
+
 #include <iostream>
 #include <string>
 
@@ -11,6 +12,8 @@ std::string Print::dumpSexp(SEXP src, size_t length) {
         return "R_UnboundValue";
     } else if (src == R_MissingArg) {
         return "R_MissingArg";
+    } else if (src == symbol::expandDotsTrigger) {
+        return "`...`";
     }
 
     static auto deparseBlt = getBuiltinFun("deparse");
--- a/rir/src/R/Protect.h
+++ b/rir/src/R/Protect.h
@@ -9,13 +9,13 @@ namespace rir {
 
 class Protect {
   public:
-    Protect(const Protect& other) = delete;
-
     Protect() {}
     explicit Protect(SEXP init) {
         Rf_protect(init);
         ++protectedValues_;
     }
+    Protect(const Protect& other) = delete;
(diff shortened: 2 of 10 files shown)


o-:
good catch!

The information would be there, namely when the `name` of the argument is R_DotsSymbol, only then it needs to be expanded. I like your solution more, so let's keep it. But it would be good to remove the name of the argument and then fix all places where we check the name, e.g. (citing from memory) in the argListAdapters and potentially other places too. Because otherwise we have two competing mechanisms and this is guaranteed to cause trouble.

o-:

* WARNING: the longest CI job test_release_2 took 2.21h
* simple suite regressed by 0.98
* Bounce_nonames regressed by 0.95
* reversecomplement_2 regressed by 0.87
* scalarFor regressed by 0.86
* emptyFor regressed by 0.88
* matrixFor regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1675233590&job_ids[]=1667952905&selection=all

o-:

* WARNING: the longest CI job test_big_inline took 3.1h
* fasta_naive regressed by 0.94
* scalarFor regressed by 0.86
* emptyFor regressed by 0.9

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1682491550&job_ids[]=1667952905&selection=all

JanJecmen:
I think we need to use both the name and the value placeholders:
1/ When we compile rir to pir, we need to differentiate between `R_DotsSymbol` that should be pushed on the stack as is, and that should be expanded when creating arglist for a call (here we don't have the name easily accessible)
2/ When we actually expand dots at runtime, the native backend will have already loaded the `...` to the stack, whereas the interpreter still has to look it up (here the value placeholder can't be used since the actual dotlist may already be on the stack)

Also, the placeholder symbol has to be serializable. This uses a normal R symbol with unprintable chars which lowers the chances of it being accidentally used from R (but it's still possible: `as.name("\x1a")`). Another option would be to make it an `EXTERNALSXP`.

o-:

* WARNING: the longest CI job test_release_2 took 2.23h

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1684401317&job_ids[]=1667952905&selection=all

PR #1118, opened 2021-10-20
fix compilation of nested loops

o-:
When we have nested simple loops, then a break or continue in the body
of an inner loop caused the loops to have contexts. The reason is that
when we compile the fallback case for the simple loop, then we forgot to
set up a loop context. This means that the break being compiled for
fallback would taint the wrong (ie. outer) loop context.

### DIFF
--- a/rir/src/ir/Compiler.cpp
+++ b/rir/src/ir/Compiler.cpp
@@ -166,6 +166,11 @@ class CompilerContext {
         code.top()->loops.emplace(next_, break_);
     }
 
+    void pushFakeLoop() {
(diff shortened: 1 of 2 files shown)


o-:

* WARNING: the longest CI job test_big_inline took 2.28h
* are-we-fast-r suite improved by 1.02
* reversecomplement_naive regressed by 0.92

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1699495081&job_ids[]=1689869075&selection=all

PR #1120, opened 2021-10-25
Osr on by default

o-:

PR #1121, opened 2021-10-26
Some fixes for deoptless

o-:
Based on osrOn branch, ie. fixes with OSR enabled


o-:
(integrated into #1120)

PR #1122, opened 2021-10-26
Fix inner function guards

o-:
When speculating on callees we should respect the inner function flag
and compile the function such that it does not depend on the environment
being stable!


o-:

* WARNING: the longest CI job test_big_inline took 2.65h
* simple suite improved by 1.12
* summary suite improved by 1.03
* spectralnorm_alt_3 regressed by 0.92
* Overall benchmarks improved by 1.03

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1725315007&job_ids[]=1716366108&selection=all

PR #1126, opened 2021-11-01
fix type checks for scalars

o-:
the noWrappedMissing flag would cause scalar checks to accidentally end up in the slow generic case. 

plus adding support for non-na checks


o-:

* WARNING: the longest CI job test_features_1 took 3.05h
* are-we-fast-r suite improved by 1.06

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1736672364&job_ids[]=1735284480&selection=all

PR #1133, opened 2021-11-08
Fix leak env analysis

JanJecmen:
Promises and closures can leak their environments but by themselves they don't leak anything, only when they are used (forced / called). Our dead store removal was imprecise here, so it would break code like the following, removing the store to `n`, even though it's observable through calling the closure:
```
function() {
  n <- 5
  function() n
}
```

### DIFF
--- a/rir/R/rir.R
+++ b/rir/R/rir.R
@@ -81,7 +81,7 @@ pir.tests <- function() {
     invisible(.Call("pirTests"))
 }
 
-# returns TRUE f, when PIR compiled, satisfies the the given checks (e.g.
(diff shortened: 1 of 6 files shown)


o-:

* WARNING: the longest CI job test_big_inline took 2.4h
* are-we-fast-r suite improved by 1.03
* spectralnorm_alt_3 regressed by 0.94
* fasta_2 regressed by 0.95
* emptyFor regressed by 0.87
* matrixFor regressed by 0.92
* vectorFor regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=1760173431&job_ids[]=1756554539&selection=all

PR #1173, opened 2022-01-31
optimize functions compiled late in the pipeline

o-:
if an match_call_args discovers and compiles a new function, then it is
way less optimized than others, because it is added to the module, while
it is in the middle of being optimized.

This is a commit to optimize that function in isolation, to catch
up...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -80,7 +80,7 @@ test_release_1:
   variables:
     GIT_STRATEGY: none
     PIR_LLVM_OPT_LEVEL: 0
-    PIR_OPT_LEVEL: 1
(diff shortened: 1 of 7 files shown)


o-:

* WARNING: the longest CI job test_release_2 took 2.35h
* are-we-fast-r suite regressed by 0.98
* Storage regressed by 0.89
* nbody_naive regressed by 0.75
* fasta_2 regressed by 0.93
* profiler-microbenchmark regressed by 0.95

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2048346430&job_ids[]=2043447063&selection=all

PR #1177, opened 2022-02-15
Support for deopts in promises

JanJecmen:
If there is a deoptimization in a promise (that is also the outermost frame, ie.,
forcing a promise that isn't inlined), we cannot longjump out of the optimized
execution, since the promise doesn't have its own context.

Thus we add a return instruction immediately after the deopt, which is only reached
if there was no longjump. In that case, we know that this is the evaluation of
a promise that deoptimized, and the result is left on the stack to be returned.

_TODO: actually use speculation in promises (most passes don't run on promises as of now)_

### DIFF
--- a/rir/src/compiler/analysis/context_stack.h
+++ b/rir/src/compiler/analysis/context_stack.h
@@ -15,7 +15,7 @@ struct ContextStackState {
             if (auto mk = MkEnv::Cast((*i)->env()))
                 it(mk);
     }
-    size_t context() const { return contextStack.size(); }
+    size_t numContexts() const { return contextStack.size(); }
     AbstractResult merge(const ContextStackState& other) {
         assert(contextStack.size() == other.contextStack.size() &&
                "stack imbalance");
--- a/rir/src/compiler/analysis/verifier.cpp
+++ b/rir/src/compiler/analysis/verifier.cpp
@@ -35,7 +35,7 @@ class TheVerifier {
     std::unordered_set<BB*> seenPreds;
 
     void operator()() {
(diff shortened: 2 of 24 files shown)


o-:
Watch out for this annotation here which needs to be removed, or actually only put there if it is a non-promise: https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/native/builtins.cpp#L2448

JanJecmen:
Oh, I just removed it.. maybe have two versions of deoptImpl then, one for promises and one for closures

o-:
> Oh, I just removed it.. maybe have two versions of deoptImpl then, one for promises and one for closures

sorry I somehow completely missed that you did already.

but yeah, it should help llvm with optimizations. ~you can also just add attributes in code, no need to have two versions.~ (of course not, its on the function, not the call...)

maybe drop the attribute, but keep the unreachable for the cases where it is not a promise? that should be equally good as the no return annotation.

o-:

* WARNING: the longest CI job test_features_3 took 2.59h
* fastaredux_naive regressed by 0.9
* lapplyDots regressed by 0.91
* matrixFor regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2479130138&job_ids[]=2467128976&selection=all

o-:

* WARNING: the longest CI job test_big_inline took 3.29h
* Bounce_nonames_simple regressed by 0.93
* nbody_naive regressed by 0.74
* emptyFor regressed by 0.86

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2599407256&job_ids[]=2578982488&selection=all

PR #1188, opened 2022-03-22
Fix a crash involving switch() and NA_character_

### LINKED ISSUE #1187: Weird switch behavior
We discovered that the test suite would fail on locales other than en_US.UTF-8 . Oli eventually tracked it down to a bug on our implementation of the `switch`  function:

```R
f= function() {print(switch(NA_character_, `x`="1", "2"))}
f()
f()
```

The first call to `f()` returns `"2"` as expected, but the next call results in an error:

```
Error in switch(NA_character_, x = "1", "2") : 
  missing value where TRUE/FALSE needed
```

hugomg:
Don't crash our implementation of switch() when receiving NA.
This also solves another issue where the test suite would crash in locales other than en_US.UTF-8.

Fixes #1187.

### DIFF
--- a/rir/src/bc/Compiler.cpp
+++ b/rir/src/bc/Compiler.cpp
@@ -1411,6 +1411,9 @@ bool compileSpecialCall(CompilerContext& ctx, SEXP ast, SEXP fun, SEXP args_,
          *   br nil
          *
          * str:
+         *   if (not is.na(value)) br strNACont
+         *   value <- "NA"
+         * strNACont:
          *   if (value == group[0]) br groupLabels[0]
          *   ...
          *   if (value == group[k]) br groupLabels[k]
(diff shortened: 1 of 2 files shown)


o-:

* WARNING: the longest CI job test_gctorture1 took 3.35h
* fasta_2 regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2235878303&job_ids[]=2225699243&selection=all

hugomg:
I don't have a good mental baseline for what those numbers mean, exactly... Do you think we need to investigate these benchmarks more closely before merging this? How long do those torture tests usually take?

o-:
The benchmarks look very much unaffected (not surprisingly). So that is certainly noise.

Why one of the CI jobs took so long to complete is unclear but I very much doubt it has anything todo with your change! These ci runners are on shared machines. The number is mainly reported back here to remind us when we start creeping up again how long the jobs take to complete... My aim was to keep them under 2h, which apparently we are failing to do again :)

o-:
Thanks a lot for fixing this properly!

PR #1191, opened 2022-03-24
fix CI x11 connection errors

o-:
The problem is that sometimes XOpenDisplay in devX11.c:1440 fails. It is
something racy, it only happens if the program is quick to re-open a new
X11 connection after closing one and the machine has a high load.

Seen at https://gitlab.freedesktop.org/mesa/waffle/-/merge_requests/93
the `-noreset` option seems to fix it...


hugomg:
Hope this does the trick! I suggest adding a comment pointing out that the -noreset is voodoo magic.

PR #1192, opened 2022-03-24
Some README improvements

hugomg:
Based on my adventures trying to compile RIR on Fedora (bad idea)

1. Strongly recommend Ubuntu, due to fetch-llvm
2. Tell how to install the compilation dependencies
3. Encourage off-tree builds and ninja by default
4. Gather the macOS instructions in a single place

[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,44 +1,86 @@
 # Ř
 
-To give it a spin try
+## Giving it as spin
+
+The easiest way to try Ř is using our pre-built docker container
 
     docker run -it registry.gitlab.com/rirvm/rir_mirror:master /opt/rir/build/release/bin/R
 
-## Building from Source
+## Building
 
-This is currently only tested on Ubuntu.
+If you want to build Ř from source, we strongly recommend that you use Ubuntu.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,44 +1,86 @@
 # Ř
 
-To give it a spin try
+## Giving it as spin
```
o-: *a* spin

README.md
```diff
@@ -1,44 +1,86 @@
...
+    # The cmake should output "Found LLVM 12"
+    cmake ..
+    ninja
+
+Congratulations! You can now run Ř with
 
     bin/R
 
-If you want to develop Ř then it is highly recommended to use off-tree builds and ninja instead of make. See further down (Hacking).
+### In-tree builds
```
o-: drop? is there any need for that?

o- (commented):
good stuff,  thanks

README.md
hugomg: Sounds reasonable, YAGNI. I suppose that if someone really wants in-tree builds, they are free to figure out the cmake commands themselves.

PR #1198, opened 2022-04-01
Automatically infer FORCE_COLORED_OUTPUT

hugomg:
Teach cmake how to deduce if a human is calling it from a terminal.
This allows us to get rid of the FORCE_COLORED_OUTPUT cmake option.
One less thing for the user to worry about.


hugomg:
To see the difference, add a compilation error to one of the C++ files and then compare what happens if you run `cmake .. | cat` vs the usual `cmake ..`.  In the first case, the compiler errors have have their color disabled. In the second case it has color.

A potential issue is if the user manually runs cmake in their terminal (therefore enabling colors) but pipes the output of ninja. In that case we might have wanted to disable colors, but this PR will cause the ninja output to contain the coloring escape sequences. Dunno if this is a real problem though. Any automated scripts that could potentially care about this are probably piping both the cmake and the ninja output, in which case this PR should behave exactly as you would expect.

o-:
nice!

PR #1200, opened 2022-04-07
Arithmetic fixes and refactorings

JanJecmen:


rir/src/compiler/native/lower_function_llvm.cpp
```diff
@@ -1691,7 +1700,10 @@ void LowerFunctionLLVM::compileBinop(
 
     builder.SetInsertPoint(done);
     if (rep == Rep::SEXP) {
-        setVal(i, box(res(), lhs->type.mergeWithConversion(rhs->type), false));
+        auto to = lhs->type.mergeWithConversion(rhs->type);
+        if (intIntToReal)
+            to = to.notT(RType::integer).orT(RType::real);
```
JanJecmen: @o- not sure this is the best way.. need to box as real but both lhs and rhs are ints

o-: sounds good to me, R agrees

```
> typeof(1L/1L)
[1] "double"
```

o-:
shit got real!

rir/src/compiler/native/lower_function_llvm.cpp
o-: wait! actually shouldnt this be `box(res, i->type, false)`?  I mean the type of `i` should really be `real` in this case, otherwise we will surely have more problems still. I am actually afraid that we made this same bug in type inference too...

JanJecmen: Ah, right.. the type of Div is `(int|real)` though.. so then the inference indeed must figure out that it in fact should be real only

o-: right, ic. not sure which one is the better solution then...

o-:

* WARNING: the longest CI job test_gctorture1 took 3.49h

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2304690748&job_ids[]=2297910374&selection=all

o-:

* WARNING: the longest CI job test_release_2 took 2.43h
* are-we-fast-r suite improved by 1.02
* Storage regressed by 0.95
* nbody_naive regressed by 0.81
* mandelbrot_naive_ascii regressed by 0.94
* emptyFor regressed by 0.82

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2331644682&job_ids[]=2326164049&selection=all

rir/src/compiler/native/lower_function_llvm.cpp
```diff
@@ -1546,10 +1560,11 @@ llvm::Value* LowerFunctionLLVM::depromise(Value* v) {
 }
 
 void LowerFunctionLLVM::compileRelop(
-    Instruction* i,
+    BinopKind kind, Instruction* i,
```
o-: I know nothing to do with your PR, but couldn't we find out the binop kind from the instruction?

rir/src/compiler/native/lower_function_llvm.cpp
```diff
@@ -1643,31 +1658,35 @@ void LowerFunctionLLVM::compileBinop(
...
                                                 "isNa", fun);
                 nacheck(llvmValue, type, isNaBr);
             }
         }
     };
-    checkNa(a, lhs->type, lhsRep);
-    checkNa(b, rhs->type, rhsRep);
+    checkNaBypassInsert(
+        a, b, [&]() { checkNa(a, lhs->type, lhsRep); },
+        [&]() { checkNa(b, rhs->type, rhsRep); });
```
o-: hmm, I find that a bit complicated tbh. any chance to make it easier. E.g.:
```
if (customNaCheck) {
  customNaCheck(a,b)
} else {
  checkNa(a, lhs->type, rhsRep);
  checkNa(b, rhs->type, rhsRep);
}
```

rir/src/compiler/native/lower_function_llvm.cpp
```diff
@@ -3652,60 +3669,120 @@ void LowerFunctionLLVM::compile() {
...
+                    });
                 break;
 
-            case Tag::Pow:
+            case Tag::Pow: {
+                auto bypassNaCheck = [&](llvm::Value* a, llvm::Value* b,
+                                         const Action& checkA,
+                                         const Action& checkB) {
+                    // In the following cases we observe GNU R to return 1 but
+                    // our check for NA would take precedence and result in NA:
```
o-: ugh, good catch!

rir/src/compiler/pir/instruction.h
```diff
@@ -412,25 +412,42 @@ class Instruction : public Value {
...
-            }
-            // the binop result becomes NA if it can't be represented in a
+
+            // The binop result becomes NA if it can't be represented in a
             // fixpoint integer (e.g. INT_MAX + 1 == NA)
-            // * the condition checks iff at least one of the arguments is an
+            // The condition checks if at least one of the arguments is an
             // integer (doesn't happen with only logicals), and the result is an
             // integer (doesn't happen with real coercion)
             if (m.maybe(RType::integer) && t.maybe(RType::integer))
```
o-, line 449: uhm, could this interact with the change of int->real below?

rir/src/compiler/pir/instruction.h
```diff
@@ -412,25 +412,42 @@ class Instruction : public Value {
     PirType inferredTypeForArithmeticInstruction(const GetType& getType) const {
         auto m = mergedInputType(getType);
         if (!m.maybeObj()) {
-            auto t = PirType::bottom();
+            // Everything but numbers throws an error
+            auto t = m & PirType::num().notMissing().orAttribsOrObj();
```
o-: why was it neccessary to add `m` here?

It seems the old strategy was start add bottom, add what is needed. The new strategy seems start with number and remove what is not needed. the old strategy seems to pontentially be more precise, not?

rir/tests/pir_binary_arith.R
```diff
@@ -0,0 +1,1873 @@
+# ================ This generates the tests below ================
```
o-, line 1: yay, tests!!

rir/src/compiler/pir/instruction.h
JanJecmen: Good catch, thanks!

rir/src/compiler/pir/instruction.h
JanJecmen: good point

o-:

* WARNING: the longest CI job test_features_3 took 2.8h
* are-we-fast-r suite improved by 1.04
* nbody_naive regressed by 0.82
* scalarFor regressed by 0.91

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2336581801&job_ids[]=2326164049&selection=all

PR #1203, opened 2022-04-25
Don't dispatch to native code that's still being compiled

JanJecmen:
In the native backend a Code object could be accessed from a DispatchTable while still being compiled
(we call deparse to print pir constants which can eval code). Such a Code would wrongly think it's bytecode
and start interpreting garbage.

This adds a pending flag that prevent dispatching to function versions that have pending Code objects. The flag
is set when we create a native Code object and reset when we finalize the LLVM module, at which point native
code can be emitted.

### DIFF
--- a/rir/src/runtime/Code.h
+++ b/rir/src/runtime/Code.h
@@ -74,12 +74,13 @@ struct Code : public RirRuntimeObject<Code, CODE_MAGIC> {
                      size_t bindingCache);
     static Code* New(Immediate ast, size_t codeSize, size_t sources,
                      size_t locals, size_t bindingCache);
-    static Code* New(Immediate ast);
+    static Code* NewNative(Immediate ast);
 
     constexpr static size_t MAX_CODE_HANDLE_LENGTH = 64;
 
   private:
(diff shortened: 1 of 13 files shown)


rir/src/runtime/Code.h
```diff
@@ -93,18 +94,20 @@ struct Code : public RirRuntimeObject<Code, CODE_MAGIC> {
...
-        if (*lazyCodeHandle_ == '\0')
+        if (pending_ || *lazyCodeHandle_ == '\0')
             return nullptr;
         return lazyCompile();
     }
 
     bool isCompiled() const {
         return *lazyCodeHandle_ != '\0' && nativeCode_ != nullptr;
     }
+    bool pending() const { return pending_; }
```
o-: pending is a bit non-descript. can we make it more concrete? like `pendingCompilation` and maybe a comment about the possible in-between state?

o-:

* WARNING: the longest CI job test_release_2 took 2.31h
* are-we-fast-r suite improved by 1.04
* reversecomplement regressed by 0.87

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2377761846&job_ids[]=2369656380&selection=all

PR #1208, opened 2022-05-04
Reduce version cloning

### LINKED ISSUE #1217: Reduce unnecessary cloning of versions AND slow convergence to the optimal context
When running the snippet of code below using `PIR_WARMUP=3 PIR_DEBUG=PrintPirAfterOpt,PrintToStdout`, we see that `f1 (f[0])` shows up several times in the compilation log, each time with a more specific context. The reason is that it takes several passes to the compiler to reach the optimal context.  When the eager_call pass picks up a better context, it clones the current version and changes it. This is also not ideal since oftentimes we could reuse the same version and change it in place (if we can show that is not referenced somewhere else).
Both problems should be looked into. The PR linked here attempts to limit the cloning of versions by keeping a simple refCount mechanism. It doesn't seem to show much benefits,  so maybe it is not worth merging. However, it does fix the issue in the example and manages to reduce the number of times the function gets cloned.

```
f <- function() {

    f1 <- function(x) {
        #print("inner")
         x + 4L
    }

    for (i in 1:10) {
        print(f1(i))
    }

    rir.functionInvocations(f1)
}

f()
f()
f()
f()
f()
f()
f()
```

skrynski:
It tries to avoid the cloning  of versions and applies modifications in-place. Added some flags to track references to versions and catch simple cases


skrynski:
Be careful with lastSeen variable, that might get GCed and reused. It might be better to use incremental ids

o-:

* WARNING: the longest CI job test_features_3 took 3.89h
* are-we-fast-r suite regressed by 0.96
* simple suite improved by 1.03
* Bounce_nonames regressed by 0.85
* mandelbrot_naive_ascii regressed by 0.9
* fasta regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2443680603&job_ids[]=2427588749&selection=all

o-:

* WARNING: the longest CI job test_features_1 took 2.66h

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2468496070&job_ids[]=2467128976&selection=all

PR #1210, opened 2022-05-12
Fix patching the constant pool

JanJecmen:
This should fix the bug where occasionally the close_ bytecode would find a Function in the constant pool instead of a DispatchTable.

Some cells in the constant pool are created by makeSpace with the intention that we can later patch these (used for invalidating targets after deopts).
However, the patching would also write to the C++ contents cache that maps SEXPs to pool indices, and patching multiple times would leave the old mapping outdated.

For example:
1. makeSpace 42
2. patch X into 42
3. patch Y into 42
4. X gets garbage collected
5. Z gets allocated to where X used to be
6. insert of Z finds X in contents and returns 42
Now:
- constant pool at 42 returns Y
- Z may get collected since it's not reachable from the constant pool

Also, this makes sure that we cannot patch a location that doesn't come from calling makeSpace.

### DIFF
--- a/rir/src/utils/Pool.cpp
+++ b/rir/src/utils/Pool.cpp
@@ -8,6 +8,7 @@ namespace rir {
 std::unordered_map<double, unsigned> Pool::numbers;
 std::unordered_map<int, unsigned> Pool::ints;
 std::unordered_map<SEXP, size_t> Pool::contents;
+std::unordered_set<size_t> Pool::patchable;
 
 BC::PoolIdx Pool::getNum(double n) {
     if (numbers.count(n))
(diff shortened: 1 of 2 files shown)


o-:

* WARNING: the longest CI job test_features_3 took 2.87h
* are-we-fast-r suite regressed by 0.97
* simple suite improved by 1.02
* Bounce_nonames_simple regressed by 0.88
* reversecomplement regressed by 0.92
* nbody regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2445941207&job_ids[]=2427588749&selection=all

PR #1214, opened 2022-05-16
Context computed for StaticCall when compiling rir2pir was wrongly handling MissingArg

skrynski:
Context computed for static call handled MissingArg as a special case instead of  dispatching to arg->callArgTypeToContext() which already handles that case. The logic on rir2pir.cpp was different to the on the method callArgTypeToContext. This led to some  unnecessary compilation of versions since the context was not properly computed


o-:

* WARNING: the longest CI job test_features_1 took 3.48h
* fasta regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=2460872045&job_ids[]=2458195032&selection=all

PR #1240, opened 2023-05-31
Compiler client+server

Jakobeha:
Compiler client+server

What I have so far:

- zeromq (libzmq and cppzmq) dependency added to `external`
- executable prints additional messages whether or not `PIR_CLIENT_ADDR` or `PIR_SERVER_ADDR` are set
- If `PIR_CLIENT_ADDR` is set to an address, executable will connect to the compiler server at the address, and then whenever `pirCompile` is called it will:
  - send a message before compiling (currently the `SEXP` pointer + assumptions + name + debug options)
  - receive a response (currently the `SEXP` pointer)
  - check the response is the same pointer (testing that messages are transmitted without corruption)
  - check the "response PIR" (currently a dummy string) with the actual final-pir, and log if there is a discrepancy (currently always the case)
- If `PIR_SERVER_ADDR` is set to an address, executable will bind to the address and then go into an infinite loop waiting for and responding to messages


o-:

* WARNING: the longest CI job test_release_2 took 2.09h
* mandelbrot_naive_ascii regressed by 0.93

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=4384648625&job_ids[]=4371607270&selection=all

PR #1241, opened 2023-06-08
Refactor missing type

### LINKED ISSUE #1196: Fix missingness type.
Currently `RType::missing` is always interpreted as outside of the promise. E.g., if we have a promised type that has `RType::missing` and `RType::int/real` set, then this is interpreted as `((int|real)~|missing)` and not `(int|real|missing)~` as expected. I.e., we say it is (1) the missing value or (2) a promise that returns int or real. More natural interpretation would be: a promise that returns: int,real, or missing. This would be more in line with all other uses of RTypes.

The reason is that often it is important to know that a particular value is not *the* missing value, since for instance ChkMissing only errors on the missing value and not on a promised missing value.

This also forces us to have the `Force` instruction re-add the missing. I.e. `(t)~ ---[Force]---> (t|missing)`

In the following I use `prom` to mean `maybePromiseWrapped` or `maybeLazy`.

I.e., currently this is our interpretation of the missing type:

```
RType           Flag              Interpretation                           Printed As
int                               N                                        int
missing                           R_MissingValue                           missing
int,missing                       N | R_MissingValue                       missing
* - missing                       ...                                      val  (ugh!)
*                                 ...                                      val? (ugh!)
missing         prom              Promise(void) | R_MissingValue           missing~  (ugh!!! wrong!!!!)
int,missing     prom              Promise(N) | R_MissingValue              (missing | int)~ (ugh!!! wrong!!!!)
```

But this non orthogonal interpretation of the RType flags leads to many confusions, so we should fix it! A cleaner solution would be to have a type flag `maybeMissing`. The downside is that for non promised types the `maybeMissing` flag can contradict the Rtypes, which must be avoided.

We can express the different types as follows:

```
RType           Flag                   Interpretation                                  Printed As
int                                    N                                               int
int             maybeMissing           ILLEGAL!
missing         maybeMissing           R_MissingValue                                  missing
missing                                ILLEGAL!
int,missing                            N | R_MissingValue                              int|missing
missing         prom                   Promise(R_MissingValue)                         missing^
int,missing     prom                   Promise(N | R_MissingValue)                     (int|missing)^
missing         prom,maybeMissing      Promise(R_MissingValue) | R_MissingValue        missing^?
int,missing     prom,maybeMissing      Promise(N | R_MissingValue) | R_MissingValue    (int|missing)^?
```

This would also allow us to remove the hacky flag called `noWrappedMissing` which is purely an artifact of our mis-use of the `RType::missing`.

Under this new interpretation `Force` does not change `RType::missing`, but simply adds the missing flag, if the rtype contains missing.

We should also refactor all missingness related apis on PirType. Here is a proposal for a new api:

```
isTheMissingValue()       :=  rtype == (missing) && !promiseWrapped
maybeTheMissingValue()    :=  flag.include(maybeMissing)
evaluatesToMissing()      :=  rtype == (missing)
maybeEvaluatesToMissing() :=  rtype.include(missing)
```

### LINKED ISSUE #1197: constantfold ChkMissing
If a value can be proved to not be *the* R_MissingValue, then ChkMissing should be removed here: https://github.com/reactorlabs/rir/blob/master/rir/src/compiler/opt/constantfold.cpp#L332

This is e.g. the case if the value is guaranteed to be a promise, for instance a `MkArg`, or if the value is guaranteed to not be a promise and not has the missing type.

skrynski:


rir/src/compiler/test/PirTests.cpp
```diff
@@ -1074,7 +1080,9 @@ static Test tests[] = {
              return test42("{a<- 41L; b<- 1L; f <- function(x,y) x+y; f(a,b)}");
          }),
     Test("Test dead store analysis", &testDeadStore),
-    Test("Test type rules", &testTypeRules)};
+    Test("Test type rules", &testTypeRules)
+
```
Jakobeha: Extra newline?

rir/src/api.h
```diff
@@ -29,4 +29,6 @@ REXPORT SEXP rirDeserialize(SEXP file);
 REXPORT SEXP rirSetUserContext(SEXP f, SEXP udc);
 REXPORT SEXP rirCreateSimpleIntContext();
 
+REXPORT SEXP playground();
```
Jakobeha, line 34: Where is this used / what for? Maybe add a comment

o-:

* WARNING: the longest CI job test_release_2 took 2.09h
* binarytrees_2 regressed by 0.9
* fasta_naive_2 regressed by 0.91
* nbody regressed by 0.94
* profiler-microbenchmark regressed by 0.94

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=4455202256&job_ids[]=4452171149&selection=all

rir/src/compiler/test/PirTests.cpp
skrynski: thanks

PR #1276, opened 2024-04-12
added missing set/reset dispatch table for  Function

skrynski:
Also added/fixed  a few asserts


skrynski:
The changes come from branch `record-rebase`

PR #1309, opened 2024-08-09
Remove redundant CMake target

rihafilip:
The CMakeLists.txt file has an extra target, that does nothing.
The `build-gnur.sh` script changes it's behaviour only with the `--macos_gcc9` argument, thus calling it again with `custom-r` is redundant and might even be incorrect on MacOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,14 +146,8 @@ else()
     )
 endif()
 
-add_custom_target(default-gnur
-    DEPENDS dependencies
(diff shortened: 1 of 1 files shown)


o-:

* WARNING: the longest CI job test_release_2 took 2.38h
* Bounce_nonames regressed by 0.92
* reversecomplement_naive regressed by 0.92
* fasta_2 regressed by 0.93
* reversecomplement_2 regressed by 0.93

Please find your performance results at https://rir-benchmarks.prl.fit.cvut.cz/diff?job_ids[]=7546830816&job_ids[]=7545763944&selection=all

